PR #85, opened 2019-07-13
Update README with installation instructions

maxneuvians:
Update README with installation instructions


maxneuvians:
Don't run tests after 8 PM EST

PR #144, opened 2019-07-24
Modify csv bucket name

timarney:
Note: we can pull this out to the .env file later


damcom333:
This change looks temporary. I'm ok with it and we can revisit once we go to the next phase of delivery.

Issue #60, opened 2019-07-08
When submitting forms that call ZenDesk we should replace it with a Simple contact method for now so meet immediate needs


willeybryan:
Support page and form changes are on ticket https://github.com/cds-snc/notification-api/issues/68

**What?**
- (/service-settings/request-to-go-live/estimate-usage), (/branding-request/email) (https://notifier.cdssandbox.xyz/support/report-problem),(https://notifier.cdssandbox.xyz/support/ask-question-give-feedback) all use Zen Desk
- We need to remove links to Zen Desk. 

- Cut Email Branding
![image](https://user-images.githubusercontent.com/46505127/60914873-72e5df00-a258-11e9-893f-f787ee4c05c2.png)

- Change Request to go live to contact us with link to /support in order to stream line
![image](https://user-images.githubusercontent.com/46505127/60914940-9577f800-a258-11e9-85f2-c244059fdca1.png)

**Why?**
- We need to keep support just scale it back to immediate need so that is on https://github.com/cds-snc/notification-api/issues/68
- Go live can be handled manually for as we need to figure out payment it is still a ways away. 
- We can add in Email branding later and do it differently then having it create a support ticket some where. 
- Change to Platform Email CDSPlatform.PlateformesSNC@tbs-sct.gc.ca

willeybryan:
@timarney I saw you dropped this in, I turned it to an issue and want to fill it out. 

I was wanted to confirm where it uses Zendesk?
is it in the 'Tell us how many messages you expect to send' in the 'request to go live' area? /service-settings/request-to-go-live/estimate-usage

If so is the option to have that sent out as an email instead?

willeybryan:
I imagine these may also use zendesk https://notifier.cdssandbox.xyz/support/report-problem and https://notifier.cdssandbox.xyz/support/ask-question-give-feedback

timarney:
Yes they do.

 First one (is when you request a template / branding for your service)
<img width="1087" alt="Screen Shot 2019-07-09 at 1 01 48 PM" src="https://user-images.githubusercontent.com/62242/60908393-c13fb180-a249-11e9-9c53-439a522690b5.png">

willeybryan:
Yeah I see that now, under /branding-request/email

PR #161, opened 2019-07-29
Add endpoint for support emails

brdunfield:
Adds the endpoint for use in #60 

Only tested / built with the contact us form for now. Endpoint is called in https://github.com/cds-snc/notification-admin/pull/85

Still needs tests


app/config.py
```diff
@@ -155,6 +155,7 @@ class Config(object):
     MOU_SIGNED_ON_BEHALF_SIGNER_RECEIPT_TEMPLATE_ID = 'c20206d5-bf03-4002-9a90-37d5032d9e84'
     MOU_SIGNED_ON_BEHALF_ON_BEHALF_RECEIPT_TEMPLATE_ID = '522b6657-5ca5-4368-a294-6b527703bd0b'
     MOU_NOTIFY_TEAM_ALERT_TEMPLATE_ID = 'd0e66c4c-0c50-43f0-94f5-f85b613202d4'
+    CONTACT_US_TEMPLATE_ID = '8ea9b7a0-a824-4dd3-a4c3-1f508ed20a69'
```
maxneuvians: Rest of PR looks good, the only open issue is that the CONTACT US template needs to be generated using a database migration so that if you have to stand the product up from scratch you can. Take a look here how GDS does it:

https://github.com/cds-snc/notification-api/blob/master/migrations/versions/0298_add_mou_signed_receipt.py

Note that they hard code the ID so it remains consistent.

Issue #199, opened 2019-08-06
The Time zone needs to be configurable by the API as it cannot be made dynamic to browser.


willeybryan:
Created by https://github.com/cds-snc/notification-admin/pull/110

Related to https://github.com/cds-snc/notification-api/issues/118

**What?**
- Make the API timezone configurable for when it runs the reports.
- should we make it a setting on this page: https://notification.cdssandbox.xyz/services/142de8b0-7555-4df9-aca5-f64ca998aa62/api with a default of eastern?
- OR should it be a variable passed into the api when requesting a repot, and if it isn't present you get UTC

**Why?**
- API needs control over the time zone for services to make use of

PR #213, opened 2019-08-08
Added functionality for local timezone in API

maxneuvians:
Closes https://github.com/cds-snc/notification-api/issues/199. You can now use the `TIMEZONE` env to set a timezone for the server. It currently defaults to `America/Toronto`. The tests have been changed to reflect an EST timezone. 

The problem currently is that there is a a column in one of the tables called `bst_date` which holds the date for when a report is run and the aggregate statistics. This will cause some confusion and will need to be removed in a next iteration. However, we need more validation of this product before doing so.

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -8,7 +8,7 @@ from flask import (
     Blueprint
 )
 from notifications_utils.letter_timings import letter_can_be_cancelled
-from notifications_utils.timezones import convert_utc_to_est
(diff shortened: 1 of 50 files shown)


app/service/rest.py
```diff
@@ -611,6 +613,7 @@ def get_monthly_template_usage(service_id):
             end_date=end_date,
             service_id=service_id
         )
+        print(data)
```
timarney: Should remove this ya?

PR #305, opened 2019-08-30
Adds attachments to emails if they are part of download links

### LINKED ISSUE #287: Allow Attachments to be attached to emails in Notify
**Story**
Developers need to be able to attach PDF files with the API so they can send people a letter as part of their email to satisfy their obligations.

**What?**
- Validate that we can attachment
- Allow files to be attached to the actual emails with the API (If possible???)
- Test with PDF files
- Toggle in platform Admin interface

**Backup plan**
- We send users a SEND link with a password via email
- We Text users the Password so that they both aren't emails

**Why?**
- Services want to send customized PDF letters to people they would otherwise mail. Allowing them to do this improves chances of adoption

maxneuvians:
Closes #287 

Added the following changes:

- Instead of returning just the public URL the upload document now returns the direct download and public URL, which is then substituted in for the personalization.
- When the message is processed for sending it checks if any of the personalizations are a download link.
- If it is a download link, it downloads the content, determines the mime type, and if it is a PDF, attaches it to the message.
- Currently only AWS SES is implemented (also it is the only option we have)

### DIFF
--- a/app/clients/document_download.py
+++ b/app/clients/document_download.py
@@ -51,5 +51,4 @@ class DocumentDownloadClient:
             )
 
             raise error
-
-        return response.json()['document']['url']
+        return response.json()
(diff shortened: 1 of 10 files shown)


willeybryan:
So is there going to be a public URL always?

Issue #211, opened 2019-08-07
Uploads should be scanned for Viruses


willeybryan:
**What?**
- Select a Virus scan that can run on stuff uploaded to S3 Buckets
- Make it run on all CSV uploads at first
- Test to make sure they still work

**Why?**
-  We will need it for people uploading attachments to sister attachment service
- .CSV files is all people can upload on trial modes and such

maxneuvians:
Still waiting on approval from mlwr.ca

maxneuvians:
Received approval and tested out the API which looks like it works well. The big difference is that this service is async. You send up your file, the server tells you it is in a queue, and then you query until the file is processed. The UK code uses a sync version where you upload the file and then the wait till the server scans it and sends a response back to you.

Making the existing code async will require time. Also we need to verify the level of confidentiality of files we can upload to mlwr.ca. As a result we should prove out the upload file use case.

PR #307, opened 2019-09-03
Added callback to mlwr if an sid is in the document

maxneuvians:
Closes https://github.com/cds-snc/notification-api/issues/211

The code does the following. If a `mlwr_sid` is included in the payload for a document download it checks if the mlwr scan is complete. If it is not, it delays the sending by 60 seconds and then checks again. If the score of file is 500 or greater it fails the sending of the message with the status that it did not pass the malware scan.


timarney:
I think this is a reasonable solution for now as far as the mlwr_sid.  We'll likely need to take a deeper look at this in the future.

Issue #333, opened 2019-09-11
We need a backup email service should something happen to AWS SES


willeybryan:
**Story**
Developers and Decision makers want to know they can rely on the service to send emails reliably so we should have a backup for our email provider

**What?**
- Investigate other cloud based service providers of email sending
- Connect us to one so they show up in the list here https://notification.alpha.canada.ca/providers
- Use them as a backup incase SES fails
- Send grid as backup?

**Why?**
- We could get banned from SES for reasons beyond our control.
- A user of notify shouldn't be able to take it down for everyone.

PR #351, opened 2019-09-19
SendGrid integration

maxneuvians:
Closes https://github.com/cds-snc/notification-api/issues/333


app/clients/email/sendgrid_client.py
```diff
@@ -0,0 +1,96 @@
...
+                   source,
+                   to_addresses,
+                   subject,
+                   body,
+                   html_body='',
+                   reply_to_address=None,
+                   attachments=[]):
+        try:
+            if "<" in source:
+                source = source.split("<")[1].split(">")[0]
```
timarney: Let's add an example comment here.

PR #459, opened 2019-11-20
Alert account change

timarney:


app/user/rest.py
```diff
@@ -132,8 +132,23 @@ def update_user_attribute(user_id):
 
     service = Service.query.get(current_app.config['NOTIFY_SERVICE_ID'])
 
+    change_type = "see account"
+
+    if 'email_address' in update_dct:
```
timarney: @max not crazy about doing something like this.  Can break into another function but still the same type of logic ... thoughts?

Also - need a few more types ... would be nice to capture permission changes + password updates.

willeybryan:
image

PR #520, opened 2019-12-13
SMTP PoC

### LINKED ISSUE #517: SMTP Relays
Investigate steps of provisioning SMTP relays as part of the service with minimal admin intervention and facilitating the same level of tracking (received, rejected, bounced).

maxneuvians:
Closes #517

### DIFF
--- a/app/clients/smtp/aws.py
+++ b/app/clients/smtp/aws.py
@@ -1,0 +1,207 @@
+import boto3
+from flask import current_app
+
+import hmac  # required to compute the HMAC key
+import hashlib  # required to create a SHA256 hash
+import base64  # required to encode the computed key
+import time
+
+
+def smtp_add(name):
+    ses_client = boto3.client('ses', region_name=current_app.config["AWS_REGION"])
+    r53_client = boto3.client('route53', region_name=current_app.config["AWS_REGION"])
+    iam_client = boto3.client('iam', region_name=current_app.config["AWS_REGION"])
+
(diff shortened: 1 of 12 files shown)


app/clients/smtp/aws.py
```diff
@@ -0,0 +1,207 @@
...
+        )
+
+        response = client.create_access_key(
+            UserName=user_name
+        )
+
+        credentials = {
+            "iam": user_name,
+            "domain": name,
+            "name": "email-smtp.us-east-1.amazonaws.com",
```
obrien-j: Guessing this is likely a fixed string, but is it worth pulling in from env just for future proofing and isolation of config from code?

app/clients/smtp/aws.py
```diff
@@ -0,0 +1,207 @@
...
+import time
+
+
+def smtp_add(name):
+    ses_client = boto3.client('ses', region_name=current_app.config["AWS_REGION"])
+    r53_client = boto3.client('route53', region_name=current_app.config["AWS_REGION"])
+    iam_client = boto3.client('iam', region_name=current_app.config["AWS_REGION"])
+
+    # name = name + '.m.' + current_app.config["NOTIFY_EMAIL_DOMAIN"]
+    name = name + '.m.' + 'notification.alpha.cdssandbox.xyz'
```
obrien-j: Assume there's no way to have some string termination in either the strcat, or in amazon's parsing of the domain, that would allow for a domain that's *not* a sub of the cdssandbox.xyz... May be worth a quick check.

PR #590, opened 2020-01-29
Patch UTF-8 chars issue in email sender name

maxneuvians:


timarney:
**AWS reference issue:** 

https://github.com/aws/aws-sdk-js/issues/1585

**IDNA**

https://docs.python.org/3/library/codecs.html#module-encodings.idna

```
A domain name containing non-ASCII characters (such as www.Alliancefrançaise.nu) is converted into an ASCII-compatible encoding (ACE, such as www.xn--alliancefranaise-npb.nu). The ACE form of the domain name is then used in all places where arbitrary characters are not allowed by the protocol, such as DNS queries, HTTP Host fields, and so on. 
```

Issue #620, opened 2020-02-14
Simplify CI pipeline and setup scripts for easier dev access, better deployment, and improved security


maxneuvians:

maxneuvians:
Also https://github.com/cds-snc/notification-api/commit/42233264ac60d5a77b8f084d689b959e2f9893e4

PR #644, opened 2020-02-26
Try pushing to Github docker repo

maxneuvians:
Part of https://github.com/cds-snc/notification-api/issues/620


maxneuvians:
GitHub needs to make their public images actually public :( https://github.community/t5/GitHub-Actions/docker-pull-from-public-GitHub-Package-Registry-fail-with-quot/td-p/32782

PR #698, opened 2020-03-18
Added a new endpoint for api key stats

smcmurtry:
Added a new endpoint to get stats for an api key. Added very simplistic query to get total notifications sent per key.

### DIFF
--- a/app/dao/fact_notification_status_dao.py
+++ b/app/dao/fact_notification_status_dao.py
@@ -207,7 +207,7 @@ def get_total_notifications_sent_for_api_key(api_key_id):
     """
     SELECT count(*) as total_send_attempts
     FROM notifications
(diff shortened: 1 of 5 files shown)


app/dao/fact_notification_status_dao.py
```diff
@@ -203,6 +203,20 @@ def fetch_notification_status_for_service_for_today_and_7_previous_days(service_
     ).all()
 
 
+def get_total_notifications_sent_for_api_key(api_key_id):
+    """
+    SELECT count(*) as total_send_attempts
+    FROM notifications
+    WHERE api_key_id = 'whatever'
```
timarney: let's change whatever here to api_key_id

timarney:
Can you add a few tests for this.

Issue #695, opened 2020-03-17
API Key monitoring page(s)


willeybryan:
**Story**
As a Notify platform administrator I need to be able to monitor api key usage.

**What?**
Add additional details + a new summary page to allow better insights into api key usage. 

**Why?**
- Risk register req

**Notes**

1) API key overview (overview all keys that have sent in last X days)
=============================
Top senders last X amount of time
[Service name](link to service name)
Key name - total sends (email, sms) - services/e609f6b0-0390-45b0-aaae-f4cc92c713e4/api/keys so you can drill down

2) Key details ... services/e609f6b0-0390-45b0-aaae-f4cc92c713e4/api/keys)
=============================
​
Key name - total sends (email, sms)
  - key created, expires, last activity
  - [daily sends](link to graph sends per day)
  - Click on day

3) Day view for API key (linked from daily sends graph)
=============================
Summary of notifications sent 
date time , Template ID (link to template I

**To do for api:**
- [x] add another endpoint to get a list of api keys that have sent in the past 48 hours, sorted by number of sends
- [x] allow the client to specify the total for: last n days
- [x] allow the client to specify: email sends, sms sends, or total
- [ ] allow the client to specify: all notifications ever queued, all notifications queued but unsent, all successful sends, all failed sends, etc.

**To do for admin site:**
- [x] add a new page visible to platform admin, but not yet linked in the side nav
- [x] get the ranked list of api keys on the page, with links
- [x] add send stats
- [ ] implement some filtering/searching

timarney:
@smcmurtry feel free to break this down into smaller tickets

timarney:
I think we should hold on the filtering/searching for now - we'll get a better idea of what we actually need one we have the initial version of the overall summary page up.

timarney:
Calling this "done for now". 

What @smcmurtry has built out gives us much better insights.

PR #710, opened 2020-03-24
Add new endpoint to get api keys ranked by # of notifications created

smcmurtry:
Working on  #695 

Added a new endpoint to get a list of api keys ranked by # of notifications created. Returned data is like:

```
[
  {
    'api_key_name': 'live key 1', 
    'api_key_type': 'normal', 
    'service_name': 'test service 1',
    'api_key_id': UUID('3dddee16-946c-4f84-bf12-94f68cb6757d'),
    'service_id': UUID('d1125123-9a19-4498-90fa-7946e8f10d97'), 
    'last_notification_created': '2020-03-16T18:20:17.803858Z', 
    'email_notifications': 10,
    'sms_notifications': 1,
    'total_notifications': 11
  }, 
  ... 
]
```


app/dao/fact_notification_status_dao.py
```diff
@@ -242,6 +243,63 @@ def get_last_send_for_api_key(api_key_id):
...
+
+    return db.session.query(
+        table_a.c.api_key_id,
+        table_a.c.service_id,
+        table_a.c.last_notification_created,
+        table_a.c.total_notifications,
+        ApiKey.name,
+        ApiKey.key_type,
+        Service.name
+    ).join(
```
smcmurtry: Doing the join on the subquery means we will be joining on a lot fewer rows.

app/api_key/rest.py
```diff
@@ -25,3 +29,26 @@ def get_api_key_stats(api_key_id):
     }
     data["total_sends"] = data["email_sends"] + data["sms_sends"]
     return jsonify(data=data)
+
+
+@api_key_blueprint.route('/ranked-by-notifications-created/<n_days_back>', methods=['GET'])
+def get_api_keys_ranked(n_days_back):
+    try:
+        n_days_back = int(n_days_back)
```
smcmurtry: How much should I protect against bad input here? should I check that `n_days_back` isn't negative?

app/api_key/rest.py
```diff
@@ -25,3 +29,26 @@ def get_api_key_stats(api_key_id):
...
+    except ValueError:
+        return jsonify(data=[])
```
smcmurtry: Should I be returning something here that says there was an error? or is this adequate?

app/dao/fact_notification_status_dao.py
```diff
@@ -242,6 +243,63 @@ def get_last_send_for_api_key(api_key_id):
...
+        SELECT
+            api_key_id,
+            service_id,
+            max(created_at) as last_send,
+            count(*) as total_notifications
+        FROM notifications
+        WHERE created_at > start_date and api_key_id is not null
+        GROUP BY api_key_id, service_id
+        ORDER BY count(*) DESC
+        LIMIT 50
```
smcmurtry: This limit is arbitrary but I wanted to limit the size of the returned data. In the future it might make sense to do pagination if we want to view more than the top 50 keys or however many fit on the 1st page.

app/dao/fact_notification_status_dao.py
```diff
@@ -242,6 +243,63 @@ def get_last_send_for_api_key(api_key_id):
...
+        Notification.service_id
+    ).order_by(
+        func.count(Notification.id).desc()
+    ).subquery()
+
+    return db.session.query(
+        table_a.c.api_key_id,
+        table_a.c.service_id,
+        table_a.c.last_notification_created,
+        table_a.c.total_notifications,
```
smcmurtry: Not sure what the `.c.` means, I just copied it from another example of a subquery on this page. And it seems to work!

tests/app/api_key/test_rest.py
```diff
@@ -52,3 +55,32 @@ def test_get_api_key_stats_no_sends(admin_request, notify_db, notify_db_session)
...
+
+    assert api_keys_ranked[0]["api_key_name"] == api_key_1.name
+    assert api_keys_ranked[0]["service_name"] == service.name
+    assert api_keys_ranked[0]["total_notifications"] == total_sends + 1
+    assert "last_send" in api_keys_ranked[0]
+
+    assert api_keys_ranked[1]["api_key_name"] == api_key_2.name
+    assert api_keys_ranked[1]["service_name"] == service.name
+    assert api_keys_ranked[1]["total_notifications"] == total_sends
+    assert "last_send" in api_keys_ranked[0]
```
smcmurtry: I'm not testing that every key is there, or what the exact `last_send` time is. I figured I would just test the main functionality.

tests/app/dao/test_fact_notification_status_dao.py
```diff
@@ -338,6 +340,32 @@ def test_get_last_send_for_api_key(notify_db_session):
     assert abs(time_delta.total_seconds()) < 1
 
 
+def test_get_api_key_ranked_by_notifications_created(notify_db_session):
```
smcmurtry: My tests for the query vs the endpoint are not radically different - mostly cut and pasted but the returned data is in a different shape. Not sure if I should be trying to make these tests more different. The query tests seem a bit redundant since the queries are being run and the returned data checked as part of the endpoint tests.

app/dao/fact_notification_status_dao.py
```diff
@@ -242,6 +243,105 @@ def get_last_send_for_api_key(api_key_id):
...
+        b.c.sms_notifications,
+        b.c.total_notifications
+    ).join(
+        ApiKey, ApiKey.id == b.c.api_key_id
+    ).join(
+        Service, Service.id == b.c.service_id
+    ).order_by(
+        b.c.total_notifications.desc()
+    ).limit(
+        50
```
smcmurtry: This limit is arbitrary but I wanted to limit the size of the returned data. In the future it might make sense to do pagination if we want to view more than the top 50 keys or however many fit on the 1st page.

app/dao/fact_notification_status_dao.py
```diff
@@ -242,6 +243,105 @@ def get_last_send_for_api_key(api_key_id):
     ).all()
 
 
+def get_api_key_ranked_by_notifications_created(n_days_back):
```
smcmurtry: This is a monster query with case statements. Basically I'm getting the total email and sms notifications created for each api key. I could have moved more of this logic into python, but I figured SQL would be faster and there are other ugly queries in this file with `CASE` statements, like `fetch_monthly_notification_statuses_per_service` on line 595.

tests/app/api_key/test_rest.py
```diff
@@ -52,3 +55,36 @@ def test_get_api_key_stats_no_sends(admin_request, notify_db, notify_db_session)
...
+    assert api_keys_ranked[0]["email_notifications"] == total_sends + 1
+    assert api_keys_ranked[0]["total_notifications"] == total_sends + 1
+    assert "last_notification_created" in api_keys_ranked[0]
+
+    assert api_keys_ranked[1]["api_key_name"] == api_key_2.name
+    assert api_keys_ranked[1]["service_name"] == service.name
+    assert api_keys_ranked[1]["sms_notifications"] == 0
+    assert api_keys_ranked[1]["email_notifications"] == total_sends
+    assert api_keys_ranked[1]["total_notifications"] == total_sends
+    assert "last_notification_created" in api_keys_ranked[0]
```
smcmurtry: I'm not testing that every key is there, or what the exact last_send time is. I figured I would just test the main functionality.

Issue #732, opened 2020-04-01
Configure Pinpoint SMS sender


willeybryan:
**Story**
When sending SMS from my Notify service to the public I want it to come from the same number(s) every time so that we can build a trusting communication relationship

**What**
- Configure pinpoint for one dedicated long code
- Test it to see if it works for sending 1 sms at a time
- Test to see what happens if we through 10 at it or something

**Why?**
- Government comms people don't trust SNS's American numbers and don't want to send from that

**Done When?**
- We can safely offer long codes to a service for their use. 

**Notes**
Pinpoint https://docs.aws.amazon.com/pinpoint/latest/userguide/channels-sms-countries.html

timarney:
Docs
https://docs.aws.amazon.com/pinpoint/latest/developerguide/send-messages-sms.html

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/pinpoint.html

timarney:
Based on the docs above we'll be looking to adjust the current SES client send to use PinPoint

> I'm pretty sure you can mirror the ses client and tweak a few things @obrien-j

willeybryan:
Notes for remembering
- If a platform admin bumps up Pinpoint to the top priority, that would stop sms from being delivered for any service without an inbound sms assigned.
- We are still using SNS for sending sms, except when a service has an "inbound sms" assigned to it. For that we use Pinpoint.
- if we don't assign or create any longcodes, there won't be any inbound SMS numbers available to assign to services. All sms will be sent through SNS as before
- "Does every service need a long code?" No - until we have rate limiting in place, we should not assign inbound sms to any services. Otherwise we could exceed the rate limits enforced by telcos and get our longcodes blocked. Once that code is in place we could start with 1 service and see how that goes before assigning more.
- STOP/HELP messages are configurable in the Pinpoint dashboard - it looks like they will be the same messages across all our long codes for now

willeybryan:
Rate limiting seems like the continuation of this ticket.

timarney:
> Rate limiting seems like the continuation of this ticket.

Correct

ClementineHahn:
Hi @smcmurtry @timarney , what would you say is left to do to consider this done ? 
Because from my understanding/memories there is nothing left but well im not sure i have the best understanding of it ;)  Thank you !

timarney:
Another card could be opened for "Rate limiting"

https://github.com/cds-snc/notification-api/issues/732#issuecomment-621751346

[...]

PR #772, opened 2020-04-18
Implement AWS pinpoint client

smcmurtry:
Working on #732 

This PR adds a new client for AWS Pinpoint. It is configured w/ the following new env variables: `AWS_PINPOINT_APP_ID`, ~~`AWS_PINPOINT_LONG_CODE`~~, `AWS_PINPOINT_KEYWORD`.

I haven't implemented anything that would rotate through a number of longcodes to increase the send rate. The next step would be to implement that. I would prefer not to make this PR bigger than it needs to be, but I can include that work here if it makes sense. ~~What do you think @maxneuvians?~~ Direction from @willeybryan is that this is enough for now.


maxneuvians:
I would not recommend merging as is:

```
In the United States and Canada, sending rates for long codes are restricted to 1 message per second. This restriction is set by the phone carriers, and isn't a limitation of Amazon Pinpoint. This restriction might be higher or lower in other countries and regions. If you send a large volume of messages from a long code, wireless carriers might begin to block your messages. If you send SMS messages programmatically, your applications should limit the number of messages that they send each second.
```

from https://docs.aws.amazon.com/pinpoint/latest/userguide/channels-sms-awssupport-long-code.html

The code is currently limiting us to one long code.

willeybryan:
@maxneuvians 
Does merging this preclude us from keeping SNS?
In the providers section of the platform admin we can't we have pinpoint be a lower level provider? 
That way it is available so we can hook it up to a specific service for a specific phone number.

maxneuvians:
You unfortunately can not assign a specific provider based on service to my knowledge. 

In addition the current set up has a mechanism to switch between SMS providers automatically if it thinks one is too slow, so it actually needs to be functional if it is active.

The code also needs to be able to dynamically load a list of long codes from AWS so that we can add or remove long codes without the app having to be restarted. Let me know if that make sense, happy to jump in a call, or elaborate further here.

obrien-j:
We'll likely want to think through how we can 'make notify go slow' as it relates to single long codes per service.

I imagine the UK has the same challenges, might be worth reaching out to them.

maxneuvians:
`make notify go slow` can be done by rate limiting the SMS queue in SQS

willeybryan:
Ah crap, So recipient trust of SMS is the primary problem I am trying to solve and so here is what I was thinking;

We can not assign a specific provider based on service to my knowledge either, but what I had figured what we might do make that next. 

I want to Keep SNS because, it works, and we have trial accounts and they don't need anything fancy right now but we will have clients that want to send a few messages from a Canadian phone number and having it be the same one every time improves it's credibility. To these 2 use cases speed is not important to these people they are sending to ~200 people, PHAC and ISC.

NOW 
Is rate limit the queue for a specific provider(Pinpoint) and building a DB way to assign a number to a service more or less lift than building out the load balancer now? Or is the load balancer a blocker to do that? Do we need  the load balancer or is it just a nice to have?

Right now The way I see it we have the speed solution SNS, a trust solution with pinpoint, then we will have the speed trust solution which us short codes. Then we rate limit them independently and assign them per service.  

Could that work? and is that easier?

maxneuvians:
In that case what you need to do is:

1. Shutdown the code that checks for slow SMS providers (https://github.com/cds-snc/notification-api/blob/master/app/config.py#L223-L226)
2. Assign people phone numbers using the InboundNumber feature (https://github.com/cds-snc/notification-api/blob/master/app/dao/services_dao.py#L26) or something similar
3. When a notification is sent, check if there is an inbound number set, if yes, use pinpoint, if not use SNS
4. Ensure you are somehow rate limiting that task (https://celery.readthedocs.io/en/latest/userguide/tasks.html#Task.rate_limit). It might have to be a separate sending task (https://github.com/cds-snc/notification-api/blob/master/app/delivery/send_to_providers.py#L64-L69)

willeybryan:
After reviewing your comment and talking to the developers about what is involved in the load balancer solution.

I have decided that Max is right about the the code, ah he said tt needs to be able to dynamically load a list of long codes from AWS so that we can add or remove long codes without the app having to be restarted. 

Let's do that thing and get those people some trust!

smcmurtry:
I now understand more of how assigning an SMS reply-to number works currently:
- GDS had a command that would insert phone numbers from a file into the db (inbound_numbers table)
- one of these numbers can then be assigned to a service by a platform admin from the service settings page
- the db table is updated with the service id, and it is no longer available to be assigned to another service

Unfortunately AWS does not provide an API method to get the available long codes from pinpoint, so we will need to get them in our system manually.

I created a new s3 bucket on staging with a file in it that contains the longcodes. When you request a new longcode through the AWS dashboard, you will then have to add it to this file.
I created a celery task that runs every 1 minute (planning to change to 5 min) that reads this file, checks if it contains longcodes not yet in the db, and adds them.

I added code to force sending SMS through pinpoint if a service has a reply to/longcode set.
I also disabled the code that switches provider on slow sends, so that if a service has a longcode it will only send through pinpoint.

I tested everything locally and I am able to assign a longcode to a service, send from that longcode successfully. And I set up a new service w/ no longcode, and that sends successfully from a different number with SNS. 

Still need to write tests.

smcmurtry:
This PR should now be good to go. When approved it should be merged before https://github.com/cds-snc/notification-admin/pull/477. Here is an updated summary of what is included:
- pinpoint client
- 2 new environment variables: `AWS_PINPOINT_APP_ID`, `AWS_PINPOINT_KEYWORD`. These will need to be added to notification-staging-tf before merging this
- disabled the `switch-current-sms-provider-on-slow-delivery` celery job
- added an endpoint and query to allow notification-admin to add sms numbers to the `inbound_numbers` db table
- modified the logic in the `provider_to_use` function in app/delivery/send_to_providers.py so that, if an inbound sms number is set (`sender`), pinpoint is used to send the sms regardless of priority.
- db migration to add pinpoint to the providers db table
- lots of new tests and fixes for existing tests

app/config.py
```diff
@@ -220,11 +222,11 @@ class Config(object):
             'schedule': timedelta(minutes=66),
             'options': {'queue': QueueNames.PERIODIC}
         },
-        'switch-current-sms-provider-on-slow-delivery': {
-            'task': 'switch-current-sms-provider-on-slow-delivery',
-            'schedule': crontab(),  # Every minute
-            'options': {'queue': QueueNames.PERIODIC}
-        },
+        # 'switch-current-sms-provider-on-slow-delivery': {
```
timarney, line 225: @smcmurtry let's pull this out vs comment out

timarney:
@smcmurtry 

The code is looking good.  I just want to get clarity so we're all on the same page.  

We need some back story / history tracked here i.e. as @sastels comes on board. 

For tracking purpose can you outline here what is affected with the current setup / how or what needs to change with the current setup. Are the accounts setup enough to start using this.

i.e. For clarity all around 
- Can we merge this now or do we need to wait.
- if we merge this what could break
- are we still using SNS
- what happens if we don't assign / create any long codes 
- does every service need a long code
- what are we doing for slowing down sends
- what needs to be in place outside of merging this before the admin
- what needs to be setup in the account i.e. STOP messages etc...

If you don't know reach out with the pieces you don't know about.

Also @maxneuvians @willeybryan @obrien-j jump in as you see fit.

smcmurtry:
Thanks @timarney, let me know if the following addresses your request:

- We need to add 2 env variables to notification-staging-tf before merging, and also to notification-tf before updating prod
- If we merge this after the env variables are added, nothing should break. But this is touching the code that selects the sending provider, and removes the code that switches providers on slow send. So if anything is wrong there, that could effect all sends going through Notify.
- If a platform admin bumps up Pinpoint to the top priority, that would stop sms from being delivered for any service without an inbound sms assigned. 
- We are still using SNS for sending sms, except when a service has an "inbound sms" assigned to it. For that we use Pinpoint.
- if we don't assign or create any longcodes, there won't be any inbound SMS numbers available to assign to services. All sms will be sent through SNS as before
-  "Does every service need a long code?" No - until we have rate limiting in place, we should not assign inbound sms to any services. Otherwise we could exceed the rate limits enforced by telcos and get our longcodes blocked. Once that code is in place we could start with 1 service and see how that goes before assigning more.
- STOP/HELP messages are configurable in the Pinpoint dashboard - it looks like they will be the same messages across all our long codes for now

timarney:
@maxneuvians @obrien-j 

Any reasons you see not to merge this?

https://github.com/cds-snc/notification-api/pull/772#issuecomment-621326496

> If we merge this after the env variables are added, nothing should break. But this is touching the code that selects the sending provider, and removes the code that switches providers on slow send. So if anything is wrong there, that could effect all sends going through Notify.

> if we don't assign or create any longcodes, there won't be any inbound SMS numbers available to assign to services. All sms will be sent through SNS as before

maxneuvians:
Can be merged I think - just need to test in staging thoroughly before going to prod

timarney:
Approving and moving to staging for review.

Issue #763, opened 2020-04-15
Improve the dev experience of previewing template / branding markup


brdunfield:
Towards #737 , #744 
Pre-requisite: #759 

**Why:**
As detailed in #744 , the process of updating templates is very long and complex. Part of this complexity comes from being unsure of how the template / branding markup will actually render in an email. This can't be tested until very late in the process, and if it's not right we have to start over.

This will partially be addressed by #759 , however we can further improve the dev process by adding a way for devs to more easily preview their template changes.

**Ideas:**
- Expose a route that renders template markup in the browser. (Tim has done this in the past and it has helped, but doesn't always correspond exactly to how templates render in an email)
- ~~Add a way to send a test email with new branding markup. This would be a platform admin only function, which would allow the dev to send themselves an email to quickly preview a new or edited branding before it replaces existing markup in whatever form it's saved from #759~~ Devs can send an email using local jinja as of #796

brdunfield:
Closing this as devs can now send an email using local jinja as of #796 . I don't think putting work into exposing a new route would be worth it.

PR #796, opened 2020-04-22
Add ability to use local jinja template for html emails

### LINKED ISSUE #759: Simplify template setup process
Towards #744 
Pre-requisite: #758 

**Story:**
Once we understand the flow of template markup ( #758 ) we can make a decision on how exactly to simplify the template setup process.

**Concerns:**
We need to limit the ability for a dev to push up bad markup which affects live services without there being some kind of check or test.

**Solution:**
- Leave the existing flow unchanged, but add the ability to test using local markup (not pulled from notification-utils). That way local markup can be iterated until satisfactory, and only pushed through once it's correct.

brdunfield:
Closes #759 and possibly #763 

This PR, in tandem with the matching PRs in notification-utils and notification-admin, allows devs to use local in-project jinja for sending emails. This allows us to easily preview our markup changes without having to have first pushed them through the notification-utils repo.

Tests will fail until notification-utils has been updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,22 @@ make freeze-requirements
 
 `requirements.txt` should be committed alongside `requirements-app.txt` changes.
 
+## Using Local Jinja for template changes
+
+HTML Email Templates (`email_template.jinja2`) are pulled in from the [notification-utils](https://github.com/cds-snc/notification-utils) repo. To test jinja changes locally (without needing to update the upstream), follow this procedure:
+
+1. Create a `jinja_templates` folder in the project root directory, and put a .gitignore in it so nothing is tracked or pushed to this repo.
+
+2. Copy `./notification-utils/jinja_templates/email_template.jinja2` from [notification-utils](https://github.com/cds-snc/notification-utils) into the `jinja_templates` folder created in step 1
+
+3. Set a new .ENV variable: `USE_LOCAL_JINJA_TEMPLATES=True`
+
+4. Make markup changes, and see them locally!
+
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -118,6 +118,22 @@ make freeze-requirements
 
 `requirements.txt` should be committed alongside `requirements-app.txt` changes.
 
+## Using Local Jinja for template changes
```
timarney, line 121: Maybe 

## Using Local Jinja for testing template changes

Just want to be 100% clear this is for local testing only

README.md
```diff
@@ -118,6 +118,22 @@ make freeze-requirements
...
+
+HTML Email Templates (`email_template.jinja2`) are pulled in from the [notification-utils](https://github.com/cds-snc/notification-utils) repo. To test jinja changes locally (without needing to update the upstream), follow this procedure:
+
+1. Create a `jinja_templates` folder in the project root directory, and put a .gitignore in it so nothing is tracked or pushed to this repo.
```
timarney, line 125: The .gitignore can be commited but might want to note that it is.

application.py
```diff
@@ -33,3 +33,12 @@
 application = Flask('app')
 application.wsgi_app = ProxyFix(application.wsgi_app)
 create_app(application)
+
+if os.environ.get('USE_LOCAL_JINJA_TEMPLATES') == 'True':
+    print('')
```
timarney, line 38: Maybe add 

print(.env `USE_LOCAL_JINJA_TEMPLATES=True`)

As a pointer back to that file

Issue #761, opened 2020-04-15
Simplify template and branding variables


brdunfield:
Towards #737 
Pre-requisite: #759 

**Story:**
The Jinja markup for templates (`email_template.jinja2`) is difficult to read and causing branding problems.

**What:**
Issue will be resolved after:
- [x] Simplify / make more clear the 'Brand type' radio buttons on the admin page. What do they do? (Notification & Branding, Branding Only, Branding banner, No Branding).
- [x] Update variable names: 'gov_uk' features prominently in the markup

brdunfield:
Brand type radio buttons:
- 'Notification and Branding' = 'FIP and Custom Logo'
- 'Branding Only' = 'Custom Logo'
- 'Branding Banner' = 'Custom Logo on a background colour'
- 'No Branding' = no branding

PR #828, opened 2020-05-01
Template variable cleanup / rename

brdunfield:
This PR is in tandem with other PR's in notification-admin and notification-utils.

This addresses the first 2 bullet points of cds-snc/notification-api#761

Cleans up variable names, and makes it more clear how the radio buttons in the admin interface correspond to what the code does.

Tests will fail until notification-utils is merged.

This PR adds a new DB migration to support the variable renaming.


app/delivery/send_to_providers.py
```diff
@@ -26,7 +26,7 @@
     SMS_TYPE,
     KEY_TYPE_TEST,
     BRANDING_BOTH,
-    BRANDING_ORG_BANNER,
+    BRANDING_ORG_BANNER_NEW,
```
timarney, line 29: Let's add a comment here to outline why this was needed.  

Can link right to the line of the other file.

As a new dev to the project seeing `BRANDING_ORG_BANNER_NEW` would leave a pick question mark as to what is going on.

app/delivery/send_to_providers.py
```diff
@@ -224,8 +224,8 @@ def get_html_email_options(service):
 
     if service.email_branding is None:
         return {
-            'govuk_banner': True,
-            'brand_banner': False,
+            'fip_banner_english': True,
```
timarney, line 227: Let's add a comment to spell out FIP - again for folks who won't know what that is.

PR #835, opened 2020-05-06
Un-fixes the SES sending region

maxneuvians:
This reverts a previous fixing of the SES region to us-east-1 because the staging cluster is in ca-central-1 and there was no SES in ca-central-1 (See https://github.com/cds-snc/notification-api/pull/671). Now that ca-central-1 has SES we can use SES in the region that the cluster is in. The production cluster will default to us-east-1 and staging will default to ca-central-1.


sastels:
I'll take a look. Give me a reason to get the api running locally and look around in the AWS logs

PR #838, opened 2020-05-06
Template variable cleanup redux

maxneuvians:
Does DB persist in CI? Fixes https://github.com/cds-snc/notification-api/pull/828


maxneuvians:
Fixed in original 🎉

PR #876, opened 2020-05-19
Add "default_branding_is_french" column to organisations as well

### LINKED ISSUE #875: Organization-level branding fixes
The French FIP PR's have broken the ability to change an organization's branding to the defaults. This only affects CHANGING the branding, not branding that is already set.

Organizations will also need the new db column added to enable them to access the two branding options.

Note for self: Add a test for changing organization branding, or check the existing test if it exists.

brdunfield:
Closes #875 
Adds the 'default_branding_is_french' column to the organisation table in the DB as well, similarly to my last PR for services.

### DIFF
--- a/migrations/versions/0306c_branding_organisation.py
+++ b/migrations/versions/0306c_branding_organisation.py
@@ -1,0 +1,19 @@
+"""
+
+Revision ID: 0306c_branding_organisation
+Revises: 0306b_branding_french_fip
+Create Date: 2020-05-19 09:30:00
+
+"""
+from alembic import op
+
+
+revision = '0306c_branding_organisation'
+down_revision = '0306b_branding_french_fip'
+
+
+def upgrade():
+    op.execute('ALTER TABLE organisation ADD COLUMN default_branding_is_french boolean DEFAULT false')
+
+def downgrade():
+    op.execute('ALTER TABLE organisation DROP COLUMN default_branding_is_french')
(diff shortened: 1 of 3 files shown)


migrations/versions/0306c_branding_organisation.py
```diff
@@ -0,0 +1,19 @@
...
+
+revision = '0306c_branding_organisation'
+down_revision = '0306b_branding_french_fip'
+
+
+def upgrade():
+    op.execute('ALTER TABLE organisation ADD COLUMN default_branding_is_french boolean DEFAULT false')
+
+def downgrade():
+    op.execute('ALTER TABLE organisation DROP COLUMN default_branding_is_french')
```
sastels, line 19: just curious, did something generate this file for you?

sastels: where did the revision id come from?

brdunfield: Not really. It's not clearly defined. Originally, GDS just gave each new revision an incremental number (no letters). When we forked, we added revisions as letters to the most recent of their revisions (so we could continue to merge in their updates without everything lighting on fire). But eventually we stopped pulling in their migrations. We were up to 'm' or something, and we decided we could just start making our new increments. However I've added letters to these migrations mostly out of personal preference, because they're all related to this topic of branding / variables, and it made sense to me to just group them all under a single number and add letters for new migrations on this same topic..

PR #905, opened 2020-05-27
Change max CSV row limit for bulk sending services

smcmurtry:
I can't actually test this locally because csv jobs don't seem to work locally - they are just stuck in pending. Not sure why. So I'll need to test this on staging instead. I have verified that the API behaves as before - nothing will change until the env variables `HC_EN_SERVICE_ID` and `HC_FR_SERVICE_ID` are set.

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -3,7 +3,7 @@ import os
 from datetime import datetime, timedelta
 
 import pytz
-from flask import url_for
+from flask import current_app, url_for
 from sqlalchemy import func
 from notifications_utils.template import SMSMessageTemplate, WithSubjectTemplate, get_html_email_body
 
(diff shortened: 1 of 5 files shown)


app/utils.py
```diff
@@ -133,3 +133,14 @@ def update_dct_to_str(update_dct):
         str += "- {}".format(key.replace("_", " "))
         str += "\n"
     return str
+
+
+def get_csv_max_rows(service_id):
+    bulk_sending_services = [
+        current_app.config['HC_EN_SERVICE_ID'],
```
brdunfield, line 140: Does this fail gracefully or throw an error if these vars don't exist?

smcmurtry: Good point - I will add default values for these.

PR #910, opened 2020-05-29
Fix csv sending from the locally running api

smcmurtry:
Use an in memory broker for celery for the dev environment


smcmurtry:
Nevermind, csv send was already working.

The actual issue is that you can't tell whether the csv send has happened or not unless you include your own email address. Because the messages just keep their "sending" status, even after they've been sent:
Screen Shot 2020-05-29 at 11 14 30 AM

PR #919, opened 2020-06-03
Email two-factor now sends a 5 digit code instead of a clickable link

### LINKED ISSUE #895: GC Office 365 Migration Broke Email 2fa for GC staff M/L
**Steps to Reproduce**
- Invite your TBS email to a Notify service
- using your CDS user Set your TBS user to Email sign in
- Try to log in as your TBS user

**Expected**
- Clicking the link logs me in

**Actual**
- Href is destroyed by security software that opens the link before hand to scan it, causing the link to show up as expired to the user. 
- The user has to copy and paste it into the browser

**Proposed Solution**
- SMS texts a 5 digit # Code that you enter into a field in the web browser.
- Email a 5 digit number code instead. (Steam by Valve does email 2fa in this fashion)

**notes**
- Related issue https://github.com/cds-snc/we_are_platform/issues/214

brdunfield:
Closes #895 
Most of the changes were on the admin side. All I had to do here was switch the email 2fa to use the existing code generation that we use for SMS, and remove the old code.

Also added a new DB migration to update the email 2fa template. I did this as a migration to ensure the new variables / content filter through to staging / local dev environments etc. I did not add this as a new version, but the content will replace whatever the current version is on each of these systems. I did this to account for there possibly being different version numbers on staging vs prod, and not knowing what version the template will have anytime this migration is run.
A consequence of this is that all old versions of this template will also have the new content written into them, so we will effectively lose all older versions of the email 2fa template. I did archive the current prod content in the db migration, so if we downgrade, it will return to that. But anything before this will be lost.

If this is a dealbreaker, I will do more work to try and find a better way, but this seemed an acceptable loss to me.

### DIFF
--- a/app/user/rest.py
+++ b/app/user/rest.py
@@ -242,13 +242,15 @@ def verify_user_code(user_id):
     if user_to_verify.failed_login_count >= current_app.config.get('MAX_VERIFY_CODE_COUNT'):
         raise InvalidRequest("Code not found", status_code=404)
     if not code:
-        # only relevant from sms
         increment_failed_login_count(user_to_verify)
         raise InvalidRequest("Code not found", status_code=404)
-    if datetime.utcnow() > code.expiry_datetime or code.code_used:
+    if datetime.utcnow() > code.expiry_datetime:
         # sms and email
         increment_failed_login_count(user_to_verify)
         raise InvalidRequest("Code has expired", status_code=400)
+    if code.code_used:
+        increment_failed_login_count(user_to_verify)
+        raise InvalidRequest("Code has already been used", status_code=400)
 
     user_to_verify.current_session_id = str(uuid.uuid4())
     user_to_verify.logged_in_at = datetime.utcnow()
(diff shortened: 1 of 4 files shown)


app/user/rest.py
```diff
@@ -289,6 +290,7 @@ def send_user_sms_code(user_to_send_to, data):
 
     secret_code = create_secret_code()
     personalisation = {'verify_code': secret_code}
+    print(secret_code)
```
sastels, line 293: debugging while writing the PR?

tests/app/conftest.py
```diff
@@ -847,10 +847,10 @@ def email_2fa_code_template(notify_db, notify_db_session):
         content=(
             'Hi ((name)),'
             ''
-            'To sign in to GOV.​UK Notify please open this link:'
+            'To sign in to Notify please open this link:'
             '((url))'
```
sastels, line 851: Should this test / template also change?

app/user/rest.py
brdunfield: Yes, interesting that tests didn't catch that 🤔

tests/app/conftest.py
brdunfield: I don't think it needs to totally match the live one, but yes I'll modify it so it's clear it's using a code and not a link (to avoid confusion)

PR #980, opened 2020-06-19
Update some error messages for translation

### LINKED ISSUE #953: FR Bug: Security code error message
Screen Shot 2020-06-10 at 1 51 39 PM

Code not found = **Code non trouvé**

How might we reveal more of these error messages that remain without translation rather than stumble across them?

brdunfield:
Just updating a couple error strings.

I split an error message into multiple keys to be strung back together on the admin side to account for a variable in the middle.

Closes #953

### DIFF
--- a/app/dao/users_dao.py
+++ b/app/dao/users_dao.py
@@ -157,7 +157,8 @@ def get_user_and_accounts(user_id):
 @transactional
 def dao_archive_user(user):
     if not user_can_be_archived(user):
-        msg = "User cannot be removed from service. Check that all services have another team member who can manage settings"
+        msg = "User cannot be removed from service. "\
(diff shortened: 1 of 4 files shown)


app/dao/users_dao.py
```diff
@@ -157,7 +157,8 @@ def get_user_and_accounts(user_id):
 @transactional
 def dao_archive_user(user):
     if not user_can_be_archived(user):
-        msg = "User can’t be removed from a service - check all services have another team member with manage_settings"
+        msg = "User cannot be removed from service. "\
+            "Check that all services have another team member who can manage settings"
```
sastels, line 161: just curious - why split this line?

Issue #962, opened 2020-06-16
Update api and remaining admin references to use 'safelist'


brdunfield:
Last year some work was done to remove references in the admin interface to 'whitelist' and instead call it a 'safelist' https://github.com/cds-snc/notification-admin/pull/98

While this fixed the front-end visible interface, there remains a large amount of references to 'whitelist' within the api repo (and a few stragglers in the admin repo). We should update these.

This is probably a M in size

sastels:
@brdunfield @willeybryan should we also use "blocklist" instead of "blacklist" (or is there a better alternative)?

sastels:
@anikbrazeau Any thoughts on using "blocklisted" rather than blacklisted? In particular, we have a form validation error we present to the user:
"Password is blacklisted.","Le mot de passe est sur une liste noire."

What should the English / French here be?

willeybryan:
Yeah, Blocklist is then preferred nomenclature. 

Could the error be something like
"Password found on our block list"
"Nous trouvons ce mot de pass en notre liste d'interdiction"

anikbrazeau:
Yes! Let's absolutely change it.

If it's just for passwords:
(EN) Password blocklist = (FR) Liste de mots de passes bloqués

Or, how about describing that the password is not allowed instead?
- EN: "This password is not allowed. Try a different password."
- FR: "Ce mot de passe n'est pas autorisé. Essayez un autre mot de passe."

PR #1043, opened 2020-07-27
use new utils (43.0.1)

sastels:
re https://github.com/cds-snc/notification-api/issues/962

use new version (43.0.1) of notification-utils


sastels:
huh, after I removed the database-related changes there aren't really any code changes needed to use the new version of utils 🤷‍♂️ 
Probably still useful to have api running with the latest utils though.

PR #1057, opened 2020-07-31
rename service.whitelist property

sastels:
re https://github.com/cds-snc/notification-api/issues/962

change property service.whitelist to service.safelist
should have been part of https://github.com/cds-snc/notification-api/pull/1040 🤷

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -674,7 +674,7 @@ class ServiceSafelist(db.Model):
 
     id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
     service_id = db.Column(UUID(as_uuid=True), db.ForeignKey('services.id'), index=True, nullable=False)
-    service = db.relationship('Service', backref='whitelist')
+    service = db.relationship('Service', backref='safelist')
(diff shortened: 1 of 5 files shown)


app/models.py
```diff
@@ -674,7 +674,7 @@ class ServiceSafelist(db.Model):
 
     id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
     service_id = db.Column(UUID(as_uuid=True), db.ForeignKey('services.id'), index=True, nullable=False)
-    service = db.relationship('Service', backref='whitelist')
+    service = db.relationship('Service', backref='safelist')
```
brdunfield, line 677: What does this do? Does this need a DB upgrade?

sastels: no, we're still using the `service_whitelist` database table here, but saying that it's joined to a service via `service.safelist` (ie `service.safelist` now gives the list of corresponding rows from the `service_whitelist` table.

Possibly I'm splitting things up a bit too much, but figured I'd make the db migration PR as small as possible since I haven't done one of them before 😬

brdunfield: That makes sense. Thank you for the clarification!

PR #1076, opened 2020-08-07
Adding a celery pod type to separate out the SMS queue and rate limit it

### LINKED ISSUE #807: SMS load Balancer rate limiting
**Story**
When sending SMS I want it to come from a canadian number so that  can build a trusting communication relationship with my recipients.

**What?**
- Rate limit SMS long codes to 1per second per long code or less, like .75 to be safe. CWTA says max is 1/sec and more that that is blocked. 
- Stretch Goal, keep SNS up and running for our two factor and other testing. 

**Why?**
- Government comms people don't want to send from US numbers, they don't think it will be trusted
- SMS scams are everywhere
- SMS use cases are not more than a couple 100 at once so max speed isn't essential. 
- Helps use cases for embassy staff, geographically remote staff, Quarantine patients. 

**Done When?**
- We can safely offer a long code to a client for their use case and trust it.

brdunfield:
Closes #807 

The goal here is to add a single celery worker / pod to be in charge of sending SMS. We can limit that worker to only do a send 1x per second to comply with telcos. The reason why it needs to be on its own worker is a quirk of how celery's rate limiting works. The rate limit only applies to each worker individually. That is, if I set a rate limit of 1 per second and we have 5 workers, Notify will send up to 5 messages per second. By limiting SMS sends to a single worker, we can ensure that the rate limit is actually set at 1 per second. We need to do a bit of testing to ensure we are confident this is actually working and abiding by the rate limit.

will require https://github.com/cds-snc/notification-staging-tf/pull/4 

I have renamed several of the celery tasks with their queue prefixes, although I am not 100% sure if this is necessary.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -50,6 +50,13 @@ jobs:
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
       with:
         args: set image deployment.apps/celery-beat celery-beat=gcr.io/cdssnc/notify/api:${GITHUB_SHA::7} -n=notification-canada-ca --kubeconfig=/github/home/.kube/config
+    - name: Apply celery-sms
+      uses: docker://gcr.io/cdssnc/aws-kubectl:latest
+      env:
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      with:
+        args: set image deployment.apps/celery-sms celery-sms=gcr.io/cdssnc/notify/api:${GITHUB_SHA::7} -n=notification-canada-ca --kubeconfig=/github/home/.kube/config
     - uses: cds-snc/notification-pr-bot@master
       env:
         TOKEN: ${{ secrets.TOKEN }}
--- a/app/celery/celery.py
+++ b/app/celery/celery.py
@@ -47,3 +47,13 @@ class NotifyCelery(Celery):
         )
 
         self.conf.update(app.config)
+
+class NotifyCelerySMS(Celery):
+    def init_app(self, app):
+        super().__init__(
+            app.import_name,
+            broker=app.config['BROKER_URL'],
+            task_cls=make_task(app),
+        )
+
+        self.conf.update(app.config)
--- a/app/celery/celery.py
+++ b/app/celery/celery.py
@@ -47,13 +47,3 @@ class NotifyCelery(Celery):
         )
 
         self.conf.update(app.config)
-
-class NotifyCelerySMS(Celery):
-    def init_app(self, app):
-        super().__init__(
-            app.import_name,
-            broker=app.config['BROKER_URL'],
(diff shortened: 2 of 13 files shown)


scripts/run_celery_sms.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+set -e
+
+celery -A run_celery.notify_celery sms --loglevel=INFO --concurrency=1 -queue send-sms-tasks
```
brdunfield, line 5: the 'sms' here should just be 'worker', but with 'worker' it's the exact same command as in run_celery.sh, which means it just creates an identical instance. I'm not sure how to differentiate it.

app/config.py
```diff
@@ -22,7 +22,7 @@ class QueueNames(object):
     PERIODIC = 'periodic-tasks'
     PRIORITY = 'priority-tasks'
     DATABASE = 'database-tasks'
-    SEND_SMS = 'send-sms-tasks'
+    # SEND_SMS = 'send-sms-tasks'
```
brdunfield, line 25: This is the queue I need split out into the new worker

app/celery/celery.py
```diff
@@ -47,3 +47,13 @@ def init_app(self, app):
         )
 
         self.conf.update(app.config)
+
+class NotifyCelerySMS(Celery):
```
brdunfield, line 51: I assume here is where I'll somehow tell it to hook into the other worker, and define which queues I want?

app/celery/process_ses_receipts_tasks.py
```diff
@@ -216,7 +216,11 @@ def process_ses_results(self, response):
         self.retry(queue=QueueNames.RETRY)
 
 
-@notify_celery.task(bind=True, name="process-ses-smtp-results", max_retries=5, default_retry_delay=300)
+@notify_celery.task(
+    bind=True,
+    name="pservice-callbacks:rocess-ses-smtp-results",
```
sastels, line 221:
```suggestion
    name="service-callbacks:process-ses-smtp-results",
```

.github/workflows/docker.yaml
```diff
@@ -50,6 +50,13 @@ jobs:
...
+    - name: Apply celery-sms
+      uses: docker://gcr.io/cdssnc/aws-kubectl:latest
+      env:
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      with:
+        args: set image deployment.apps/celery-sms celery-sms=gcr.io/cdssnc/notify/api:${GITHUB_SHA::7} -n=notification-canada-ca --kubeconfig=/github/home/.kube/config
     - uses: cds-snc/notification-pr-bot@master
       env:
         TOKEN: ${{ secrets.TOKEN }}
```
sastels, line 62: why do we need these docker changes?

brdunfield: This ties into the PR here: https://github.com/cds-snc/notification-staging-tf/pull/4

It all ensures that the new command and new celery worker get run and deployed.

brdunfield: I will need the other PR approved in order to test this as well :)

PR #1083, opened 2020-08-13
delete unused environment variable SMS_INBOUND_WHITELIST

sastels:
re #962 

Delete the environment variable SMS_INBOUND_WHITELIST instead of renaming it, as the code doesn't use it after it's set, and we don't set it locally or in staging or prod.

App seems to run fine locally without it.


brdunfield (approved):

sastels:
💃

Issue #1016, opened 2020-07-07
Make account changes email more bilingual


sastels:
**steps to reproduce**
1) Navigate to your profile in French
2) Select to changez votre Nom
3) Lissez votre corriel. 

**Expected Behaviour**
- Email says Name changed, Nom Changed.

**Actual** 
See Screen shot, says name twice

**Notes**
Users get an email when details of their account change (directly from api). These emails have bilingual content, but the actual items that are changing are only in English:

![image](https://user-images.githubusercontent.com/8228248/86806522-f2aeb100-c046-11ea-83cd-33497d335239.png)

We should make the list bilingual as well. This would probably entail:
* making the template for that email take two `change_type` variables instead of one
* making a translations dict for the words that appear (name, password, ...)
* using the translations dict to set the `change_type` variables.

Note that currently the staging template is different from the production template.

anikbrazeau:
Translate ((change_type)) 

<img width="562" alt="Screen Shot 2020-07-09 at 7 35 24 AM" src="https://user-images.githubusercontent.com/38330843/87035245-0da03300-c1b7-11ea-8b2a-aba74f3604d3.png">

Are there other possible variables that are untranslated too?

anikbrazeau:
In "Automated You're Live Now" --> ((message limit)) variable format would be different in English and French
ex: EN = 25,000 // FR = 25 000

anikbrazeau:
In a similar vein, should we be linking directly to the FR URL in the FR part of the email? For example --> communiquez avec nous goes to FR Contact us page?

brdunfield:
I don't think we currently have separate URLs for languages. If the user's browser language is set to French, they will see the french page on load.

willeybryan:
On getting rid of ((change_type)) and it's impact on the user journey 

The history for this feature is that we added that to be compliant with Government of Canada security policies.
It also is generated when some one else, like an administrator edits a users information. Including a users system access or permissions.
The only time anyone has ever emailed us, I had changed their permissions in a service at the request of the service owner, and they emailed us.

It was helpful to know that their permissions had changed so I could be certain it was my actions that caused it.

According to the security control, we need to alert them that there has been a change, but not explicitly to what it was.

So we can remove the name, but for the two edge cases, I would want to communicate what happened. The edge cases being  an admin changes some one's permissions or an admin changes a users two factor authentication.

[...]

PR #1128, opened 2020-09-16
Make account changes email bilingual

AntoineAugusti:
Fixes https://github.com/cds-snc/notification-api/issues/1016

Related previous PR https://github.com/cds-snc/notification-api/pull/459

Local demo:
Screen Shot 2020-09-16 at 16 25 03

### DIFF
--- a/app/user/rest.py
+++ b/app/user/rest.py
@@ -637,7 +637,7 @@ def create_fido2_keys_user(user_id):
     id = uuid.uuid4()
     key = decode_and_register(cbor_data, get_fido2_session(user_id))
     save_fido2_key(Fido2Key(id=id, user_id=user_id, name=cbor_data["name"], key=key))
-    _update_alert(user)
+    _update_alert(user, changes={'security_key_created': None})
     return jsonify({"id": id})
 
 
@@ -706,7 +706,7 @@ def fido2_keys_user_validate(user_id):
 def delete_fido2_keys_user(user_id, key_id):
     user = get_user_and_accounts(user_id)
     delete_fido2_key(user_id, key_id)
-    _update_alert(user)
+    _update_alert(user, changes={'security_key_deleted': None})
     return jsonify({"id": key_id})
 
 
--- a/app/user/rest.py
+++ b/app/user/rest.py
@@ -140,7 +140,7 @@ def update_user_attribute(user_id):
     user_alert_dct = update_dct.copy()
     user_alert_dct.pop('blocked', None)
     user_alert_dct.pop('current_session_id', None)
-    if user_alert_dct:
+    if not updated_by and user_alert_dct:
         _update_alert(user_to_update, user_alert_dct)
 
     # Alert that team member edit user
--- a/app/user/rest.py
+++ b/app/user/rest.py
@@ -141,7 +141,7 @@ def update_user_attribute(user_id):
     user_alert_dct.pop('blocked', None)
     user_alert_dct.pop('current_session_id', None)
     if user_alert_dct:
-        _update_alert(user_to_update, update_dct_to_str(user_alert_dct))
+        _update_alert(user_to_update, user_alert_dct)
 
     # Alert that team member edit user
     if updated_by:
@@ -164,8 +164,7 @@ def update_user_attribute(user_id):
             personalisation={
                 'name': user_to_update.name,
                 'servicemanagername': updated_by.name,
-                'email address': user_to_update.email_address,
-                'change_type': update_dct_to_str(update_dct)
+                'email address': user_to_update.email_address
             },
             notification_type=template.template_type,
             api_key_id=None,
@@ -522,7 +521,7 @@ def set_permissions(user_id, service_id):
     change_dict = {service_key: service_id, "permissions": permission_list}
 
     try:
-        _update_alert(user, update_dct_to_str(change_dict))
+        _update_alert(user, change_dict)
     except Exception as e:
         current_app.logger.error(e)
 
@@ -605,10 +604,10 @@ def update_password(user_id):
         raise InvalidRequest(errors, status_code=400)
 
     update_user_password(user, pwd)
-    change_type = update_dct_to_str({'password': "password updated"})
+    changes = {'password': "password updated"}
 
     try:
-        _update_alert(user, change_type)
+        _update_alert(user, changes)
     except Exception as e:
         current_app.logger.error(e)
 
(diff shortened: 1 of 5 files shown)


app/user/rest.py
```diff
@@ -164,8 +164,7 @@ def update_user_attribute(user_id):
             personalisation={
                 'name': user_to_update.name,
                 'servicemanagername': updated_by.name,
-                'email address': user_to_update.email_address,
-                'change_type': update_dct_to_str(update_dct)
```
AntoineAugusti, line 168: Discovered that the `change_type` is not used in the templates associated with the method.

Can you double-check? Template IDs should be:
```
TEAM_MEMBER_EDIT_EMAIL_TEMPLATE_ID = 'c73f1d71-4049-46d5-a647-d013bdeca3f0'
TEAM_MEMBER_EDIT_MOBILE_TEMPLATE_ID = '8a31520f-4751-4789-8ea1-fe54496725eb'
```

app/utils.py
```diff
@@ -127,12 +127,36 @@ def email_address_is_nhs(email_address):
...
+    }
+
+    content = "\n"
     for key in update_dct:
-        str += "- {}".format(key.replace("_", " "))
-        str += "\n"
-    return str
+        try:
+            key_name = values[lang][key]
+        except KeyError:
```
AntoineAugusti, line 155: My implementation is: have values in both languages that I'm almost certain they will be used after carefully looking at the code, but provide a fallback if I missed keys.

The fallback is English attribute name + replace `_` to `_` (`email_address` to `email address`). The previous code was only doing this part.

migrations/versions/0310_account_change_type.py
```diff
@@ -0,0 +1,50 @@
...
+        'If you did not make this change, [contact us](((contact_us_url)) "contact us") immediately.',
+        '',
+        '___',
+        '',
+        "Les informations de votre compte d''utilisateur ont été modifiées : ((base_url)).",
+        '',
+        "Informations mises à jour : ((change_type_fr))",
+        '',
+        "Si vous n''avez pas effectué ce changement, [communiquez avec nous](((contact_us_url)) \"communiquez avec nous\") immédiatement.",
+    ]
```
AntoineAugusti, line 31: Used the template on staging, it was different in the local database.

Notable change is `change_type_en` + `change_type_fr`

tests/app/test_utils.py
```diff
@@ -70,11 +70,19 @@ def test_midnight_n_days_ago(current_time, arg, expected_datetime):
 def test_update_dct_to_str():
     test_dict = {
         "email_address": "test@test.com",
-        "auth_type": "sms"
+        "auth_type": "sms",
+        "dummy_key": "nope",
```
AntoineAugusti, line 74: Added this value to prove the fallback mechanism

app/utils.py
```diff
@@ -127,12 +127,36 @@ def email_address_is_nhs(email_address):
     ))
 
 
-def update_dct_to_str(update_dct):
-    str = "\n"
+def update_dct_to_str(update_dct, lang):
+    if lang not in ['EN', 'FR']:
+        raise NotImplementedError
```
AntoineAugusti, line 132: Is there a cleaner way to do this on the API side? We don't have supported lang + translation files here.

app/user/rest.py
willeybryan: Yes!
To find the templates you can go to Service 0, which handles GC Notify's Notifications
https://notification.alpha.canada.ca/services/d6aa2c68-a2d9-4437-ab19-3ae8eb202553/templates/[TEMPLATE_ID]

EDIT_EMAIL has variables ((servicemanagername)) ((name)) and ((email address)), So rather then saying your email was changed it says the new email they changed it too, this is funny because it will go to the new email.

EDIT_MOBILE has only ((servicemanagername)) and of course the number it goes it new phone number. 

So they are clear of the change type and this particular problem.

AntoineAugusti: I exactly did that but wanted to make sure I did not miss anything 💯

migrations/versions/0310_account_change_type.py
brdunfield: You used the staging template here? Can you confirm if this is the same as prod or not? If not, we'll want the prod one here.

app/utils.py
```diff
@@ -127,12 +127,36 @@ def email_address_is_nhs(email_address):
...
+
+    values = {
+        'EN': {
```
brdunfield, line 135: Have a look in rest.py - there's a set_permissions() function there that also sends a change alert when a users' service permissions change. There may be some extra values needed for that.

app/utils.py
brdunfield: Correct. Right now translations typically happen by passing a key to the admin which handles the translating. The API itself doesn't do enough translation to have a system for it.

Unless we passed these values from the admin (which could be a good idea given the way our logic is laid out, but would be a more complicated fix). Otherwise, this would be the best way.

brdunfield:
Paul just opened up this ticket on the admin repo:
https://github.com/cds-snc/notification-admin/issues/761

We could potentially roll this into this PR? Should be a pretty small fix.

migrations/versions/0310_account_change_type.py
AntoineAugusti: Confirmed it's the same in staging and prod 👌

app/utils.py
AntoineAugusti: Here is the current email sent when changing permissions. I guess it's not ideal but I don't see how we could improve it easily. The 🇫🇷   part is correct from a copywriting perspective

Screen Shot 2020-09-17 at 11 39 47

AntoineAugusti:
Fixed https://github.com/cds-snc/notification-admin/issues/761 in https://github.com/cds-snc/notification-api/pull/1128/commits/1dc35c956963eba59db93e415356d1ba8ec36045

brdunfield:
I am not sure if it was like this before this PR or not, but when I make a chance as a service manager to one of my team member's accounts (I changed their email) - the team member gets both the email that someone changed their email, and an email stating THEY changed their email.

AntoineAugusti:
> I am not sure if it was like this before this PR or not, but when I make a change as a service manager to one of my team member's accounts (I changed their email) - the team member gets both the email that someone changed their email, and an email stating THEY changed their email.

It's not a regression but I fixed it in https://github.com/cds-snc/notification-api/pull/1128/commits/6cd54178a0d78d8589e7224f1b5a42eb6c3aaa03. Reproduced it locally and fixed it

brdunfield (approved):
LGTM, it's working for me locally as intended 👍 🦑

tests/app/user/test_rest.py
```diff
@@ -372,6 +372,7 @@ def test_post_user_attribute_with_updated_by(
     assert json_resp['data'][user_attribute] == user_value
 
     if arguments:
+        assert mock_persist_notification.call_count == 1
```
AntoineAugusti, line 375: Was 2 calls before making the change. One "something on your profile changed" and another "an admin changed a thing". Making sure only the "admin changed a thing" is sent

AntoineAugusti:
Thanks for your careful and speedy review @brdunfield! I'll merge as soon as the build is green :shipit:

PR #1132, opened 2020-09-17
Bump utils to 43.0.2

AntoineAugusti:
To be merged after https://github.com/cds-snc/notification-utils/pull/60 is merged and released


brdunfield:
We probably want to bump admin as well, even if it's not using any of the new code - just so we don't forget.

AntoineAugusti:
Yep, actually it was needed it admin to fix the real bug! Here is the PR https://github.com/cds-snc/notification-admin/pull/764

brdunfield (approved):
👍

PR #1142, opened 2020-09-24
Adds a docker-compose file

PierrePaul:
https://trello.com/c/ZUQWhMtW/37-new-docker-compose-file

Adds a docker-compose file to an env closer to production in local.


docker-compose.yml
```diff
@@ -0,0 +1,96 @@
...
+    image: notification-api
+    restart: on-failure
+    build:
+      context: ./
+      dockerfile: ci/Dockerfile
+    environment:
+      - REDIS_URL=redis://redis:6379/0
+      - SQLALCHEMY_DATABASE_URI=postgres://postgres:postgres@postgres:5432/postgres
+    command: >
+      bash -c "rm /tmp/celery.pid; sh scripts/run_celery.sh"
```
AntoineAugusti, line 57: Curious why you need this? Also, don't you need a `rm -f` because this file may not exist?

AntoineAugusti (changes requested):
Why did you delete the `ci/cloudbuild-*.yaml` files? I think that they are required to push containers to GCloud

PierrePaul:
> Why did you delete the `ci/cloudbuild-*.yaml` files? I think that they are required to push containers to GCloud

From what I understand so far, the container image is built via Github actions and push to Google Cloud via another Github actions. Again, from what I understand/guessed, those cloudbuild files were used when the container image was built on Google Cloud instead of Github.

docker-compose.yml
PierrePaul: Good question! When testing I found out the pid file was still in there, but it doesn't make sense. The /tmp folder is not a docker volume and should not persist between container up/down. I will do more investigation on this.

PierrePaul:
Looks like the build is launched from github actions but it's actually made on google cloud ( via GoogleCloudPlatform/github-actions/setup-gcloud@master ) so I will add them back in.

AntoineAugusti:
> Looks like the build is launched from github actions but it's actually made on google cloud ( via GoogleCloudPlatform/github-actions/setup-gcloud@master ) so I will add them back in.

If it's using an action from https://github.com/GoogleCloudPlatform/github-actions that's fine, just want to make sure that we don't remove things that we actually use 👍 

If you're unsure, double-check with @maxneuvians

docker-compose.yml
PierrePaul: Alright, I would need more devices to make test, but under linux, when stopping and restarting a container, `/tmp` is not cleared.

As for your other question :

> Also, don't you need a rm -f because this file may not exist?

If I was using `&&` instead of `;`, yes. The `;` operator doesn't care what the output of the previous version is.

AntoineAugusti (approved):

Issue #1117, opened 2020-09-11
Adjust DMARC reporting of Notify


ClementineHahn:
Adjust Notify DMARC reporting so that forensic reports (ruf) don't go to CCCS. Leave the aggregate reports (rua) alone.

sastels:
resolved by https://github.com/cds-snc/dns/pull/137

AntoineAugusti:
@willeybryan While looking at code coverage, I discovered another place in the codebase where we set DMARC settings when creating an SMTP relay.

Do we need to update this piece as well? https://github.com/cds-snc/notification-api/blob/a07681ed50be7041a504ccb37a632ad604263ea1/app/smtp/aws.py#L40-L45

willeybryan:
@AntoineAugusti  Uh Yeah! good catch. If you take care of that you might need to ship your code review to @maxneuvians as he was the last and only? person to touch that bit of code.

obrien-j:
Yup, please strip the ruf from this line @AntoineAugusti and toss in an issue to check existing smtp dns records.

PR #1144, opened 2020-09-28
SMTP relay: remove DMARC's ruf setting

AntoineAugusti:
Follow up of https://github.com/cds-snc/notification-api/issues/1117


maxneuvians (approved):
Good catch!

PR #1147, opened 2020-09-29
Fill the requester's email in a branding request

AntoineAugusti:
**Story**
When users upload branding images we need to have the email address included so we know who to respond to.

**What?**
Branding requests currently come in like this https://assistance.cds-snc.ca/a/tickets/901 They need an email.

**Done When**
A branding Request has email in it.

Demo: (note the different recipient/request address - Freshdesk is the recipient in prod)
Screen Shot 2020-09-29 at 15 21 53

https://trello.com/c/0fGhyLwF/45-include-users-email-in-a-branding-request-notification

### DIFF
--- a/app/user/rest.py
+++ b/app/user/rest.py
@@ -478,7 +478,7 @@ def send_branding_request(user_id):
         recipient=to['email'],
         service=service,
         personalisation={
-            'email': to['email'],
+            'email': get_user_by_id(user_id=user_id).email_address,
             'serviceID': to['serviceID'],
             'service_name': to['service_name'],
             'filename': to['filename']


app/user/rest.py
```diff
@@ -478,7 +478,7 @@ def send_branding_request(user_id):
         recipient=to['email'],
         service=service,
         personalisation={
-            'email': to['email'],
+            'email': get_user_by_id(user_id=user_id).email_address,
```
AntoineAugusti, line 481: Previously the recipient and requester were filled with the same email address which resulted in this bug.

Here is what the template looks like.
Screen Shot 2020-09-29 at 15 24 04

brdunfield (approved):
🦑

PR #1153, opened 2020-10-13
Chore clean up providers list

PierrePaul:
Remove the following providers from the codebase : MMG, Firetext, Twilio, Loadtesting


PierrePaul:
> But we may also want to verify how inbound SMS currently work (or if they even do - if they don't it would be good to know we need to write code here) to make sure taking this out doesn't break anything.

I talked about this at the scrum, so let's keep a trace of it as well.
All the providers I have removed have the same "callback" logic when handling an incoming SMS message.
You essentially go into the provider admin interface and set an URI as callback. Whenever the provider get a SMS tied to your longcode, it calls your URI with the data.

Amazon SNS doesn't support incoming SMS.
Amazon Pinpoint does, but has a very different callback system. You actually need to have a long running thread that listen to a SNS topic (see https://docs.aws.amazon.com/pinpoint/latest/userguide/channels-sms-two-way.html#channels-sms-two-way-configure-enable ).

Since all the code/tests we have for incoming SMS is going to be vastly different from what we have right now (for the no longer used providers), I removed it.

AntoineAugusti:
Env variables to delete

- FIRETEXT_API_KEY
- FIRETEXT_INBOUND_SMS_AUTH
- FIRETEXT_URL
- LOADTESTING_API_KEY
- MMG_API_KEY
- MMG_INBOUND_SMS_AUTH
- MMG_INBOUND_SMS_USERNAME
- MMG_URL
- TWILIO_ACCOUNT_SID
- TWILIO_AUTH_TOKEN
- TWILIO_FROM_NUMBER

AntoineAugusti:
Took a look at this PR and this looks very good! Made a little more cleaning.

We removed inbound SMS support, something we don't use right now. As Pierre-Paul pointed out, the Pinpoint logic is very different, we'll need to write different code if we want to add back this feature.

brdunfield (approved):
👍 🦑

PR #1160, opened 2020-10-21
Auth with API key only

AntoineAugusti:
This PR adds an alternative authentication mechanism to JWT tokens: API keys only. It's limited in scope for now:
- we don't advertise it so we don't change error messages for now
- it's backend only, no UI changes
- it's not straightforward to get only your API key from the UI for now (as we display key name + service ID + key)

To test it locally:
- create an API key, keep only the API key part (the displayed key is `key_name-service_id-api_key`
- hit an endpoint where you need an API key, for example, `curl -H 'Authorization: ApiKey-v1 $api_key' http://localhost:6011/notifications`

Authentication is well covered by tests. `pytest -k 'test_authentication.py' --maxfail=1 --ff -v` reports 40 tests executed.

Trello card: https://trello.com/c/xXoa6LwD/83-spike-what-is-needed-to-remove-jwt-from-api-auth

### DIFF
--- a/app/authentication/auth.py
+++ b/app/authentication/auth.py
@@ -8,9 +8,11 @@ from sqlalchemy.orm.exc import NoResultFound
 from app.dao.api_key_dao import get_api_key_by_secret
 from app.dao.services_dao import dao_fetch_service_by_id_with_api_keys
 
+JWT_AUTH_TYPE = 'jwt'
+API_KEY_V1_AUTH_TYPE = 'api_key_v1'
 AUTH_TYPES = [
-    ('Bearer', 'jwt'),
-    ('ApiKey-v1', 'api_key_v1'),  # TODO decide if ApiKey-v1 scheme is a good scheme name
+    ('Bearer', JWT_AUTH_TYPE),
+    ('ApiKey-v1', API_KEY_V1_AUTH_TYPE),
 ]
 
 
@@ -61,7 +63,7 @@ def requires_admin_auth():
     request_helper.check_proxy_header_before_request()
 
     auth_type, auth_token = get_auth_token(request)
-    if auth_type != 'jwt':
+    if auth_type != JWT_AUTH_TYPE:
         raise AuthError("Invalid scheme: can only use JWT for admin authentication", 401)
     client = __get_token_issuer(auth_token)
 
@@ -76,7 +78,7 @@ def requires_auth():
     request_helper.check_proxy_header_before_request()
 
     auth_type, auth_token = get_auth_token(request)
-    if auth_type == 'api_key_v1':
+    if auth_type == API_KEY_V1_AUTH_TYPE:
         _auth_by_api_key(auth_token)
         return
     client = __get_token_issuer(auth_token)
--- a/app/authentication/auth.py
+++ b/app/authentication/auth.py
@@ -5,7 +5,13 @@ from notifications_utils import request_helper
 from sqlalchemy.exc import DataError
 from sqlalchemy.orm.exc import NoResultFound
 
+from app.dao.api_key_dao import get_api_key_by_secret
 from app.dao.services_dao import dao_fetch_service_by_id_with_api_keys
+
+AUTH_TYPES = [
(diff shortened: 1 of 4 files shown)


maxneuvians:
I really like this change - my hope if that this leads to the API endpoints becoming lambdas :)

AntoineAugusti:
Our plan isn't to go away from JWT token yet as it requires changes for all our users + maintaining our own clients 😬  But to offer this alternative auth method for people who can't use JWT tokens for a variety of reasons

app/authentication/auth.py
```diff
@@ -5,8 +5,16 @@
...
 from sqlalchemy.orm.exc import NoResultFound
 
+from app.dao.api_key_dao import get_api_key_by_secret
 from app.dao.services_dao import dao_fetch_service_by_id_with_api_keys
 
+JWT_AUTH_TYPE = 'jwt'
+API_KEY_V1_AUTH_TYPE = 'api_key_v1'
+AUTH_TYPES = [
+    ('Bearer', JWT_AUTH_TYPE),
+    ('ApiKey-v1', API_KEY_V1_AUTH_TYPE),
```
AntoineAugusti, line 15: Decide if this is a good auth type. People will need to put this in the `Authorization` header (`Authorization: ApiKey-v1 $api_key`)

app/authentication/auth.py
```diff
@@ -36,12 +44,15 @@ def get_auth_token(req):
...
+    for el in AUTH_TYPES:
+        scheme, auth_type = el
+        if auth_header.lower().startswith(scheme.lower()):
+            token = auth_header[len(scheme) + 1:]
+            return auth_type, token
 
-    return auth_header[7:]
+    # TODO: decide if error message stays the same.
+    # In fact we support another auth method than Bearer
+    raise AuthError('Unauthorized, authentication bearer scheme must be used', 401)
```
AntoineAugusti, line 55: Not sure if we want to keep this exact message or announce that we support other auth types here

app/authentication/auth.py
```diff
@@ -51,7 +62,9 @@ def requires_no_auth():
 def requires_admin_auth():
```
AntoineAugusti, line 62: `admin_auth` is for requests coming from the admin to the API. Some endpoints can only be called by admin frontends

app/dao/api_key_dao.py
```diff
@@ -29,6 +30,12 @@ def expire_api_key(service_id, api_key_id):
     db.session.add(api_key)
 
 
+def get_api_key_by_secret(secret):
+    return ApiKey.query.filter_by(
+        _secret=encryption.encrypt(str(secret))
```
AntoineAugusti, line 35: The `secret` column already has a unique index so this will be a fast lookup in the database

app/authentication/auth.py
brdunfield: This would be a good @willeybryan question

willeybryan: When would this message trigger if we remove the requirement to use it?

brdunfield (approved):
This LGTM code-wise. I'd like to get Bryan's thoughts on it before we merge!

app/authentication/auth.py
AntoineAugusti: This message is displayed when:
- the HTTP header `Authorization` is not set
- you put the HTTP header `Authorization` but not a supported auth type like `Bearer`  and `ApiKey-v1`. These schemes are not case-sensitive

willeybryan: Ah so,  AuthError, could respond with ('Unauthorized, authentication bearer scheme or apikey-v1 must be used', 401) and could still be accurate. 

Lets not touch that yet, I am not prepared to answer questions about it yet. Make sure we circle back when we launch it with content.

AntoineAugusti: Yes, that would be the correct error message in the future. But I thought it was good to hold from updating it for now and we want to silently launch it yet.

PR #1163, opened 2020-10-29
Change AWS region for Pinpoint

AntoineAugusti:
Our production is in `us-east-1` for now. We also need to change the `AWS_PINPOINT_APP_ID` env variable


AntoineAugusti:
Going to close this PR. @maxneuvians explained the reasoning behind using another region.

If we put a dedicated phone number in the same region/project, SNS will pick up this number randomly, which we don't want. The hack is to put dedicated phone numbers in another region/account so that they'll not be used by other services.

PR #1165, opened 2020-11-03
Endpoint to get delivered notifications stats by month

AntoineAugusti:
Needed by the new activity dashboard.

Would be good to ship before https://github.com/cds-snc/notification-admin/pull/810 so that the review app could call the new endpoint


brdunfield (approved):
👍 🦑

PR #1167, opened 2020-11-04
Store the provider's ID when sending SMS

AntoineAugusti:
If we want to track delivery statuses for SNS and Pinpoint, we should store the AWS ID in the `Notification` model, in the `reference` column. AWS will include this ID in the callback later and we'll use this information to update the notification status.

SNS and Pinpoint clients already return this ID, we need to store it.
https://github.com/cds-snc/notification-api/blob/6781341dfe03a26ac145228265517ab8b62d5f15/app/clients/sms/aws_sns.py#L53
https://github.com/cds-snc/notification-api/blob/6781341dfe03a26ac145228265517ab8b62d5f15/app/clients/sms/aws_pinpoint.py#L96

The code is very similar to the email part
https://github.com/cds-snc/notification-api/blob/6781341dfe03a26ac145228265517ab8b62d5f15/app/delivery/send_to_providers.py#L176-L186

### DIFF
--- a/app/delivery/send_to_providers.py
+++ b/app/delivery/send_to_providers.py
@@ -68,7 +68,7 @@ def send_sms_to_provider(notification):
 
         else:
             try:
-                provider.send_sms(
+                reference = provider.send_sms(
                     to=validate_and_format_phone_number(notification.to, international=notification.international),
                     content=str(template),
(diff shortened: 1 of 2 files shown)


brdunfield (approved):
LGTM 🦑

PR #1168, opened 2020-11-04
Change AWS region for SNS

AntoineAugusti:
This will only affect staging as production is on `us-east-1` at the moment.

Checked that pods in staging and production have the expected value for this env variable


brdunfield (approved):
🤞

PR #1169, opened 2020-11-04
Celery task for SNS callbacks

AntoineAugusti:
This PR adds a new Celery task to process SNS callbacks. This will be used to update the notification status of SMS: `delivered` or `failed`.

It's very similar to the SES logic in https://github.com/cds-snc/notification-api/blob/master/app/celery/process_ses_receipts_tasks.py.

The complete flow is : SNS -> CloudWatch Logs -> Lambda -> SQS -> Celery task [this PR]

PR #1170, opened 2020-11-05
Celery: register process_sns_receipts_tasks

AntoineAugusti:
Follow-up of https://github.com/cds-snc/notification-api/pull/1169

The Celery task was not properly registered. Importing the file does the trick ✨


brdunfield (approved):
🦑

PR #1172, opened 2020-11-09
docs: Added more details for the postgres setup

jimleroyer:
## Changes

* Added how to create the *postgres* user if not already existing. 
* Removed a few unnecessary env var setup from the shell profile instructions for `virtualenvwrapper`. It seems these were copied from its documentation but are only useful in the context of the tool's example. Please see: https://virtualenvwrapper.readthedocs.io/en/latest/

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,6 @@ if it did not, take a look here: https://github.com/pyenv/pyenv/issues/660
 6. Add the following to your shell rc file. ex: `.bashrc` or `.zshrc`
 
 ```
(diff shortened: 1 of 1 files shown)


AntoineAugusti (commented):
What do you think of moving this to the "Frequent problems" section? I'm not sure everyone has this error when setting up the app the first time

jimleroyer:
@AntoineAugusti Moved the documentation changes into the **Frequent problems** section.

AntoineAugusti (approved):

PR #1175, opened 2020-11-12
Add failed to NOTIFICATION_STATUS_TYPES_FAILED

AntoineAugusti:
We currently set the status to `failed` for SMS, which may be a deprecated type. We're going to change it soonish to temporary/permanent failure.

In the meantime, adding `failed` to this enum will make sure that notifications with the status set to `failed` show up in the UI/CSV exports


app/service/statistics.py
```diff
@@ -86,11 +91,9 @@ def create_zeroed_stats_dicts():
 def _update_statuses_from_row(update_dict, row):
     if row.status != 'cancelled':
         update_dict['requested'] += row.count
-    if row.status in ('delivered', 'sent'):
+    if row.status in NOTIFICATION_STATUS_SUCCESS:
         update_dict['delivered'] += row.count
-    elif row.status in (
-            'failed', 'technical-failure', 'temporary-failure',
-            'permanent-failure', 'validation-failed', 'virus-scan-failed'):
+    elif row.status in NOTIFICATION_STATUS_TYPES_FAILED:
```
brdunfield, line 96: 👍 This is much cleaner!

brdunfield (approved):
👍 🦑

PR #1177, opened 2020-11-14
Setup VSCode devcontainer for notification-api

jimleroyer:
@maxneuvians did great effort to setup a devcontainer previously. I am only following his foot steps to finish it.

# Intent

A reliable developer's environment where the issues are minimized because the setup is shared and tested. It's also a quick way to setup a complete ready-to-work IDE featuring the application's configuration done, database and VSCode plugins loaded.

# Changes

* Common CLI tools loaded into the Linux installation such as:
  * tldr, manpages
  * curl, unzip, netstat, ssh
  * flask, pip, python, psql client
  * [FiraCode NerdFont](https://github.com/ryanoasis/nerd-fonts/tree/master/patched-fonts/FiraCode) loaded into VSCode and the integrated terminal
  * [starship prompt](https://starship.rs/) for plenty of useful information at the prompt
* Loaded VSCode plugins
  * Python extension pack (Python, MagicPython, Jinja, Django, Visual Studio IntelliCode)
  * VSCode Docker
  * Pylance
  * Gitlens
  * VSCode Snyk support
  * Markdown support
  * PostgreSQL client plug-in
  * Google Cloud Tools plug-in
* Setup PostgreSQL with `postgres` user
  * `notification_api` schema loaded
  * Easy to connect with a client outside of the container via 5432 port. 
* Python dependencies are loaded for both dev and test requirements profiles.
* **Easy extension** of the environment setup via the entrypoint shell scripts.

# Tests

Many reloads of the devcontainer on my laptop. It definitely needs more testing with us all sharing this setup.


brdunfield:
I think @AntoineAugusti is a better reviewer than me on this - he worked with / reviewed Pierre Paul's recent work on this subject.

jimleroyer:
@brdunfield okay, I assigned you because you use vscode as well but there is no obligation. I will put Antoine as a reviewer then.

.devcontainer/DbDockerfile
```diff
@@ -0,0 +1,3 @@
+FROM postgres:13.0
+
+COPY db-notif-entrypoint.sh /docker-entrypoint-initdb.d/
```
AntoineAugusti, line 3: Why notif instead of notify?

.devcontainer/dev-notif-entrypoint.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
AntoineAugusti, line 1: Maybe rename notif to notify

.devcontainer/db-notif-entrypoint.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
AntoineAugusti, line 1: Maybe rename notif to notify

.devcontainer/devcontainer.json
```diff
@@ -1,18 +1,29 @@
...
+		"ms-azuretools.vscode-docker",
+		"ms-python.vscode-pylance",
+		"eamodio.gitlens",
+		"wholroyd.jinja",
+		"pmbenjamin.vscode-snyk",
+		"visualstudioexptteam.vscodeintellicode",
+		"yzhang.markdown-all-in-one",
+		"ms-ossdata.vscode-postgresql",
+		"googlecloudtools.cloudcode"
 	]
```
AntoineAugusti, line 28: Sounds like personal preferences to me. Maybe not the extensions, but definitely the font 😉  Can we leave some things out?

AntoineAugusti (changes requested):
Looks good! A few suggestion. I didn't try it out as I don't use VS Code personally. Could you ask another dev to try it out?

.devcontainer/DbDockerfile
jimleroyer: Good question, I'll change it.

.devcontainer/devcontainer.json
jimleroyer: Yeah, some extensions in there are quite standard for development and Python. I am conflicted with the font because the default font provided by the integrated console lacks many characters that modern prompt makes use of, but __I can certainly remove the font for the vscode editor at large__. Although, I must say nerd font is pretty nice and popular among devs in general.

AntoineAugusti (approved):
LGTM! We may need to refine this as other developers use this

brdunfield:
I'm in the process of getting this running. I'll report back once I do

brdunfield:
A good idea would be a readme update with an explanation / pre-reqs / commands on how to use this.

With Jimmy's help we realized the reason it was not working for me was that I did not have the correct VSCode extension installed.

brdunfield (approved):
Seems to work as expected! 👍 

This would likely have saved me a lot of time a year ago when I was first onboarding on the project. I don't know if I will use it personally, as I'm quite set up in a different way, but this looks very useful!

Jimmy has mentioned he will do the Readme update as a separate PR before End of Week.

jimleroyer:
I will update the README this week with setup instructions (along with the docker compose fix as well). 👍  I will merge it for now as this can have value for new members of the team.

PR #1178, opened 2020-11-18
fix: repaired the docker-compose up setup

jimleroyer:
## Description

The `docker-compose up` series of commands did not work from a fresh installation. The database especially was missing creation and the `flask db upgrade` command. This should fix it.

I created new files in the `/local` folder. I didn't want to touch the existing Docker files in `/ci` as at least one is used by the GitHub workflows. 

So now a command such as `docker-compose up web` should bring the necessary ecosystem to work in tandem with the notification-admin project for example.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,96 +1,107 @@
-version: '2'
+version: '3'
 
 services:
-  postgres:
+  db:
     image: postgres:11.2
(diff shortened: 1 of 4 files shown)


AntoineAugusti (approved):
LGTM! Thanks for taking care of this

PR #1179, opened 2020-11-18
fix: The celery dependency should be in the docker-compose workers

jimleroyer:
## Changes

The celery binary was not installed anymore in the workers defined in the docker-compose.yml file. This might have been provoked by the latest changes I did, as I moved the pip install upon containers startup via entrypoint shell scripts. Hence simply adding entrypoint shell script for the workers was sufficient to fix this minor issue. 

## Test

Manually tested in my setup. Having the workers up and running allowed me to register locally in the setup as I received the expected registration emails.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -47,6 +47,7 @@ services:
     environment:
       - REDIS_URL=redis://redis:6379/0
       - SQLALCHEMY_DATABASE_URI=postgres://postgres:chummy@db:5432/notification_api
+    entrypoint: local/scripts/notify-worker-entrypoint.sh
     command: >
       bash -c "sh scripts/run_celery_beat.sh"
     volumes:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -98,8 +101,9 @@ services:
     environment:
       - REDIS_URL=redis://redis:6379/1
       - SQLALCHEMY_DATABASE_URI=postgres://postgres:chummy@db:5432/test_notification_api
+    entrypoint: local/scripts/notify-worker-entrypoint.sh
     command: >
-      bash -c "./scripts/run_tests.sh"
+      bash -c "/app/scripts/run_celery.sh"
```
AntoineAugusti, line 106: Why are you running this command in the image `notification-api-tests`?

jimleroyer: Good catch, my mistake. A copy/paste gone wrong probably.

AntoineAugusti (approved):
:shipit:

PR #1185, opened 2020-12-03
Adjust rate limit for SMS

AntoineAugusti:
Added a new SMS task, SMS will now be:
- unthrottled and handled by Celery workers if they are not sent using a specific phone number
- throttled system-wide at 1 SMS/s by a single SMS Celery worker if they are sent using a dedicated long code phone number

This introduces a new Celery task and a new queue but doesn't require Kubernetes changes.

The code changes aren't too bad but I really want to test this extensively in staging as it is critical.

### DIFF
--- a/app/celery/provider_tasks.py
+++ b/app/celery/provider_tasks.py
@@ -14,32 +14,25 @@ from app.models import NOTIFICATION_TECHNICAL_FAILURE
 
 @notify_celery.task(
     bind=True,
+    name="deliver_throttled_sms",
+    max_retries=48,
+    default_retry_delay=300,
+    rate_limit="1/s",
+)
+@statsd(namespace="tasks")
+def deliver_throttled_sms(self, notification_id):
+    _deliver_sms(self, notification_id)
+
+
+@notify_celery.task(
+    bind=True,
     name="deliver_sms",
     max_retries=48,
     default_retry_delay=300,
-    rate_limit="1/s")
+)
 @statsd(namespace="tasks")
 def deliver_sms(self, notification_id):
-    try:
-        current_app.logger.info("Start sending SMS for notification id: {}".format(notification_id))
-        notification = notifications_dao.get_notification_by_id(notification_id)
-        if not notification:
-            raise NoResultFound()
-        send_to_providers.send_sms_to_provider(notification)
-    except Exception:
-        try:
-            current_app.logger.exception(
-                "SMS notification delivery for id: {} failed".format(notification_id)
-            )
-            if self.request.retries == 0:
-                self.retry(queue=QueueNames.RETRY, countdown=0)
-            else:
-                self.retry(queue=QueueNames.RETRY)
-        except self.MaxRetriesExceededError:
-            message = "RETRY FAILED: Max retries reached. The task send_sms_to_provider failed for notification {}. " \
-                      "Notification has been updated to technical-failure".format(notification_id)
-            update_notification_status_by_id(notification_id, NOTIFICATION_TECHNICAL_FAILURE)
-            raise NotificationTechnicalFailureException(message)
+    _deliver_sms(self, notification_id)
 
 
 @notify_celery.task(bind=True, name="deliver_email", max_retries=48, default_retry_delay=300)
@@ -67,6 +60,29 @@ def deliver_email(self, notification_id):
         except self.MaxRetriesExceededError:
             message = "RETRY FAILED: Max retries reached. " \
                       "The task send_email_to_provider failed for notification {}. " \
+                      "Notification has been updated to technical-failure".format(notification_id)
+            update_notification_status_by_id(notification_id, NOTIFICATION_TECHNICAL_FAILURE)
+            raise NotificationTechnicalFailureException(message)
+
+
+def _deliver_sms(self, notification_id):
+    try:
+        current_app.logger.info("Start sending SMS for notification id: {}".format(notification_id))
+        notification = notifications_dao.get_notification_by_id(notification_id)
+        if not notification:
+            raise NoResultFound()
+        send_to_providers.send_sms_to_provider(notification)
+    except Exception:
+        try:
+            current_app.logger.exception(
+                "SMS notification delivery for id: {} failed".format(notification_id)
+            )
+            if self.request.retries == 0:
+                self.retry(queue=QueueNames.RETRY, countdown=0)
+            else:
+                self.retry(queue=QueueNames.RETRY)
+        except self.MaxRetriesExceededError:
+            message = "RETRY FAILED: Max retries reached. The task send_sms_to_provider failed for notification {}. " \
                       "Notification has been updated to technical-failure".format(notification_id)
             update_notification_status_by_id(notification_id, NOTIFICATION_TECHNICAL_FAILURE)
             raise NotificationTechnicalFailureException(message)
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -72,7 +72,10 @@ from app.models import (
     SMS_TYPE,
     DailySortedLetter,
 )
-from app.notifications.process_notifications import persist_notification
+from app.notifications.process_notifications import (
+    persist_notification,
+    send_notification_to_queue,
+)
 from app.service.utils import service_allowed_to_send_to
 from app.utils import get_csv_max_rows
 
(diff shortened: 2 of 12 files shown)


app/celery/provider_tasks.py
```diff
@@ -12,34 +12,27 @@
 from app.models import NOTIFICATION_TECHNICAL_FAILURE
 
 
+@notify_celery.task(
+    bind=True,
+    name="deliver_throttled_sms",
+    max_retries=48,
+    default_retry_delay=300,
+    rate_limit="1/s",
```
AntoineAugusti, line 20: This is the only difference between `deliver_throttled_sms` and `deliver_sms`

app/celery/provider_tasks.py
```diff
@@ -70,3 +63,26 @@ def deliver_email(self, notification_id):
...
+            )
+            if self.request.retries == 0:
+                self.retry(queue=QueueNames.RETRY, countdown=0)
+            else:
+                self.retry(queue=QueueNames.RETRY)
+        except self.MaxRetriesExceededError:
+            message = "RETRY FAILED: Max retries reached. The task send_sms_to_provider failed for notification {}. " \
+                      "Notification has been updated to technical-failure".format(notification_id)
+            update_notification_status_by_id(notification_id, NOTIFICATION_TECHNICAL_FAILURE)
+            raise NotificationTechnicalFailureException(message)
```
AntoineAugusti, line 88: Nothing new, just moved down the file

app/celery/tasks.py
```diff
@@ -72,7 +72,10 @@
     SMS_TYPE,
     DailySortedLetter,
 )
-from app.notifications.process_notifications import persist_notification
+from app.notifications.process_notifications import (
+    persist_notification,
+    send_notification_to_queue,
```
AntoineAugusti, line 77: Refactored this file to use the helper method `send_notification_to_queue`. If we change the logic later, we'll not have to hunt down the logic here, which was the same than in the helper

app/delivery/send_to_providers.py
```diff
@@ -211,12 +212,16 @@ def provider_to_use(notification_type, notification_id, international=False, sen
...
 
-    if sender is not None and notification_type == SMS_TYPE and sender[0] == "+":
-        return clients.get_client_by_name_and_type("pinpoint", notification_type)
+    if _sms_send_on_pinpoint(notification_type, sender):
+        return clients.get_client_by_name_and_type(PINPOINT_PROVIDER, notification_type)
 
     return clients.get_client_by_name_and_type(active_providers_in_order[0].identifier, notification_type)
 
 
+def _sms_send_on_pinpoint(notification_type, sender):
```
AntoineAugusti, line 221: I don't really like this name, wondering if we could rename it to something else. In fact, we're not forced to send these using Pinpoint now, the logic is really "is it going to be sent with a dedicated phone number"

app/notifications/process_notifications.py
```diff
@@ -127,9 +127,16 @@ def send_notification_to_queue(notification, research_mode, queue=None):
...
 
     if notification.notification_type == SMS_TYPE:
+        deliver_task = provider_tasks.deliver_sms
+        if send_to_providers._sms_send_on_pinpoint(
+            notification.notification_type,
+            notification.reply_to_text
+        ):
+            deliver_task = provider_tasks.deliver_throttled_sms
+            if not queue:
+                queue = QueueNames.SEND_THROTTLED_SMS
```
AntoineAugusti, line 137: Different task and queue is the SMS is sent with a dedicated phone number

scripts/run_celery.sh
```diff
@@ -2,4 +2,4 @@
 
 set -e
 
-celery -A run_celery.notify_celery worker --pidfile="/tmp/celery.pid" --loglevel=INFO --concurrency=4 -Q database-tasks,job-tasks,notify-internal-tasks,periodic-tasks,priority-tasks,reporting-tasks,research-mode-tasks,retry-tasks,send-email-tasks,service-callbacks
+celery -A run_celery.notify_celery worker --pidfile="/tmp/celery.pid" --loglevel=INFO --concurrency=4 -Q database-tasks,job-tasks,notify-internal-tasks,periodic-tasks,priority-tasks,reporting-tasks,research-mode-tasks,retry-tasks,send-sms-tasks,send-email-tasks,service-callbacks
```
AntoineAugusti, line 5: `send-sms-tasks` is back in the normal Celery worker

scripts/run_celery_sms.sh
```diff
@@ -2,4 +2,4 @@
 
 set -e
 
-celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=1 -Q send-sms-tasks
+celery -A run_celery.notify_celery worker --loglevel=INFO --concurrency=1 -Q send-throttled-sms-tasks
```
AntoineAugusti, line 5: Only the throttled SMS queue will be run by the SMS Celery worker

tests/app/celery/test_provider_tasks.py
```diff
@@ -5,41 +5,74 @@
...
-from app.celery.provider_tasks import deliver_sms, deliver_email
+from app.celery.provider_tasks import deliver_sms, deliver_throttled_sms, deliver_email
 from app.clients.email.aws_ses import AwsSesClientException
 from app.exceptions import NotificationTechnicalFailureException
 
 
+sms_methods = [
+    (deliver_sms, 'deliver_sms'),
+    (deliver_throttled_sms, 'deliver_throttled_sms'),
+]
```
AntoineAugusti, line 16: Used `@pytest.mark.parametrize` on every test method which was using `deliver_sms` before to prove that the 2 methods have the same logic

tests/app/notifications/test_process_notification.py
```diff
@@ -242,21 +242,24 @@ def test_persist_notification_increments_cache_if_key_exists(sample_template, sa
 
 
 @pytest.mark.parametrize((
-    'research_mode, requested_queue, notification_type, key_type, expected_queue, expected_task'
+    'research_mode, requested_queue, notification_type, key_type, reply_to_text, expected_queue, expected_task'
```
AntoineAugusti, line 245: The addition of the `reply_to_text` text on every line creates a huge diff. The real difference is new lines with a `reply_to_text` set, others have `None`

tests/app/notifications/test_process_notification.py
```diff
@@ -265,17 +268,24 @@ def test_send_notification_to_queue(
...
     key_type,
+    reply_to_text,
     expected_queue,
     expected_task,
     mocker,
 ):
-    mocked = mocker.patch('app.celery.{}.apply_async'.format(expected_task))
-    Notification = namedtuple('Notification', ['id', 'key_type', 'notification_type', 'created_at'])
+    if '.' not in expected_task:
+        expected_task = f'provider_tasks.{expected_task}'
```
AntoineAugusti, line 277: This dance is used to avoid long lines when setting up the test data, flake8 isn't happy otherwise

brdunfield:
So are we 100% committed / able to request long codes in a separate region? We should make sure that's very well documented in whatever process we have for requisitioning long codes for services..

AntoineAugusti:
@brdunfield We made sure to set up Pinpoint in [us-west-2](https://github.com/cds-snc/notification-terraform/blob/main/aws/common/pinpoint.tf) as we were doing before. This is how it is set in the codebase as well https://github.com/cds-snc/notification-api/blob/ea5e5f1835c0d250b571f935cd4930db1c0d94b3/app/clients/sms/aws_pinpoint.py#L14

We're forced to order long code in another region than ca-central-1

brdunfield:
Do we need to store the numbers in a CA region? Are there multiple CA regions?

AntoineAugusti:
@brdunfield I'm sure you knew this but no, no other regions in Canada. A single region, multiple availability zones at the moment.

Screen Shot 2020-12-08 at 17 09 21

app/delivery/send_to_providers.py
brdunfield: What if we rename it to something like "send_dedicated_number_sms"?
And PINPOINT_PROVIDER could be "DEDICATED_NUMBER_PROVIDER"? Longer but more clear?

app/notifications/process_notifications.py
brdunfield: This makes sense codewise but takes a minute to understand the order of how things get assigned in which conditions - perhaps a quick comment explaining the logic?

brdunfield:
So I guess the only concern is whether policy wise we are allowed to store the numbers outside of Canada.

brdunfield (approved):
LGTM on the code end. I'd just like confirmation about the policy here before we 🚢 

🦑 

cc @willeybryan or @sboots

willeybryan:
Looks good to me from a Policy and Legal Stand point. 
Budget looks good too.

PR #1186, opened 2020-12-03
fix: resolved dependency hell with latest pip resolver upgrade

jimleroyer:
## Changes

The `pip` tool changed its internal resolver for dependency resolution. The dependency management is stricter and some of our dependencies were found to be in conflict. To know more about these changes, [head over to this blog post](https://pyfound.blogspot.com/2020/11/pip-20-3-new-resolver.html).

I aligned a few libraries in the *notification-api* project with the ones in *notification-utils*. 

## Tests

I've cleaned my local virtual environments a few times, used the `--no-cache-dir` of `pip install` as much as I could and reinstall these changes couple of times, referring to my local *notification-utils* project from the *notification-admin* project.


AntoineAugusti (approved):
Thanks Jimmy for the hard work on this!

We didn't manage to update `moto` to the latest version but this is more than enough for now I think. Please make sure you tagged utils on `master` before merging

PR #1190, opened 2020-12-10
feat: increment statsd counter when services reach daily limit

jimleroyer:
## Before

Review this PR in notification-utils as this PR relies on it:
https://github.com/cds-snc/notification-utils/pull/72

## Issue

https://trello.com/c/62gscLJH/169-graph-the-services-going-over-the-daily-rate-limit

## Changes

Added a statsd metric when daily rate usage of services have been reached.

## Testing 

* Added two tests covering the two scenarios of when the service rate is within and outside the limit.

### DIFF
--- a/app/notifications/validators.py
+++ b/app/notifications/validators.py
@@ -34,7 +34,7 @@ def check_service_over_api_rate_limit(service, api_key):
             raise RateLimitError(rate_limit, interval, api_key.key_type)
 
 
-@statsd_catch(namespace="validators", counter_name="rate.limit.daily", exception=TooManyRequestsError)
+@statsd_catch(namespace="validators", counter_name="rate_limit.service_daily", exception=TooManyRequestsError)
 def check_service_over_daily_message_limit(key_type, service):
     if key_type != KEY_TYPE_TEST and current_app.config['REDIS_ENABLED']:
         cache_key = daily_limit_cache_key(service.id)
--- a/app/notifications/validators.py
+++ b/app/notifications/validators.py
@@ -7,6 +7,7 @@ from notifications_utils.recipients import (
     get_international_phone_info
 )
 from notifications_utils.clients.redis import rate_limit_cache_key, daily_limit_cache_key
+from notifications_utils.statsd_decorators import statsd_catch
 
 from app.dao import services_dao, templates_dao
 from app.dao.service_sms_sender_dao import dao_get_service_sms_senders_by_id
@@ -33,6 +34,7 @@ def check_service_over_api_rate_limit(service, api_key):
             raise RateLimitError(rate_limit, interval, api_key.key_type)
 
 
+@statsd_catch(namespace="validators", counter_name="rate.limit.daily", exception=TooManyRequestsError)
 def check_service_over_daily_message_limit(key_type, service):
     if key_type != KEY_TYPE_TEST and current_app.config['REDIS_ENABLED']:
         cache_key = daily_limit_cache_key(service.id)
--- a/tests/app/conftest.py
+++ b/tests/app/conftest.py
@@ -5,7 +5,6 @@ from datetime import datetime, timedelta
 import pytest
 import pytz
 import requests_mock
-from unittest.mock import Mock
 
 from flask import current_app, url_for
 from sqlalchemy import asc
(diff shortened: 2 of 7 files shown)


tests/app/conftest.py
```diff
@@ -1219,5 +1221,16 @@ def delete(endpoint, _expected_status=204, **endpoint_kwargs):
...
 
+@pytest.fixture(scope='function')
+def app_statsd():
+    current_app.config['NOTIFY_ENVIRONMENT'] = "test"
+    current_app.config['NOTIFY_APP_NAME'] = "api"
+    current_app.config['STATSD_HOST'] = "localhost"
+    current_app.config['STATSD_PORT'] = "8000"
+    current_app.config['STATSD_PREFIX'] = "prefix"
+    current_app.statsd_client = Mock()
+    return current_app
```
AntoineAugusti, line 1232: Wondering if you could reuse the fixture from utils to avoid duplicating this. I think that in other tests StatsD is mocked with `mocker` directly if you want to avoid adding a new fixture

app/notifications/validators.py
```diff
@@ -33,6 +34,7 @@ def check_service_over_api_rate_limit(service, api_key):
             raise RateLimitError(rate_limit, interval, api_key.key_type)
 
 
+@statsd_catch(namespace="validators", counter_name="rate.limit.daily", exception=TooManyRequestsError)
```
AntoineAugusti, line 37: I'm not convinced by the name. I fee like introducing 2 namespaces `rate.limit` is too much, "rate limit" should be a thing on its own. Maybe `rate_limit.service_daily`? Do we have other counters in the codebase related to rate limit?

AntoineAugusti (changes requested):
Good work 👍  A few things to consider. Please request another review when the utils PR has been merged and tagged and the CI passes here

tests/app/conftest.py
jimleroyer: Thanks, I am not super familiar with these Python test libraries so any hint is appreciated. I'll try to apply the changes.

jimleroyer: Regarding reusing the fixtures from *utils*: it seems unfortunately that the module shadowing occurring by both `conftest` files in the same `tests` folder will not allow this.

app/notifications/validators.py
jimleroyer: Other counters in *api* project so far:

```
'callback.{}.{}'
'callback.sendgrid.{}'
redis.daily_limit_cache_key(service.id)
'callback.sns.{notification_status}'
'callback.ses.{}'
'callback.ses-smtp.{}'
"user-agent.{}"
"clients.ses.error"
"clients.ses.error"
"clients.ses.error"
"clients.ses.success"
"clients.sendgrid.error"
"clients.sendgrid.success"
"clients.sns.error"
"clients.sns.error"
"clients.sns.success"
"clients.sns.error"
"clients.pinpoint.error"
"clients.pinpoint.error"
"clients.pinpoint.error"
"clients.pinpoint.success"
"clients.pinpoint.error"
```

jimleroyer: Ah naming, one of these hard things.. I will make that change. 

Just one comment to make sure we agree on metrics naming and goals around it.

I want to represent hierarchies of rates using many metrics that we'll be able to catch with this annotation. For example, in the same *validators* module, we could have right now `rate.limit.api` for the `check_service_over_api_rate_limit` method.  

We should be careful in how we name these because we can perform aggregate operations on hierarchies of metrics (which the parent-child associations are represented with `.`s), so that all families of business related metrics should be grouped together.

That said, I am not sure the name I got is the right one... most probably not and we'll not get this right the first time. I would say let's review all these names when we start leveraging statsd and metrics dashboard in a robust manner.

tests/app/notifications/test_validators.py
```diff
@@ -141,6 +141,41 @@ def test_check_service_message_limit_in_cache_over_message_limit_fails(
...
+    mocker.patch('app.notifications.validators.redis_store.set')
+
+    # When
+    service = create_service(notify_db, notify_db_session, restricted=True, limit=4)
+    check_service_over_daily_message_limit("normal", service)
+
+    # Then
+    app_statsd.statsd_client.incr.assert_not_called()
+
+
```
RenDavidDimen, line 178: I really like the set up for these tests. Very easy to follow/understand! 👍

RenDavidDimen (approved):
Everything seems to check out for me. I really like your test set up 🙂

AntoineAugusti (approved):
Good to go 🚢 

You're right to point out that we should revisit StatsD metrics to make sure that they are properly named. I cleaned StatsD prefixes recently https://github.com/cds-snc/notification-api/pull/1183 (+ linked PRs).

Could you open a task in the backlog? 🙏

jimleroyer:
[Head over here for the task](https://trello.com/c/FbZPcyp0/268-review-statsd-metrics-in-api-admin-projects) opened in the product backlog.

jimleroyer:
I will wait for [review of this PR](https://github.com/cds-snc/notification-admin/pull/849) in the *admin* counterpart before merging.

PR #1193, opened 2020-12-14
Increase keepalive between Gunicorn and LB

AntoineAugusti:
In production only, in order to stop the 502 alarms we're seeing.

Long Slack thread explaining the rationale https://gcdigital.slack.com/archives/CNWA63606/p1607417483005100

Same as https://github.com/cds-snc/notification-admin/pull/852

### DIFF
--- a/gunicorn_config.py
+++ b/gunicorn_config.py
@@ -10,6 +10,14 @@ worker_connections = 256
 bind = "0.0.0.0:{}".format(os.getenv("PORT"))
 accesslog = '-'
 
+# See AWS doc
+# > We also recommend that you configure the idle timeout of your application
+# to be larger than the idle timeout configured for the load balancer.
+# > By default, Elastic Load Balancing sets the idle timeout value for your load balancer to 60 seconds.
+# https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#connection-idle-timeout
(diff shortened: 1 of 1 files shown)


gunicorn_config.py
```diff
@@ -10,6 +10,14 @@
 bind = "0.0.0.0:{}".format(os.getenv("PORT"))
 accesslog = '-'
 
+# See AWS doc
+# > We also recommend that you configure the idle timeout of your application
+# to be larger than the idle timeout configured for the load balancer.
+# > By default, Elastic Load Balancing sets the idle timeout value for your load balancer to 60 seconds.
+# https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#connection-idle-timeout
+if os.environ.get("NOTIFY_ENVIRONMENT", "") == "production":
```
AntoineAugusti, line 18: This env is set to this value in staging as well so we will be able to test it beforehand

AntoineAugusti:
Added an environment variable acting as a kill switch for this feature as it could cause troubles. It would be faster to revert in production if something came up: we would need to remove an environment variable on k8s and not revert a commit on admin/API + push a new Docker image + update k8s manifests with the new commit hash.

dorianjp (approved):
Let's test this in staging!

PR #1194, opened 2020-12-15
Adjust Celery rate limit for deliver_sms

AntoineAugusti:
The SNS rate limit is set at 20 messages per second but we asked AWS for a limit increase.

As we run 6 Celery instances with 4 workers, we need to set a rate limit at the task level if
we want to avoid hitting rate limits per second on AWS, even with increased limits.

Follow up of https://github.com/cds-snc/notification-api/pull/1185
Slack thread for more details https://gcdigital.slack.com/archives/CNWA63606/p1607979769008500

### DIFF
--- a/app/celery/provider_tasks.py
+++ b/app/celery/provider_tasks.py
@@ -12,6 +12,11 @@ from app.exceptions import NotificationTechnicalFailureException, MalwarePending
 from app.models import NOTIFICATION_TECHNICAL_FAILURE
 
 
+# Celery rate limits are per worker instance and not a global rate limit.
+# https://docs.celeryproject.org/en/stable/userguide/tasks.html#Task.rate_limit
+# This task is dispatched through the `send-throttled-sms-tasks`
+# queue. This queue is consumed by 1 Celery instance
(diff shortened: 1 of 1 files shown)


app/celery/provider_tasks.py
```diff
@@ -24,11 +30,18 @@ def deliver_throttled_sms(self, notification_id):
...
+# This queue is consumed by 6 Celery instances with 4 workers in production.
+# The maximum throughput is therefore 6 instances * 4 workers = 24 tasks per second
+# if we set rate_limit="1/s" on the Celery task
 @notify_celery.task(
     bind=True,
     name="deliver_sms",
     max_retries=48,
     default_retry_delay=300,
+    rate_limit="1/s",
 )
```
AntoineAugusti, line 45: With the rate limit addition, the 2 SMS tasks in this file have the same settings but a different name. We could have a single task, dispatched on different queues. I thought that it was important to limit code changes and get this out quickly in production. A refactor may be needed in a few weeks.

If you think this is a better idea to do it now, I'm happy to make this change.

brdunfield (approved):
🦑

PR #1198, opened 2020-12-23
Update simulated email addresses

AntoineAugusti:
Use a canada.ca domain and not a gov.uk one. Aligns with [the documentation](https://github.com/cds-snc/notification-documentation/commit/7509df47cf00c161b2c2af037b27b91f82ce17f4).

We don't know anyone using these fake email addresses to perform tests at the moment (we never communicated these were available) so the change should not be a problem.


RenDavidDimen (approved):
Looks good to be eh 🇨🇦

PR #1202, opened 2021-01-07
fix case insensitive name exists query

dorianjp:
fix, previous name exists query was case sensitive, changed to insensitive

### DIFF
--- a/tests/app/service/test_rest.py
+++ b/tests/app/service/test_rest.py
@@ -2471,11 +2471,10 @@ def test_is_email_from_unique_returns_200_if_unique(admin_request, notify_db, no
 
 
 @pytest.mark.parametrize('name, email_from',
-                         [("UNIQUE", "unique"),
-                          ("Unique.", "unique"),
+                         [("Unique.", "unique"),
                           ("**uniQUE**", "unique")
                           ])
-def test_is_service_name_unique_returns_200_with_name_capitalized_or_punctuation_added(
+def test_is_service_name_unique_returns_200_with_punctuation_added(
     admin_request,
     notify_db,
     notify_db_session,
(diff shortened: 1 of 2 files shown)


AntoineAugusti (changes requested):
Could you add a test case proving that the `lower` function is applied?

dorianjp:
> Could you add a test case proving that the `lower` function is applied?

@AntoineAugusti good call, I fixed related test case and added another to demonstrate lower works in two scenarios with and without the same service ID.

tests/app/service/test_rest.py
```diff
@@ -2470,23 +2471,55 @@ def test_is_email_from_unique_returns_200_if_unique(admin_request, notify_db, no
...
+                          ])
+def test_is_service_name_is_not_unique_returns_200_with_capitalization_while_unique_same_service(
+    admin_request,
+    notify_db,
+    notify_db_session,
+    name,
+    email_from
+):
+    service = create_service(service_name='unique', email_from='unique')
+    different_service_id = '111aa111-2222-bbbb-aaaa-111111111111'
```
AntoineAugusti, line 2507: You should be able to get a random UUID here with `uuid.uuid4()` (you can refactor in your other tests if you feel like it)

tests/app/service/test_rest.py
```diff
@@ -2470,23 +2471,55 @@ def test_is_email_from_unique_returns_200_if_unique(admin_request, notify_db, no
...
+                         [("UNIQUE", "unique"),
+                          ])
+def test_is_service_name_is_not_unique_returns_200_with_capitalization_while_unique_same_service(
+    admin_request,
+    notify_db,
+    notify_db_session,
+    name,
+    email_from
+):
+    service = create_service(service_name='unique', email_from='unique')
```
AntoineAugusti, line 2506: I think you're not using the `email_from` from your parameters here and you should

tests/app/service/test_rest.py
```diff
@@ -2470,23 +2471,55 @@ def test_is_email_from_unique_returns_200_if_unique(admin_request, notify_db, no
...
-def test_is_service_name_unique_returns_200_with_name_capitalized_or_punctuation_added(
+def test_is_service_name_unique_returns_200_with_punctuation_added(
     admin_request,
     notify_db,
     notify_db_session,
     name,
     email_from
 ):
-    create_service(service_name='unique', email_from='unique')
+    service = create_service(service_name='unique', email_from='unique')
```
AntoineAugusti, line 2484: I think you're not using the `email_from` from your parameters here and you should

tests/app/service/test_rest.py
dorianjp: yes, I just reused the existing hardcoded Id floating around in the tests prior

dorianjp: changed to uuid method

tests/app/service/test_rest.py
dorianjp: removed unused param

tests/app/service/test_rest.py
dorianjp: removed unused param

AntoineAugusti (approved):

dorianjp:
[Decouple Email and Service Name](https://trello.com/c/yeyTL5es/158-decouple-email-and-service-name)

PR #1203, opened 2021-01-07
Speed up docker build

AntoineAugusti:
Exact same changes than in https://github.com/cds-snc/notification-admin/pull/875


dorianjp (approved):
Now we just need 32 core parallel unit testing...

PR #1204, opened 2021-01-08
feat: disable Pinpoint, use only SNS for SMS

AntoineAugusti:
This PR:
- removes disable Pinpoint as an SMS provider
- clean up code related to Pinpoint
- sends SMS in another region when an SMS is sent with a custom sender

Because we'll be running with only a single SMS provider for a few weeks/months, some features are not relevant/don't work anymore. This is the case for the automatic switch of SMS providers in case a provider is down/slow. I did not delete the logic/tests for now as we really want to avoid having a single SMS provider for too long so I skipped these tests.

If you want to test this feature locally, you'll need to:
- run this branch locally
- add an inbound number in platform admin (I can give you a long code number on Slack)
- attach this inbound number to a service you own
- run the Celery SMS worker `scripts/run_celery_sms.sh`
- send a text, see that it's sent in us-west-2 with a specific phone number

I'm happy to share my screen and to demo this.

Infrastructure work done beforehand:
https://github.com/cds-snc/notification-terraform/pull/126
https://github.com/cds-snc/notification-terraform/pull/128
https://github.com/cds-snc/notification-terraform/pull/131

Trello card https://trello.com/c/pTa9pwcR/215-clean-up-amazon-pinpoint


app/clients/sms/aws_sns.py
```diff
@@ -27,17 +38,32 @@ def send_sms(self, to, content, reference, multi=True, sender=None):
...
+
+            # If sending with a long code number, we need to use another AWS region
+            # and specify the phone number we want to use as the origination number
+            send_with_dedicated_phone_number = self._send_with_dedicated_phone_number(sender)
+            if send_with_dedicated_phone_number:
+                client = self._long_codes_client
+                attributes['AWS.MM.SMS.OriginationNumber'] = {
+                    'DataType': 'String',
+                    'StringValue': sender
+                }
```
AntoineAugusti, line 59: If sending with a specific `sender`, switch the client to send texts from another AWS region and specify the sender as an attribute

app/clients/sms/aws_sns.py
```diff
@@ -56,3 +82,6 @@ def send_sms(self, to, content, reference, multi=True, sender=None):
             self.statsd_client.incr("clients.sns.error")
             self.current_app.logger.error("No valid numbers found in {}".format(to))
             raise ValueError("No valid numbers found for SMS delivery")
+
+    def _send_with_dedicated_phone_number(self, sender):
+        return sender and re.match(self.long_code_regex, sender)
```
AntoineAugusti, line 87: On the `Notification` model, the check is only `sender[0] == "+"` but I've been extra safe here

app/config.py
```diff
@@ -134,8 +134,7 @@ class Config(object):
     AWS_SES_SMTP = os.getenv("AWS_SES_SMTP", "email-smtp.us-east-1.amazonaws.com")
     AWS_SES_ACCESS_KEY = os.getenv('AWS_SES_ACCESS_KEY')
     AWS_SES_SECRET_KEY = os.getenv('AWS_SES_SECRET_KEY')
-    AWS_PINPOINT_APP_ID = os.getenv('AWS_PINPOINT_APP_ID')
-    AWS_PINPOINT_KEYWORD = os.getenv('AWS_PINPOINT_KEYWORD')
+    AWS_PINPOINT_REGION = os.getenv('AWS_PINPOINT_REGION', "us-west-2")
```
AntoineAugusti, line 137: Will had this new env on k8s, but even if it's not defined (by mistake) the fallback value will work

app/dao/provider_details_dao.py
```diff
@@ -30,13 +28,10 @@ def get_provider_details_by_identifier(identifier):
...
-    if identifier == SNS_PROVIDER:
-        alternate_provider = PINPOINT_PROVIDER
-    elif identifier == PINPOINT_PROVIDER:
-        alternate_provider = SNS_PROVIDER
-
-    return ProviderDetails.query.filter_by(identifier=alternate_provider).one()
+    # We currently run with a single SMS provider (SNS)
+    # so this method is not implemented and does not
+    # switch providers.
+    return get_provider_details_by_identifier(identifier)
```
AntoineAugusti, line 34: Rewrote the method to do nothing for the time being: returns the same provider as the current one. This does not perform an actual provider switch.

app/notifications/process_notifications.py
```diff
@@ -128,10 +127,7 @@ def send_notification_to_queue(notification, research_mode, queue=None):
 
     if notification.notification_type == SMS_TYPE:
         deliver_task = provider_tasks.deliver_sms
-        if send_to_providers._sms_send_on_pinpoint(
-            notification.notification_type,
-            notification.reply_to_text
-        ):
+        if notification.sends_with_custom_number():
```
AntoineAugusti, line 130: This method routes SMS to different queues/Celery tasks if they are sent with a dedicated long code number or not.

The logic has been introduced in https://github.com/cds-snc/notification-api/pull/1185 and is the condition is now updated. Previously it was "if sent with Pinpoint go to the throttled queue" now it's "if the SMS is sent with a custom sender, go to the throttled queue"

migrations/versions/0017_add_failure_types.py
```diff
@@ -14,10 +14,6 @@
 import sqlalchemy as sa
```
AntoineAugusti, line 14: Nothing to do with the current feature, fixing LGTM violations

migrations/versions/0313_disable_pinpoint_provider.py
```diff
@@ -0,0 +1,20 @@
+"""
```
AntoineAugusti, line 1: Database migration disabling Pinpoint as a provider in Platform admin

tests/app/celery/test_research_mode_tasks.py
```diff
@@ -37,21 +37,6 @@ def test_make_sns_callback(notify_api, rmock):
     assert 'mobile={}'.format(phone_number) in rmock.request_history[0].text
 
 
-def test_make_pinpoint_callback(notify_api, rmock):
```
AntoineAugusti, line 40: Wasn't used in tests strangely so deleted that because it has "Pinpoint" in it 👀

tests/app/clients/test_aws_sns.py
```diff
@@ -30,3 +30,24 @@ def test_send_sms_returns_raises_error_if_there_is_no_valid_number_is_found(noti
         aws_sns_client.send_sms(to, content, reference)
 
     assert 'No valid numbers found for SMS delivery' in str(excinfo.value)
+
+
+def test_send_sms_with_long_code_successful_returns_aws_sns_response(notify_api, mocker):
```
AntoineAugusti, line 35: Main new test for the SNS provider, making sure that if a sender is specified, the `_long_codes_client` is used and the origination number is specified

tests/app/notifications/test_process_notification.py
```diff
@@ -276,10 +275,6 @@ def test_send_notification_to_queue(
     if '.' not in expected_task:
         expected_task = f'provider_tasks.{expected_task}'
     mocked = mocker.patch(f'app.celery.{expected_task}.apply_async')
-    Notification = namedtuple(
-        'Notification',
-        ['id', 'key_type', 'notification_type', 'reply_to_text', 'created_at']
-    )
```
AntoineAugusti, line 282: Previously was using a tuple to avoid using the regular `Notification` model but the model is actually needed now that we have a new method `sends_with_custom_number`

app/clients/sms/aws_sns.py
dorianjp: not a huge deal, but the naming of sender is ambiguous at first glance, it sounds like the entity/service doing the sending and not the 'send as number'.

app/config.py
dorianjp: This is fine, but what do you think about certain env vars being required on app init? I've used this method in the past where we want to enforce specificity on the environment @ init as opposed to relying on implicit fallbacks.

app/dao/provider_details_dao.py
dorianjp: Do we want to do anything with the switching logic? In an exception condition, we end up running a bunch of logic to switch from:
https://github.com/cds-snc/notification-api/blob/a35f8fe77bc9f5a7fd8b32f567ee720ce241d091/app/delivery/send_to_providers.py#L82

Which may be redundant and set off an intended chain of events.

app/clients/sms/aws_sns.py
AntoineAugusti: Yeah, agreed. I didn't pick this name. `sender` but be understood as "the number sending the text message".

The call to this method is here https://github.com/cds-snc/notification-api/blob/a35f8fe77bc9f5a7fd8b32f567ee720ce241d091/app/delivery/send_to_providers.py#L73-L78 and other providers adhere to this interface

app/config.py
AntoineAugusti: Sounds like a good idea for a future refactor! For now, config files usually have fallback but I agree we shouldn't rely on those to make things work.

I'll draft up a k8s PR

app/dao/provider_details_dao.py
AntoineAugusti: Thanks for taking a good look at this, appreciated that you went down the road.

I wanted to leave as many things untouched for this path because I that the logic is good and we'll need to it. I think that there is a cost of deleting it now and introducing it again.

If you follow the underlying method calls https://github.com/cds-snc/notification-api/blob/a35f8fe77bc9f5a7fd8b32f567ee720ce241d091/app/dao/provider_details_dao.py#L66
and then https://github.com/cds-snc/notification-api/blob/a35f8fe77bc9f5a7fd8b32f567ee720ce241d091/app/provider_details/switch_providers.py#L22
https://github.com/cds-snc/notification-api/blob/a35f8fe77bc9f5a7fd8b32f567ee720ce241d091/app/provider_details/switch_providers.py#L39

So nothing will happen for real but we'll still log messages and add new rows to the providers' history table, which I think is important for logging purposes.

app/config.py
AntoineAugusti: @dorianjp Here is the k8s PR as promised https://github.com/cds-snc/notification-manifests/pull/150

dorianjp (approved):
sounds good

PR #1192, opened 2020-12-14
Run CI tests on Python 3.6 to 3.9

AntoineAugusti:
We're currently running Notify on Python 3.6 locally and in production. Let's prepare the move to 3.9 by running tests on various Python versions.

Similar PR than https://github.com/cds-snc/notification-admin/pull/851

PS: I'll update GitHub settings to make the "test / 3.6" workflow required instead of the generic "test" workflow. Without this change, other PRs will be blocked and I'll not be able to merge this

PR #1209, opened 2021-01-22
Upgrade Python 3.9.1 plus celery 5x

dorianjp:
Trello
https://trello.com/c/POlbTAih/238-prepare-the-upgrade-to-python-39

Upgrading app to python 3.9.1
This upgrade requires Kombu 5 which requires Celery 4+, we are opting to upgrade to celery 5x and some small dependency bumps.
Dockerfiles, CI and docs updated.

@AntoineAugusti suggesting to load test partly influenced from GDS decision to revert from Celery 4 to Celery 3: https://github.com/alphagov/notifications-api/pull/2142

Test plan:

Local:
* Tested API locally with ad-hoc and scheduled jobs OK
* Tested API locally with built Docker image with ad-hoc and scheduled jobs OK
* Local minor load testing OK

Staging:
* Gather baseline load testing data with various sustained loads (done, some data linked in Trello card)
* Other testing:  https://github.com/cds-snc/notification-api/pull/1192

Staging: (todo post merge)
* Test API with ad-hoc and scheduled jobs
* Perform load test and compare with baseline data for acceptability to continue to production or reversion.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -4,7 +4,7 @@ jobs:
   test:
     runs-on: ubuntu-latest
     container:
-      image:  python:3.6-stretch
+      image:  python:3.9-alpine
     services:
       postgres:
         image: postgres:10.8
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -7,7 +7,7 @@ jobs:
       image:  python:3.9-buster
     services:
       postgres:
-        image: postgres:10.8
+        image: postgres:11.8
         env:
           POSTGRES_USER: postgres
           POSTGRES_PASSWORD: postgres
@@ -17,13 +17,20 @@ jobs:
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
 
     steps:
-    - uses: actions/checkout@v1     
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.9
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.9
+    - name: Upgrade pip
+      run: python -m pip install --upgrade pip
     - uses: actions/cache@v2
       with:
         path: ~/.cache/pip
-        key: ${{ runner.os }}-pip-${{ hashFiles('requirements_for_test.txt') }}
+        key: ${{ runner.os }}-pip-${{ hashFiles('requirements.txt') }}
         restore-keys: |
           ${{ runner.os }}-pip-
-    - run: /bin/bash -c "pip3 install -r requirements_for_test.txt --no-cache-dir && make test"
+    - name: Run tests
+      run: /bin/bash -c "pip install -r requirements_for_test.txt && make test"
       env:
-        SQLALCHEMY_DATABASE_URI: postgresql://postgres:postgres@postgres:5432/test_notification_api
+        SQLALCHEMY_DATABASE_URI: postgresql://postgres:postgres@localhost:5432/test_notification_api
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -4,7 +4,7 @@ jobs:
   test:
     runs-on: ubuntu-latest
     container:
-      image:  python:3.9-alpine
+      image:  python:3.9-buster
     services:
       postgres:
         image: postgres:10.8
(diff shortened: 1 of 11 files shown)


.github/workflows/test.yaml
```diff
@@ -4,7 +4,7 @@ jobs:
   test:
     runs-on: ubuntu-latest
     container:
-      image:  python:3.6-stretch
+      image:  python:3.9-buster
```
dorianjp, line 7: stretch is not available as an image for 3.9 and alpine failed on CI test FYI, I'm guessing this is why stretch was used instead of alpine for the original 3.6 image

ci/Dockerfile.test
```diff
@@ -1,10 +1,10 @@
 # Heavily inspired from Dockerfile, this one also install requirements_for_test.txt
 
-FROM python:3.6-alpine
+FROM python:3.9-alpine
 
 ENV PYTHONDONTWRITEBYTECODE 1
 
-RUN apk add --no-cache bash build-base git gcc musl-dev postgresql-dev g++ make libffi-dev libmagic && rm -rf /var/cache/apk/*
+RUN apk add --no-cache bash build-base git gcc musl-dev postgresql-dev g++ make libffi-dev libmagic libcurl curl-dev && rm -rf /var/cache/apk/*
```
dorianjp, line 7: new celery/kombu requires curl

.github/workflows/test.yaml
AntoineAugusti: Would you be able to update this job to run tests without a container?

I did a similar thing in this PR https://github.com/cds-snc/notification-api/pull/1192 and bumped our Postgres version to match staging/production

README.md
```diff
@@ -25,13 +25,13 @@ On OS X:
```
AntoineAugusti, line 25: Thanks for updating this!

requirements.txt
```diff
@@ -84,7 +84,7 @@ greenlet==0.4.17
 importlib-metadata==3.1.1
 Jinja2==2.11.2
 jmespath==0.10.0
-kombu==3.0.37
+kombu==5.0.2
```
AntoineAugusti, line 87: Requirement is here https://docs.celeryproject.org/en/v5.0.5/whatsnew-5.0.html#kombu

requirements.txt
```diff
@@ -4,7 +4,7 @@
 # with package version changes made in requirements-app.txt
 
 cffi==1.14.3
-celery[sqs]==3.1.26.post2 # pyup: <4
+celery[sqs]==5.0.5
```
AntoineAugusti, line 7: I've read quickly CHANGELOGs for v4 and v5.

Do we need to upgrade setting names?
Do we need to change how we call Celery on the CLI? https://docs.celeryproject.org/en/v5.0.5/whatsnew-5.0.html#step-1-adjust-your-command-line-invocation in `scripts/*.sh`
Anything you saw there that doesn't affect us but you'd like to share?

AntoineAugusti (commented):
Thanks for doing this work!

A few questions/suggestions.

I noticed we have 3 Dockerfiles + 1 dev container. I'm wondering if we can delete at least a Dockerfile because I'm not sure we use everything in there.

requirements.txt
dorianjp: I should also mention celery points out a number of setting name changes which we use which are deprecated and will be pulled in v6. This can be wrapped up in another PR to minimize impacts.

CLI is the same based on how we are using it w containers.

dorianjp:
> Thanks for doing this work!
> 
> A few questions/suggestions.
> 
> I noticed we have 3 Dockerfiles + 1 dev container. I'm wondering if we can delete at least a Dockerfile because I'm not sure we use everything in there.

We should have a chat and include @jimleroyer on this, I believe he had some work invested in the .devcontainer folder w the dockerfile in there.

I used the ci/Dockerfile for my tests and do not currently use the devcontainer pieces. But I'll go forward with testing in staging as-is for now.

.github/workflows/test.yaml
dorianjp: Sure, put that code in there and I left the matrix with a single item list in to simplify future upgrade tests etc.

.github/workflows/test.yaml
```diff
@@ -17,13 +18,22 @@ jobs:
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
 
     steps:
-    - uses: actions/checkout@v1     
+    - name: Install libcurl
+      run: sudo apt-get install libssl-dev libcurl4-openssl-dev
```
dorianjp, line 22: now required without the container

AntoineAugusti (approved):
LGTM, thanks!

PR #1213, opened 2021-01-28
feat: update contact page URL

AntoineAugusti:
The contact page is not available at the URL `/support/ask-question-give-feedback` anymore for several months.

There is a redirection in place so the current link is not broken, fortunately.
https://github.com/cds-snc/notification-admin/blob/0237926c01c58d72a927e85aba58bb5b80dd20e9/app/main/views/contact.py#L155-L157

This PR is a small clean up.

Note that we don't have a clean and reliable way to link to the admin from the API as these are 2 different codebases. Maybe we'll move some URLs to the utils repo at some point? Food for thoughts.


dorianjp (approved):
f strings are the new hotness

PR #1214, opened 2021-01-28
feat: use Celery lowercase settings

AntoineAugusti:
Follow up of https://github.com/cds-snc/notification-api/pull/1209

Changed Celery settings name because those are going to be deprecated in 6.0. Config values stay in `app/config.py` but keys are remapped when passed to the Celery app.

Followed the documentation at https://docs.celeryproject.org/en/stable/userguide/configuration.html#new-lowercase-settings and warning messages when running Celery locally.

### DIFF
--- a/app/celery/celery.py
+++ b/app/celery/celery.py
@@ -46,4 +46,13 @@ class NotifyCelery(Celery):
             task_cls=make_task(app),
         )
 
-        self.conf.update(app.config)
+        # See https://docs.celeryproject.org/en/stable/userguide/configuration.html
+        self.conf.update({
+            'beat_schedule': app.config['CELERYBEAT_SCHEDULE'],
+            'imports': app.config['CELERY_IMPORTS'],
+            'task_serializer': app.config['CELERY_TASK_SERIALIZER'],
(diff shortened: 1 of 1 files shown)


dorianjp (commented):
I get one more warning when i run celery:
```
[2021-01-28 06:06:59,526: WARNING/MainProcess] /app/api.test.3.9.venv/lib/python3.9/site-packages/celery/app/utils.py:204: CDeprecationWarning: 
    The 'BROKER_URL' setting is deprecated and scheduled for removal in
    version 6.0.0. Use the broker_url instead

  deprecated.warn(description=f'The {setting!r} setting',


```

AntoineAugusti:
@dorianjp Thank you for checking that on `master`. I should have left a comment. The broker URL is already passed before changing the configuration, that's why I left it out in the `conf.update`

https://github.com/cds-snc/notification-api/blob/8988d0201a98ee03eab5b07995e47d5f6197484d/app/celery/celery.py#L42-L47

dorianjp (approved):
the vars are the same, only celery believes otherwise

PR #1219, opened 2021-02-01
feat: explicit read write split

AntoineAugusti:
See also the other PR started by @jimleroyer, we tried various routing strategies!

This PR introduces read/write split in a few chosen places across the API. This uses SQLAlchemy reader/writer binds. We made sure that it was possible to use only a writer if leaving out the `SQLALCHEMY_DATABASE_READER_URI` env variable.

Trello card: https://trello.com/c/5bRl0MOO/218-start-using-read-replicas-in-production


app/dbsetup.py
```diff
@@ -0,0 +1,68 @@
...
+            return get_state(self.app).db.get_engine(self.app, bind='writer')
+
+    def using_bind(self, name):
+        self._name = name
+        return self
+
+    @cached_property
+    def binds_setup(self):
+        binds = self.app.config['SQLALCHEMY_BINDS'] or {}
+        return all([k in binds for k in ['reader', 'writer']])
```
jimleroyer, line 45: I was thinking we could have constants for the `reader` and `writer` strings. Hence consumers of the API wouldn't make a mistake. We could even just have `using_master` and `using_reader` methods, knowing we won't have other binds in the near future anyway.

app/dbsetup.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class RoutingSQLAlchemy(SQLAlchemy):
+    def on_reader(self):
```
jimleroyer, line 49: Well look at that, I wrote too soon my previous comment. Looks good. 👍

jimleroyer (approved):

PR #1222, opened 2021-02-02
feat: add appropriate tags when creating a ticket on Freshdesk

AntoineAugusti:
When submitting the contact form:
- add tags (`z_skip_opsgenie` and `z_skip_urgent_escalation`) to the Freshdesk ticket when the email address/user is not a member of any live services
- don't add tags on the Freshdesk ticket if the email address/user is a member of at least a live service

These tags will help triage requests on Freshdesk and send different auto-replies.

Refactored to move the Freshdesk API client to its own class and use it in the controller. As I've introduced a new env variable `FRESH_DESK_PRODUCT_ID`, I'll open up Kubernetes pull-requests to set this up.

Trello card: https://trello.com/c/WpxAgDNK/262-revisit-freshdesk-auto-replies-to-match-notify-system-name-style-signature-voice-and-tone-in-en-and-fr


tests/app/user/test_rest.py
```diff
@@ -758,36 +758,60 @@ def test_send_already_registered_email_returns_400_when_data_is_missing(client,
...
-        headers=[('Content-Type', 'application/json'), auth_header])
+        data=json.dumps(data),
+        headers=[('Content-Type', 'application/json'), create_authorization_header()]
+    )
     assert resp.status_code == 204
 
-    notification = Notification.query.first()
-    mocked.assert_called_once_with(([str(notification.id)]), queue="notify-internal-tasks")
-    assert notification.reply_to_text == notify_service.get_default_reply_to_email_address()
+    mocked.assert_called_once_with(data | {'tags': ['z_skip_opsgenie', 'z_skip_urgent_escalation']})
```
AntoineAugusti, line 776: The important assertion is here: no live service implies 2 new tags that will be used by automation rules on Freshdesk

tests/app/user/test_rest.py
```diff
@@ -758,36 +758,60 @@ def test_send_already_registered_email_returns_400_when_data_is_missing(client,
...
     resp = client.post(
         url_for('user.send_support_email', user_id=str(sample_user.id)),
-        data=data,
-        headers=[('Content-Type', 'application/json'), auth_header])
+        data=json.dumps(data),
+        headers=[('Content-Type', 'application/json'), create_authorization_header()]
+    )
+    assert resp.status_code == 204
+
+    mocked.assert_called_once_with(data | {'tags': []})
```
AntoineAugusti, line 795: The important part is here: no tags are added if you're part of a live service

tests/app/clients/test_freshdesk.py
```diff
@@ -0,0 +1,41 @@
...
+            'description': 'my message',
+            'email': 'test@example.com',
+            'priority': 1,
+            'status': 2,
+            'tags': []
+        }
+
+        encoded_auth = base64.b64encode(b'freshdesk-api-key:x').decode('ascii')
+        json_matches = request.json() == expected
+        basic_auth_header = request.headers.get('Authorization') == f"Basic {encoded_auth}"
```
RenDavidDimen, line 22: Everything looks good! I'm just curious how these two lines get their values. More specifically, how the `request` object has the json and header values assigned to it.

RenDavidDimen (approved):
Looks good!

tests/app/clients/test_freshdesk.py
AntoineAugusti: This is some requests mock magic.

The line `additional_matcher=match_json` says that we should use the `match_json` method to check if the request that was sent was indeed the one we expected.

`Freshdesk.create_ticket` sends a request and `match_json` gets the sent request and analyzes it to see it was well-formatted (in our case auth headers + JSON body)

AntoineAugusti:
@jimleroyer Do you want to review this one?

app/user/rest.py
```diff
@@ -440,32 +440,18 @@ def send_already_registered_email(user_id):
...
-        auth=HTTPBasicAuth(API_KEY, "x")
-    )
+    data['tags'] = ['z_skip_opsgenie', 'z_skip_urgent_escalation']
+    try:
+        user = get_user_by_email(data['email'])
+        has_live_services = any([not s.restricted for s in user.services])
+        if has_live_services:
+            data['tags'] = []
+    except NoResultFound:
+        pass
```
jimleroyer, line 450: Are we the ones who wrote the `get_user_by_email` function? It seems weird to raise an exception because no results were returned.

app/clients/freshdesk.py
```diff
@@ -0,0 +1,34 @@
...
+                f"{current_app.config['FRESH_DESK_API_URL']}/api/v2/tickets",
+                json=ticket,
+                auth=HTTPBasicAuth(current_app.config['FRESH_DESK_API_KEY'], "x")
+            )
+            response.raise_for_status()
+
+            return response.status_code
+        except requests.RequestException as e:
+            content = json.loads(response.content)
+            current_app.logger.warning(f"Failed to create Freshdesk ticket: {content['errors']}")
```
jimleroyer, line 33: I know we treated previous cases of users who couldn't send freshdesk tickets as a high priority, so I would ask if this log message is the only mean to know if there was a failure? If yes, we should consider it for an `error` level log.

app/user/rest.py
```diff
@@ -440,32 +440,18 @@ def send_already_registered_email(user_id):
...
 
-    if response.status_code != 201:
-        print("Failed to create ticket, errors are displayed below")
-        content = json.loads(response.content)
-        print(content["errors"])
-        print("x-request-id : {}".format(content.headers['x-request-id']))
-        print("Status Code : {}".format(str(content.status_code)))
+    status_code = Freshdesk.create_ticket(data)
```
jimleroyer, line 452: I like the new client as this makes the code more readable with less information overload.

app/user/rest.py
AntoineAugusti: Nope, introduced here https://github.com/cds-snc/notification-api/commit/e055590b0730acb810cba034456c5306671b195d

`.one()` raises an exception, otherwise, we need to deal with a potential null object. I'm fine catching the exception

app/clients/freshdesk.py
AntoineAugusti: The line just after raises the `requests` exception 👌

jimleroyer: Yeah I didn't see it, that makes sense then. 🆗

app/user/rest.py
jimleroyer: Yeah, it seems prevalent in the Python community to use exceptions for control flow.

jimleroyer (approved):

PR #1223, opened 2021-02-02
feat: run get_api_key_by_secret on reader instances

AntoineAugusti:
Run the method `get_api_key_by_secret` on reader instances as introduced in #1219


dorianjp (approved):
rubber stamp (here)

PR #1224, opened 2021-02-03
Increased the default db connection pool for sqlalchemy

jimleroyer:
# Summary | Résumé

Increased the db connection pool for sqlalchemy from 5 to 25. We need more connection in the pool as we are now using the read replicas in both staging and production environment. We'd need at least 5 per database instance based on previous assumption and now we that we have 2 reader replicas adding in the production environment, we'd need at least 15. I set it at 25 for extra buffer, but we can change if someone comes up with a smart calculation that will convince us all of a new value. 🧞  

We should also change the environment variable with the agreed upon value. This will only take effect if no environment variable was defined. 

* Stress loading in the staging environment should tell us what's the load limit we can take. The staging environment though only has one read replica where the production environment has two. So we need to account that production will be more hungry for connections.

> Things that you (the submitter) want reviewers to pay very close attention to when they review this.

---

> Éléments auxquels vous (le demandeur) souhaitez que les réviseurs prêtent attention lorsqu’ils examineront votre demande.

> Are there any related issues or tangent features you consider out of scope for this issue that could be addressed in the future?

---

> Y a-t-il des questions connexes que vous considérez hors sujet et qui pourraient être abordées plus tard?


AntoineAugusti (changes requested):
Thank you for proposing this PR.

If we want to tweak this value, we need to rely on the env variable and not change the default. I think we will pick different values between production and staging because we don't have the same number of pods, same instance types etc.

Can you submit a Kubernetes PR instead? This recent PR shows what you need to change when introducing a new env variable https://github.com/cds-snc/notification-manifests/pull/210

Regarding the value, we need to do more tests in staging before picking a number

AntoineAugusti:
Closing, work done in https://github.com/cds-snc/notification-manifests/pull/226

PR #1233, opened 2021-02-09
feat: send no reply with Notify service

AntoineAugusti:
Add a new Celery task to deliver a generic "No Reply" email using the GC Notify service.

This PR creates a new GC Notify template and adds a Celery task. This Celery task will be triggered through SES -> Lambda -> SQS.

Trello card: https://trello.com/c/PpAuHXEM/271-replies-to-notify-should-quickly-communicate-that-they-dont-work


migrations/versions/0314_no_reply_template.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+
+
+def downgrade():
+    for template in templates:
+        op.execute("DELETE FROM notifications WHERE template_id = '{}'".format(template['id']))
+        op.execute("DELETE FROM notification_history WHERE template_id = '{}'".format(template['id']))
+        op.execute("DELETE FROM template_redacted WHERE template_id = '{}'".format(template['id']))
+        op.execute("DELETE FROM templates WHERE id = '{}'".format(template['id']))
+        op.execute("DELETE FROM templates_history WHERE id = '{}'".format(template['id']))
```
AntoineAugusti, line 79: These lines are standard when creating a new Notify template, taken from a recent migration introducing the branding request template https://github.com/cds-snc/notification-api/blob/master/migrations/versions/0305b_branding_request_template.py

migrations/versions/0314_no_reply_template.py
```diff
@@ -0,0 +1,79 @@
...
+            "Your message was not delivered.",
+            "",
+            "The email address ((sending_email_address)) is not able to receive messages since this feature has not been set by the sender.",
+            "",
+            "___",
+            "",
+            "Votre message n’a pas été livré.",
+            "",
+            "L’adresse courriel ((sending_email_address)) ne peut pas recevoir de messages car cette fonction n’a pas été définie par l’expéditeur.",
+        ],
```
AntoineAugusti, line 32: Content coming from @anikbrazeau

app/config.py
```diff
@@ -190,6 +190,7 @@ class Config(object):
     CONTACT_US_TEMPLATE_ID = '8ea9b7a0-a824-4dd3-a4c3-1f508ed20a69'
     ACCOUNT_CHANGE_TEMPLATE_ID = '5b39e16a-9ff8-487c-9bfb-9e06bdb70f36'
     BRANDING_REQUEST_TEMPLATE_ID = '7d423d9e-e94e-4118-879d-d52f383206ae'
+    NO_REPLY_TEMPLATE_ID = '86950840-6da4-4865-841b-16028110e980'
```
AntoineAugusti, line 193: Generated a random UUID

app/celery/tasks.py
```diff
@@ -624,3 +624,28 @@ def process_returned_letters_list(notification_references):
             updated, updated_history, len(notification_references)
         )
     )
+
+
+@notify_celery.task(name='send-notify-no-reply')
+@statsd(namespace="tasks")
+def send_notify_no_reply(data):
```
AntoineAugusti, line 631: This is the Celery task that will be triggered by the Lambda (hooked up to SES). In reality, the Lambda will post a message to an SQS queue to trigger this task

app/celery/tasks.py
```diff
@@ -624,3 +624,28 @@ def process_returned_letters_list(notification_references):
...
+    service = dao_fetch_service_by_id(current_app.config['NOTIFY_SERVICE_ID'])
+    template = dao_get_template_by_id(current_app.config['NO_REPLY_TEMPLATE_ID'])
+
+    saved_notification = persist_notification(
+        template_id=template.id,
+        template_version=template.version,
+        recipient=payload["sender"],
+        service=service,
+        personalisation={
+            'sending_email_address': payload["recipients"][0]
```
AntoineAugusti, line 643: Special case, if people write to `foo@notification.canada.ca` and `bar@notification.canada.ca` at the same time (like multiple fields in `To`/`Cc`/`Bcc`) so always passing a list and using the first argument because this is just really weird.

The list will be logged on CloudWatch though, that's interesting to know that folks are doing this.

migrations/versions/0314_no_reply_template.py
```diff
@@ -0,0 +1,79 @@
+"""
+
+Revision ID: 0314_no_reply_template
+Revises: 0313_disable_pinpoint_provider
+Create Date: 2021-02-09 13:37:42
```
AntoineAugusti, line 5: 😏

app/celery/tasks.py
```diff
@@ -624,3 +624,28 @@ def process_returned_letters_list(notification_references):
...
+
+@notify_celery.task(name='send-notify-no-reply')
+@statsd(namespace="tasks")
+def send_notify_no_reply(data):
+    payload = json.loads(data)
+
+    service = dao_fetch_service_by_id(current_app.config['NOTIFY_SERVICE_ID'])
+    template = dao_get_template_by_id(current_app.config['NO_REPLY_TEMPLATE_ID'])
+
+    saved_notification = persist_notification(
```
jimleroyer, line 637: Can we wrap this in a `try..catch` so we make sure the error is logged and in any case, the notification is sent to the queue? Looking at the `persist_notification` function, it seems exceptions will bubble up for now and interrupt the notification to be placed on the queue.

migrations/versions/0314_no_reply_template.py
jimleroyer: 🤣

tests/app/celery/test_tasks.py
```diff
@@ -1607,3 +1608,28 @@ def test_process_returned_letters_list_updates_history_if_notification_is_alread
 
     assert [n.status for n in notifications] == ['returned-letter', 'returned-letter']
     assert all(n.updated_at for n in notifications)
+
+
+def test_send_notify_no_reply(mocker, no_reply_template):
```
jimleroyer, line 1613: We could also have a test if the database fail but the notification should be sent.

AntoineAugusti:
Added a retry logic and made sure that persisting the notification and dispatching it to the queue bubble up exceptions

jimleroyer (approved):

jimleroyer:
:shipit:

PR #1234, opened 2021-02-10
Lowering Default Rate Limit For Services

RenDavidDimen:
# Summary | Résumé

Changing the default rate limit from 3000 to 1000 as explained in [this Trello card](https://trello.com/c/UzY03qj2/258-lower-default-api-rate-limit-per-minute)


migrations/versions/0315_lower_api_rate_limit.py
```diff
@@ -0,0 +1,19 @@
...
+revision = '0315_lower_api_rate_limit'
+down_revision = '0314_no_reply_template'
+
+def upgrade():
+    op.execute("ALTER TABLE services ALTER rate_limit DROP DEFAULT")
+    op.execute("ALTER TABLE services_history ALTER rate_limit DROP DEFAULT")
+
+def downgrade():
+    op.execute("ALTER TABLE services ALTER rate_limit SET DEFAULT '1000'")
+    op.execute("ALTER TABLE services_history ALTER rate_limit SET DEFAULT '1000'")
```
AntoineAugusti, line 19: You'll need to change your upgrade and downgrade methods.

`upgrade` is called when applying the migration. For us we want to:
- alter the default value and changed it to 1000
- update rows where the value is currently 3000 and set it to 1000

`downgrade` is when we want rollback a database migration. For us we would need to:
- alter the default value and changed it to 3000
- update rows where the value is currently 1000 and set it to 3000

You need 3 SQL statements for each method

AntoineAugusti (changes requested):

AntoineAugusti (approved):
LGTM! Thanks

PR #1237, opened 2021-02-19
Avoid hard termination of API pods

jimleroyer:
# Summary | Résumé

Lowered the value for `graceful_timeout` within gunicorn configuration. The default is `30` seconds but we set it at `20` seconds.

# Rationale

Following our recent incident which caused an increase of database connections to get past the RDS instance limit, our current theory is that as the kubernetes pods kept restarting the instances, these did not clean up resources including their database connections in a clean manner.

Lowering the gunicorn configuration value for `graceful_timeout` might provide a better heads up to the gunicorn workers to properly work out the clean up of resources. Because the previous default value of 30 seconds matched the exact value for kubernetes graceful pod timeout as well (i.e. `terminationGracePeriodSeconds`), which might not give the proper time for gunicorn to gracefully shut down.

## A minor rebuttal

This all depends on how gunicorn treats its `graceful_timeout` value internally. From the documentation, we can read:

> After receiving a restart signal, workers have this much time to finish serving requests. Workers still alive after the timeout (starting from the receipt of the restart signal) are force killed.

I have the impression that it might not help, as a force kill will not help clean up the resources properly anyway. I hope I am wrong and that this will help though. 🤞  

# Test instructions and some investigation

I loaded gunicorn locally with the following command, which enabled the DEBUG level of logging:

```
gunicorn -c gunicorn_config.py application -b 127.0.0.1:6011 -p ./gunicorn.pid --access-logfile ./gunicorn.log --log-level debug
```

Right there, I could confirm the changed configuration value deviating from the default due to the newest changes:

```
...
timeout: 30
graceful_timeout: 20
keepalive: 2
...
```

Yay! But then, I wanted to actually test a pseudo scenario of what happened during the incident by sending a SIGTERM signal to the gunicorn instances while it was hanging on the database table(s). For that, I setup the following...

1. Run a SQL statement that locks on the `alembic_version` table:

```
begin;
lock table alembic_version in ACCESS exclusive mode;
-- commit;
```

The commit instruction should be discarded until we need to free the table out of the lock; so no run for now.

2. As we locked on the `alembic_version` table, this is the one that will get called by the health check endpoint. I opened a few windows tab with the local URL...

http://localhost:6011/_status

...and confirmed that the HTTP call was ongoing and could not get delivered due to the database lock.

3. It was then time to send the `SIGTERM` by retrieving the main gunicorn process (lowest ID out of `ps -ax | grep gun` and then send the signal:

```
kill -s TERM 25890
```

4. This produced the following logs where the timestamps of the SIGTERM  report +20 seconds matched the forced process kill:

```
[2021-02-18 22:44:55 -0500] [25901] [DEBUG] GET /_status
[2021-02-18 22:45:24 -0500] [25890] [INFO] Handling signal: term
[2021-02-18 22:45:25 -0500] [25900] [INFO] Worker exiting (pid: 25900)
[2021-02-18 22:45:44 -0500] [25901] [INFO] Worker exiting (pid: 25901)
[2021-02-18 22:45:44 -0500] [25890] [INFO] Shutting down: Master
[2021-02-18 22:45:44 -0500] [25890] [INFO] Stopping Notifications API
```

So we know that the graceful termination timeout took effect with the new updated gunicorn configuration value.

5. Let's test the database connections number before and after the SIGTERM signal now. First, I execute the following SQL query to get all active connections to my local postgresql database:

```
select pid as process_id, 
       usename as username, 
       datname as database_name, 
       client_addr as client_address, 
       application_name,
       backend_start,
       state,
       state_change
from pg_stat_activity
order by process_id;
```

I get these results which provide me the active connections prior to start the gunicorn workers:

|process_id|username|database_name|client_address|application_name|backend_start|state|state_change|
|----------|--------|-------------|--------------|----------------|-------------|-----|------------|
|26|||||2021-02-15 11:16:47|||
|27|||||2021-02-15 11:16:47|||
|28|||||2021-02-15 11:16:47|||
|29|||||2021-02-15 11:16:47|||
|31|postgres||||2021-02-15 11:16:47|||
|1987|postgres|notification_api|172.22.0.1|DBeaver 7.3.5 - Main <notification_api>|2021-02-18 22:16:40|idle|2021-02-18 22:21:22|
|1988|postgres|notification_api|172.22.0.1|DBeaver 7.3.5 - Metadata <notification_api>|2021-02-18 22:16:41|idle|2021-02-18 23:30:44|
|1989|postgres|notification_api|172.22.0.1|DBeaver 7.3.5 - SQLEditor <Script-1.sql>|2021-02-18 22:16:41|active|2021-02-18 23:31:40|
|1990|postgres|postgres|172.22.0.1|DBeaver 7.3.5 - Main <postgres>|2021-02-18 22:16:45|idle|2021-02-18 22:16:45|
|1991|postgres|postgres|172.22.0.1|DBeaver 7.3.5 - Metadata <postgres>|2021-02-18 22:16:45|idle|2021-02-18 22:16:45|

So we should discard the connections above to know the ones used by the workers. I started gunicorn, hit the health check endpoint a few times to activate database connections and re-executed the SQL query. I get the same connections as above along with these two new ones:

|process_id|username|database_name|client_address|application_name|backend_start|state|state_change|
|----------|--------|-------------|--------------|----------------|-------------|-----|------------|
|2151|postgres|notification_api|172.22.0.1||2021-02-18 23:31:06|idle|2021-02-18 23:31:06|
|2152|postgres|notification_api|172.22.0.1||2021-02-18 23:31:06|idle|2021-02-18 23:31:39|

I proceed afterward to lock the `alembic_version` table, hit the health check, send the SIGTERM signal to the gunicorn master process. All along, I was executing the SQL query to know if our database connections were dropped. **After** the 20 seconds of graceful period timeout, the SQL results still return our database connections, living after the workers has been killed.

It's only **when I released the lock on the table that the connections get dropped**.

### Interesting notes

a) The 2 connections owned by the workers were only released once the lock was released. 
b) Once the lock was released, there was absolutely no problem to let go of the connections. The connections did not linger at all behind.

I can't correlate and explain why we reached the connection limit on RDS during the incident. I don't know enough and the context might be different, hence different results. But it's good stuff to know and ponder about.

> Things that you (the submitter) want reviewers to pay very close attention to when they review this.

---

> Éléments auxquels vous (le demandeur) souhaitez que les réviseurs prêtent attention lorsqu’ils examineront votre demande.

> Are there any related issues or tangent features you consider out of scope for this issue that could be addressed in the future?

---

> Y a-t-il des questions connexes que vous considérez hors sujet et qui pourraient être abordées plus tard?

### DIFF
--- a/gunicorn_config.py
+++ b/gunicorn_config.py
@@ -19,6 +19,15 @@ in_production = os.environ.get("NOTIFY_ENVIRONMENT", "") == "production"
 if in_production:
     keepalive = 75
 
+# The default graceful timeout period for Kubernetes is 30 seconds, so
+# want a lower graceful timeout value for gunicorn so that proper instance
+# shutdowns.
+#
+# Gunicorn config:
+# https://docs.gunicorn.org/en/stable/settings.html#graceful-timeout
+#
+# Kubernetes config:
+# https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/
 if in_production:
     graceful_timeout = 20
 
--- a/gunicorn_config.py
+++ b/gunicorn_config.py
@@ -16,8 +16,7 @@ accesslog = '-'
 # > By default, Elastic Load Balancing sets the idle timeout value for your load balancer to 60 seconds.
 # https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#connection-idle-timeout
 in_production = os.environ.get("NOTIFY_ENVIRONMENT", "") == "production"
-if in_production:
-    keepalive = 75
+if in_production: keepalive = 75
 
 # The default graceful timeout period for Kubernetes is 30 seconds, so
 # want a lower graceful timeout value for gunicorn so that proper instance
@@ -28,8 +27,7 @@ if in_production:
 #
 # Kubernetes config:
 # https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/
-if in_production:
-    graceful_timeout = 20
+if in_production: graceful_timeout = 20
 
 
 def on_starting(server):
--- a/gunicorn_config.py
+++ b/gunicorn_config.py
@@ -16,7 +16,8 @@ accesslog = '-'
 # > By default, Elastic Load Balancing sets the idle timeout value for your load balancer to 60 seconds.
 # https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#connection-idle-timeout
 in_production = os.environ.get("NOTIFY_ENVIRONMENT", "") == "production"
-if in_production: keepalive = 75
+if in_production:
+    keepalive = 75
 
 # The default graceful timeout period for Kubernetes is 30 seconds, so
 # want a lower graceful timeout value for gunicorn so that proper instance
@@ -27,7 +28,8 @@ if in_production: keepalive = 75
 #
 # Kubernetes config:
 # https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/
-if in_production: graceful_timeout = 20
+if in_production:
+    graceful_timeout = 20
 
 
 def on_starting(server):
--- a/gunicorn_config.py
+++ b/gunicorn_config.py
@@ -10,14 +10,17 @@ worker_connections = 256
 bind = "0.0.0.0:{}".format(os.getenv("PORT"))
 accesslog = '-'
 
-# See AWS doc
+# To avoid load balancers reporting errors on shutdown instances, see AWS doc
 # > We also recommend that you configure the idle timeout of your application
-# to be larger than the idle timeout configured for the load balancer.
+# > to be larger than the idle timeout configured for the load balancer.
 # > By default, Elastic Load Balancing sets the idle timeout value for your load balancer to 60 seconds.
 # https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#connection-idle-timeout
 in_production = os.environ.get("NOTIFY_ENVIRONMENT", "") == "production"
 if in_production:
     keepalive = 75
+
+if in_production:
+    graceful_timeout = 20
 
 
 def on_starting(server):
--- a/gunicorn_config.py
+++ b/gunicorn_config.py
@@ -10,25 +10,24 @@ worker_connections = 256
 bind = "0.0.0.0:{}".format(os.getenv("PORT"))
 accesslog = '-'
 
-# To avoid load balancers reporting errors on shutdown instances, see AWS doc
-# > We also recommend that you configure the idle timeout of your application
-# > to be larger than the idle timeout configured for the load balancer.
-# > By default, Elastic Load Balancing sets the idle timeout value for your load balancer to 60 seconds.
-# https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#connection-idle-timeout
 in_production = os.environ.get("NOTIFY_ENVIRONMENT", "") == "production"
 if in_production:
+    # To avoid load balancers reporting errors on shutdown instances, see AWS doc
+    # > We also recommend that you configure the idle timeout of your application
+    # > to be larger than the idle timeout configured for the load balancer.
+    # > By default, Elastic Load Balancing sets the idle timeout value for your load balancer to 60 seconds.
+    # https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#connection-idle-timeout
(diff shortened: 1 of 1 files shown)


gunicorn_config.py
```diff
@@ -10,15 +10,18 @@
...
-# See AWS doc
+# To avoid load balancers reporting errors on shutdown instances, see AWS doc
 # > We also recommend that you configure the idle timeout of your application
-# to be larger than the idle timeout configured for the load balancer.
+# > to be larger than the idle timeout configured for the load balancer.
 # > By default, Elastic Load Balancing sets the idle timeout value for your load balancer to 60 seconds.
 # https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#connection-idle-timeout
 in_production = os.environ.get("NOTIFY_ENVIRONMENT", "") == "production"
 if in_production:
     keepalive = 75
```
jimleroyer, line 21: Unsure if I had to use the `in_production` condition on the new configuration. I would like that to apply on staging but there is no such configuration environment (what do you think?). Does `testing` applies for staging?

gunicorn_config.py
```diff
@@ -10,15 +10,18 @@
 bind = "0.0.0.0:{}".format(os.getenv("PORT"))
 accesslog = '-'
 
-# See AWS doc
+# To avoid load balancers reporting errors on shutdown instances, see AWS doc
```
jimleroyer, line 13: Just added a preliminary line here to explain the why on the configuration.

gunicorn_config.py
AntoineAugusti: `NOTIFY_ENVIRONMENT` is currently set to `production` in staging. That's could be misleading (or could be explained by the fact that you want to match really production conditions), anyway at the moment this is how it is and you can expect it to work this way.

gunicorn_config.py
```diff
@@ -10,15 +10,18 @@
...
-# to be larger than the idle timeout configured for the load balancer.
+# > to be larger than the idle timeout configured for the load balancer.
 # > By default, Elastic Load Balancing sets the idle timeout value for your load balancer to 60 seconds.
 # https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#connection-idle-timeout
 in_production = os.environ.get("NOTIFY_ENVIRONMENT", "") == "production"
 if in_production:
     keepalive = 75
 
+if in_production:
+    graceful_timeout = 20
```
AntoineAugusti, line 23: Would move the new line in the same `if` statement.

gunicorn_config.py
```diff
@@ -10,15 +10,18 @@
 bind = "0.0.0.0:{}".format(os.getenv("PORT"))
 accesslog = '-'
 
-# See AWS doc
+# To avoid load balancers reporting errors on shutdown instances, see AWS doc
 # > We also recommend that you configure the idle timeout of your application
-# to be larger than the idle timeout configured for the load balancer.
+# > to be larger than the idle timeout configured for the load balancer.
 # > By default, Elastic Load Balancing sets the idle timeout value for your load balancer to 60 seconds.
```
AntoineAugusti, line 16: Could you add links to Gunicorn documentation for keepalive and graceful_timeout while you're at it?

AntoineAugusti (approved):
Thanks for the very detailed PR and for testing the `SIGTERM` signal locally while looking at database connections.

I feel like this PR should not cause harm and will likely improve the situation but may not solve completely our issue of cleaning up open database connections when a large number of pods are killed in a short period. We could also look at setting the connection timeout on the app side at ~30s at on the RDS side at ~45s - to make sure that the API is the one closing/reopening connections under normal operations but that RDS is able to close connections that have been left open.

I believe CI failed because you committed late at night, local ET time or during the night in UTC/BST. That's bad (and something we'll need to fix) but it has not been introduced by your PR. Triggering the job should be enough to have green tests.

gunicorn_config.py
jimleroyer: Done..

gunicorn_config.py
jimleroyer: Thanks for the confirmation.

gunicorn_config.py
jimleroyer: Okay I replicated the style format of the previous line of codes with the `keepalive` config. I will adjust others too then.

jimleroyer: The build does not pass the style check this way so I will have to revert:

```
Requirements check passed
./gunicorn_config.py:19:17: E701 multiple statements on one line (colon)
./gunicorn_config.py:30:17: E701 multiple statements on one line (colon)
```

PR #1238, opened 2021-02-24
Fix test suite to run locally without errors as well as in the evening

dorianjp:
Fixes some tests and reports which break when the API test suite is run locally or when run on any platform between 00:00 UTC  and 00:00 local time (America/Toronto) or any GMT- time zone.

Trello:
https://trello.com/c/hOCUVQRc/89-fix-the-notification-api-tests-so-it-passes-in-local
https://trello.com/c/WZuI7jWL/376-admin-and-api-tests-suites-should-succeed-at-night

Example failure in CI tests:
https://github.com/cds-snc/notification-api/runs/1957724231?check_suite_focus=true#step:9:4109


README.md
```diff
@@ -95,6 +95,13 @@ file. Copy that file to `.env` and customize it to your needs.
 
 `make test`
 
+  * Ensure vars:
+    * SQLALCHEMY_DATABASE_READER_URI
+    * SQLALCHEMY_DATABASE_URI
```
AntoineAugusti, line 100:
```suggestion
    * `SQLALCHEMY_DATABASE_READER_URI`
    * `SQLALCHEMY_DATABASE_URI`
```

README.md
```diff
@@ -95,6 +95,13 @@ file. Copy that file to `.env` and customize it to your needs.
...
+  * are both set either in .env or pytest.ini (precedence goes to .env) to your local db with user access
```
AntoineAugusti, line 101:
```suggestion
  * are both set either in `.env` or `pytest.ini` (precedence goes to `.env`) to your local database with user access
```

README.md
```diff
@@ -95,6 +95,13 @@ file. Copy that file to `.env` and customize it to your needs.
...
+  for CREATE DATABASE access. Additionally the db in the URI should not be set to 'notification_api'.
```
AntoineAugusti, line 102:
```suggestion
  for CREATE DATABASE access. Additionally the database in the URI should not be set to `notification_api`.
```

app/dao/services_dao.py
```diff
@@ -444,7 +445,7 @@ def _stats_for_service_query(service_id):
 
 @statsd(namespace='dao')
 def dao_fetch_todays_stats_for_all_services(include_from_test_key=True, only_active=True):
-    today = date.today()
+    today = convert_utc_to_local_timezone(datetime.utcnow())
```
AntoineAugusti, line 448: Makes sense 👍

pytest.ini
```diff
@@ -13,6 +13,8 @@ env =
     FRESH_DESK_PRODUCT_ID=42
     FRESH_DESK_API_URL=https://example.com/freshdesk
     FRESH_DESK_API_KEY=freshdesk-api-key
+    ##### Define URIs before running local tests #####
     # SQLALCHEMY_DATABASE_URI=postgresql://postgres:LocalPassword@db:5432/test_notification_api
+    # SQLALCHEMY_DATABASE_READER_URI=postgresql://postgres:LocalPassword@db:5432/test_notification_api
```
AntoineAugusti, line 18: Could we set `SQLALCHEMY_DATABASE_URI=postgresql://postgres@localhost:5432/test_notification_api`

That should be a sensible default for developers and should work out of the box. Otherwise, what's the recommended way? Do you manually change your `.env` before running tests?

Can we not set `SQLALCHEMY_DATABASE_READER_URI` and tests will run or does it need to have a user? What about setting it to the same value (defeats the purpose or testing a read-only database user I know - but this is checked on GitHub Actions)

tests/app/test_cronitor.py
```diff
@@ -82,9 +82,10 @@ def test_cronitor_does_nothing_if_name_not_recognised(notify_api, rmock, caplog)
     }):
         assert successful_task() == 1
 
-    error_log = caplog.records[0]
-    assert error_log.levelname == 'ERROR'
-    assert error_log.msg == 'Cronitor enabled but task_name hello not found in environment'
+    if caplog.records:
```
AntoineAugusti, line 85: Curious about this one, can you explain? BTW we don't use Cronitor so it's not a deal breaker

AntoineAugusti (commented):
Thanks for drafting this PR! Running tests locally and fixing them at night seems like a very good progress 👍 

What do you think of making tests run ~every hour, at least at night (ET time). You can leverage GH Actions `schedule` https://docs.github.com/en/actions/reference/events-that-trigger-workflows#schedule

pytest.ini
dorianjp: I don't mind if we can have reader URI default to run as writer when not specified.

It needs to have a valid user, the initial connection is reused to create the test databases. I'm open to either @AntoineAugusti. 

I felt it needs to be explicitly changed by the developer, since we have the python-dotenv package installed which always looks for the .env, and often a developer may have a working dev configuration, which is not a good test configuration due to the database specificity. The .env will overwride the pytest, and either the developer needs to clear or change the .env to run clean tests. If we have a sensible default, it would work, but it may also lead to ambiguity because it's specified, but not looking there for the most common developer config using .env

Ironically, it would be nice to have the python-dotenv engage via its own env flag.

tests/app/test_cronitor.py
dorianjp: this is a strange one, but when run locally, the caplog which uses LogCaptureFixture never receives the error expected from the CRONITOR configuration applied during the test. I attempted to derive the difference from the CI which doesn't fail in the same way (the caplog.records is not an empty list in CI), but didn't see it. As I see we don't use cronitor, i felt this was. a safe compromise until we decide the long term direction for the cronitor code itself.

dorianjp:
@AntoineAugusti, The URI is now set in pytest.ini to usable defaults for both CI and local which override anything we may have in the .env file. The CI no longer needs to pre-create the reader user or inject either URIs into the env, both are discovered from pytest.ini.

Also I changed a few detaults to ca-central-1 for examples and tests, but left the fallbacks in the app codebase for now.

AntoineAugusti (approved):
Awesome, thanks for doing this work!

A suggestion for parsing DB connection strings and consider adding a `schedule` to the CI `test` job to make sure tests pass at night.

PR #1240, opened 2021-03-01
feat: remove reply-to for Notify no reply emails

AntoineAugusti:
## Steps to Reproduce
1) Reply to No-reply email from Notify
2) Check Freshdesk

## Expected Behaviour
- Email goes to the void

## Actual Behaviour
- People sending us their personal information

## Notes
We worked recently on providing feedback for people replying to emails sent by Notify where services did not put a reply-to email address (see linked Trello card). This "no reply" email is sent by the GC Notify service and uses the default reply-to email address of this service, which goes to Freshdesk.

We've seen that some people reply to these emails and provide personal information, which is a privacy concern for us. **Make sure that we don't get these emails by changing the reply-to email address of these notifications only.** We can do this by setting the reply-to email address to the Notify service email address, [this is handled by our Lambda](https://github.com/cds-snc/notification-terraform/blob/a6c08a9a7f9a69f70c700a3475af5591af0e93ba/aws/common/lambdas/ses_receiving_emails.py#L120-L126).

Trello card: https://trello.com/c/XKdRjDY9/430-block-replies-to-no-reply-emails-from-gc-notify

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -646,7 +646,11 @@ def send_notify_no_reply(self, data):
             notification_type=template.template_type,
             api_key_id=None,
             key_type=KEY_TYPE_NORMAL,
-            reply_to_text=service.get_default_reply_to_email_address(),
+            # Ensure that the reply to is not set, if people reply
+            # to these emails, they will go to the GC Notify service
+            # email address, and we handle those on the SES inbound
+            # Lambda
+            reply_to_text=None,
         )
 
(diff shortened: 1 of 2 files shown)


willeybryan (approved):

jimleroyer (approved):

AntoineAugusti:
Tested it worked fine in staging, thanks

Screen Shot 2021-03-01 at 10 18 02
Screen Shot 2021-03-01 at 10 18 22

PR #1241, opened 2021-03-02
fix: removed gov.uk from error message

QuanNguyen-CDS:
# Summary | Résumé
[Trello Link](https://trello.com/c/bEMNSBCA)
fix: updated docker-compose and pytest.ini file so tests would run using docker-compose
fix: replaced documentation links as well as added a #FIXME when we can get the proper translations (as discussed with Anik)
refactor: updated test cases to account for the new link

## Testing For Docker-Compose unit tests
1. From project root folder (on my system) modify scripts/run_tests.sh to only run single execution by modifying -n4 -> -n1
2. Launch docker-compose `docker-compose up --build`
3. Validate that the tests run to completion from the `test` container and that there are no more database connection errors

## Testing For Error Message
1. Get a python script that will exercise the sms and email sending via api (Thanks Antoine)
2. Setup a service, set it live (you have to be a platform admin), create a "Normal" api key, set your service to restricted (did this by database write directly within the services table)
3. Run the script sending to a personal account, validate the error message is `Can’t send to this recipient when service is in trial mode – see documentation.localhost:6012/en/keys.html#live` - note that this is because of your location configuration.
4. Replace localhost:6012 with the staging URI and validate that this leads to the documentation link
5. Validate that running the pytest unit tests 
`tests/app/notifications/rest/test_send_notification.py`
`tests/app/notifications/test_validators.py`
`tests/app/notifications/test_rest.py`
works

> Étapes consécutives (1., 2., 3., …) qui décrivent la façon de tester la modification. Elles aideront les développeurs à faire des tests sans avoir à jouer au détective. Veuillez aussi inclure toutes les étapes de configuration de l’environnement qui ne font pas partie des étapes normales dans le fichier README et tout élément temporel requis.

As discussed with Anik, we will want to make these error messages better but for the interim this is good for now and will help out developers if they require assistance. Future work will be to translate these error messages into both french and english and also provide a useful link (in french/english) based upon the users preference
A Trello card has been created within the product backlog for this: [Trello Link](https://trello.com/c/ohwfHAUr)
A #FIXME has been added into the code base

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -134,8 +134,8 @@ class Config(object):
     # PII check
     SCAN_FOR_PII = os.getenv("SCAN_FOR_PII", False)
 
-    # Documentation and Translation
-    DOCUMENTATION_DOMAIN = os.getenv('DOCUMENTATION_DOMAIN', f'https://documentation.{ADMIN_BASE_DOMAIN}')
+    # Documentation
+    DOCUMENTATION_DOMAIN = os.getenv('DOCUMENTATION_DOMAIN', f'documentation.{ADMIN_BASE_DOMAIN}')
 
     ###########################
     # Default config values ###
--- a/app/config.py
+++ b/app/config.py
@@ -68,7 +68,6 @@ class TaskNames(object):
 class Config(object):
     # URL of admin app
     ADMIN_BASE_URL = os.getenv('ADMIN_BASE_URL', 'http://localhost:6012')
-    ADMIN_BASE_DOMAIN = ADMIN_BASE_URL.replace('http://', '').replace('https://', '')
 
     # URL of api app (on AWS this is the internal api endpoint)
     API_HOST_NAME = os.getenv('API_HOST_NAME')
@@ -135,7 +134,7 @@ class Config(object):
     SCAN_FOR_PII = os.getenv("SCAN_FOR_PII", False)
 
     # Documentation
-    DOCUMENTATION_DOMAIN = os.getenv('DOCUMENTATION_DOMAIN', f'documentation.{ADMIN_BASE_DOMAIN}')
+    DOCUMENTATION_DOMAIN = os.getenv('DOCUMENTATION_DOMAIN', 'documentation.notification.canada.ca')
 
     ###########################
     # Default config values ###
--- a/app/config.py
+++ b/app/config.py
@@ -135,7 +135,7 @@ class Config(object):
     SCAN_FOR_PII = os.getenv("SCAN_FOR_PII", False)
 
     # Documentation and Translation
-    DOCUMENTATION_DOMAIN = os.getenv('DOCUMENTATION_DOMAIN', f'documentation.{ADMIN_BASE_DOMAIN}')
+    DOCUMENTATION_DOMAIN = os.getenv('DOCUMENTATION_DOMAIN', f'https://documentation.{ADMIN_BASE_DOMAIN}')
 
     ###########################
     # Default config values ###
--- a/app/config.py
+++ b/app/config.py
@@ -68,6 +68,7 @@ class TaskNames(object):
 class Config(object):
     # URL of admin app
     ADMIN_BASE_URL = os.getenv('ADMIN_BASE_URL', 'http://localhost:6012')
+    ADMIN_BASE_DOMAIN = ADMIN_BASE_URL.replace('http://', '').replace('https://', '')
 
     # URL of api app (on AWS this is the internal api endpoint)
     API_HOST_NAME = os.getenv('API_HOST_NAME')
@@ -132,6 +133,9 @@ class Config(object):
 
     # PII check
     SCAN_FOR_PII = os.getenv("SCAN_FOR_PII", False)
+
+    # Documentation and Translation
+    DOCUMENTATION_DOMAIN = os.getenv('DOCUMENTATION_DOMAIN', f'documentation.{ADMIN_BASE_DOMAIN}')
 
     ###########################
     # Default config values ###
--- a/app/notifications/rest.py
+++ b/app/notifications/rest.py
@@ -181,9 +181,10 @@ def _service_allowed_to_send_to(notification, service):
         if api_user.key_type == KEY_TYPE_TEAM:
             message = 'Can’t send to this recipient using a team-only API key'
         else:
+            # FIXME: hard code it for now until we can get en/fr specific links and text
             message = (
                 'Can’t send to this recipient when service is in trial mode '
-                '– see https://www.notifications.service.gov.uk/trial-mode'
+                f'– see {current_app.config["DOCUMENTATION_DOMAIN"]}/en/keys.html#live'
             )
         raise InvalidRequest(
             {'to': [message]},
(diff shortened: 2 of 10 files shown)


app/notifications/rest.py
```diff
@@ -181,9 +181,10 @@ def _service_allowed_to_send_to(notification, service):
         if api_user.key_type == KEY_TYPE_TEAM:
             message = 'Can’t send to this recipient using a team-only API key'
         else:
+            # FIXME: hard code it for now until we can get en/fr specific links and text
```
AntoineAugusti, line 184: Wondering if we're supposed to write "hard code" or "hardcode"

app/notifications/rest.py
```diff
@@ -181,9 +181,10 @@ def _service_allowed_to_send_to(notification, service):
         if api_user.key_type == KEY_TYPE_TEAM:
             message = 'Can’t send to this recipient using a team-only API key'
```
AntoineAugusti, line 182: You could link to a documentation page as well here

app/notifications/validators.py
```diff
@@ -81,9 +81,10 @@ def service_can_send_to_recipient(send_to, key_type, service, allow_safelisted_r
         if key_type == KEY_TYPE_TEAM:
             message = 'Can’t send to this recipient using a team-only API key'
```
AntoineAugusti, line 82: You could link to a documentation page as well here

pytest.ini
```diff
@@ -16,7 +16,7 @@ env =
     AWS_REGION=ca-central-1
     ASSET_DOMAIN=assets.notification.canada.ca
     NOTIFY_EMAIL_DOMAIN=notification.canada.ca
-    SQLALCHEMY_DATABASE_URI=postgresql://postgres:postgres@localhost:5432/test_notification_api
-    SQLALCHEMY_DATABASE_READER_URI=postgresql://reader:postgres@localhost:5432/test_notification_api
+    D:SQLALCHEMY_DATABASE_URI=postgresql://postgres:postgres@localhost:5432/test_notification_api
+    D:SQLALCHEMY_DATABASE_READER_URI=postgresql://reader:postgres@localhost:5432/test_notification_api
```
AntoineAugusti, line 20: Wonderful, thanks for doing that!

In the future, could you draft up another PR to do these changes? This file and the Docker compose changes are not related to other changes.

tests/app/notifications/rest/test_send_notification.py
```diff
@@ -254,7 +254,7 @@ def test_should_not_send_notification_if_restricted_and_not_a_service_user(notif
             assert response.status_code == 400
             assert [(
                 'Can’t send to this recipient when service is in trial mode '
-                '– see https://www.notifications.service.gov.uk/trial-mode'
+                f'– see {current_app.config["DOCUMENTATION_DOMAIN"]}/en/keys.html#live'
```
AntoineAugusti, line 257: Wondering if it would be better to use a hardcoded value in your tests, to prove that the value is indeed what you expect. I think that at the moment you're missing the `https://` part for example

tests/app/notifications/test_validators.py
```diff
@@ -276,16 +277,19 @@ def test_service_can_send_to_recipient_fails_when_ignoring_safelist(
...
                           ('normal',
-                           "Can’t send to this recipient when service is in trial mode – see https://www.notifications.service.gov.uk/trial-mode")])  # noqa
-def test_service_can_send_to_recipient_fails_when_recipient_is_not_on_team(recipient, key_type, error_message,
-                                                                           notify_db, notify_db_session):
+                           "Can’t send to this recipient when service is in trial mode – see ")])  # noqa
+def test_service_can_send_to_recipient_fails_when_recipient_is_not_on_team(recipient: str,
+                                                                           key_type: str,
+                                                                           error_message: str,
+                                                                           notify_db: RoutingSQLAlchemy,
+                                                                           notify_db_session: RoutingSQLAlchemy):
```
AntoineAugusti, line 285: Would be a more common indentation
```suggestion
def test_service_can_send_to_recipient_fails_when_recipient_is_not_on_team(
    recipient: str,
    key_type: str,
    error_message: str,
    notify_db: RoutingSQLAlchemy,
    notify_db_session: RoutingSQLAlchemy
):
```

app/config.py
```diff
@@ -133,6 +134,9 @@ class Config(object):
     # PII check
     SCAN_FOR_PII = os.getenv("SCAN_FOR_PII", False)
 
+    # Documentation and Translation
```
AntoineAugusti, line 137: Translation?

AntoineAugusti (changes requested):

app/notifications/rest.py
qcorporation-cds: ha! good question, I've looked online and they are both accepted 👍

pytest.ini
qcorporation-cds: thanks for the note- i'll do that in the future. This was blocking me for testing the changes so I lumped it in. But in the future i'll make a separate change.

app/config.py
qcorporation-cds: it was aspirational actually this comment, I wanted any future work, which I envision we would do in regards to offering english and french translations, to go here. I can remove and add it in later when we do

tests/app/notifications/rest/test_send_notification.py
QuanNguyen-CDS: added https into DOCUMENTATION_DOMAIN

app/config.py
QuanNguyen-CDS: as discussed on slack, we can follow up on this after. I was hoping translation constants can also be below this comment as well along with the DOCUMENTATION_DOMAIN. Currently only DOCUMENTATION_DOMAIN is defined but in the future others can be related to translation

tests/app/notifications/test_validators.py
QuanNguyen-CDS: formatting done

app/notifications/rest.py
QuanNguyen-CDS: linked added for team api

app/config.py
AntoineAugusti: Can you remove the "Translation" part for now until we decide if we do further changes?

app/config.py
```diff
@@ -133,6 +134,9 @@ class Config(object):
...
+    DOCUMENTATION_DOMAIN = os.getenv('DOCUMENTATION_DOMAIN', f'https://documentation.{ADMIN_BASE_DOMAIN}')
```
AntoineAugusti, line 138: Can you leave out the `https://` from here? The `_DOMAIN` name gives the impression that it will not include a protocol.

The same variable/env name on the admin does not include the protocol https://github.com/cds-snc/notification-admin/blob/f947e0d2053b372d2321670dcf7d0260287dba8d/app/config.py#L32

If it's a burden to add the protocol multiple maybe you can create a small helper to add the protocol and the `/en/` prefix if you feel like it

AntoineAugusti (changes requested):

app/config.py
QuanNguyen-CDS: agreed, refactored and added a helper method called `get_document_url` that follows other utility functions

AntoineAugusti (approved):
LGTM! Thanks 👏 

I'll let you add a quick test for the `get_document_url` method before merging and that's a wrap 🚢

Issue #1217, opened 2021-01-29
Different psycopg error message in (failed) test


davidbgk:
Hello 👋,

From a fresh install, following instructions from the README file, using the latest available version of [Posgres.app](https://postgresapp.com/downloads.html) (v2.4.1 with Postgres 13.1):

```
% py.test --disable-pytest-warnings --cov=app --cov-report=term-missing tests/ --junitxml=test_results.xml -n4 -v --maxfail=10 -k test_create_service_and_history_is_transactional -vv
=============================================================== test session starts ================================================================
platform darwin -- Python 3.9.1, pytest-3.10.1, py-1.10.0, pluggy-0.13.1 -- /Users/david/Sites/notification-api/venv/bin/python3
cachedir: .pytest_cache
rootdir: /Users/david/Sites/notification-api, inifile: pytest.ini
plugins: cov-2.6.1, requests-mock-1.8.0, env-0.6.2, mock-1.10.4, forked-1.3.0, xdist-1.27.0
[gw0] darwin Python 3.9.1 cwd: /Users/david/Sites/notification-api
[gw1] darwin Python 3.9.1 cwd: /Users/david/Sites/notification-api
[gw2] darwin Python 3.9.1 cwd: /Users/david/Sites/notification-api
[gw3] darwin Python 3.9.1 cwd: /Users/david/Sites/notification-api
[gw0] Python 3.9.1 (default, Jan 29 2021, 14:12:33)  -- [Clang 12.0.0 (clang-1200.0.32.29)]
[gw1] Python 3.9.1 (default, Jan 29 2021, 14:12:33)  -- [Clang 12.0.0 (clang-1200.0.32.29)]
[gw2] Python 3.9.1 (default, Jan 29 2021, 14:12:33)  -- [Clang 12.0.0 (clang-1200.0.32.29)]
[gw3] Python 3.9.1 (default, Jan 29 2021, 14:12:33)  -- [Clang 12.0.0 (clang-1200.0.32.29)]
gw0 [1] / gw1 [1] / gw2 [1] / gw3 [1]
scheduling tests via LoadScheduling

tests/app/dao/test_services_dao.py::test_create_service_and_history_is_transactional 
[gw0] [100%] FAILED tests/app/dao/test_services_dao.py::test_create_service_and_history_is_transactional 

===================================================================== FAILURES =====================================================================
_________________________________________________ test_create_service_and_history_is_transactional _________________________________________________
[...]

PR #1245, opened 2021-03-05
docs: mention the correct Postgres version to use

davidbgk:
When using version 13.2, one test (test_create_service_and_history_is_transactional) is failing because the `IntegrityError` message slightly differs from the one using version 12.6 of the database.

See #1217 for details.


README.md
```diff
@@ -57,7 +57,7 @@ source  ~/.pyenv/versions/3.9.1/bin/virtualenvwrapper.sh
 
 `workon notifications-api`
 
-9. Install [Postgres.app](http://postgresapp.com/).
+9. Install [Postgres.app](http://postgresapp.com/) (version 12.6, not the latest).
```
AntoineAugusti, line 60: We run 11.9 in staging/production at the moment. May be good to update the version here while we're at it! https://github.com/cds-snc/notification-api/blob/0ede6a61b48289236d1873124965d2bc22a9b27b/docker-compose.yml#L5

README.md
```diff
@@ -196,3 +198,8 @@ __Problem__: Messages are in the queue but not sending
 
 __Solution__: Check that `celery` is running. 
 
+---
+
+__Problem__: Running tests, `test_create_service_and_history_is_transactional` is failing with this error: `assert 'column "name" violates not-null constraint' in '(psycopg2.errors.NotNullViolation) null value in column "name" of relation "services_history" violates not-null constraint`
+
+__Solution__: Ensure your PostgreSQL version is 12.6, this error happens with version 13+.
```
AntoineAugusti, line 205:
```suggestion
__Solution__: Ensure your PostgreSQL version is 11.9, this error happens with version 13+.
```

AntoineAugusti (changes requested):
Thanks for opening a PR!

README.md
davidbgk: I would like to emphasis the fact that 13+ is not supported yet, but maybe we could provide a range for that particular known error, something like: `Ensure your PostgreSQL version is above 11 but below 13.`

README.md
davidbgk: The Postgres app [currently allows](https://postgresapp.com/downloads.html) 13.2, 12.6, 11.11, 10.16 and older ones. Given that version 12.6 is the upper one which still works I would recommend it in the install process for the dev environment. We might suggest 11.11 to be closer to the staging/prod env.

sastels:
@davidbgk @jimleroyer is this still relevant (looks like it is?) / can we merge or close?

davidbgk:
> @davidbgk @jimleroyer is this still relevant (looks like it is?) / can we merge or close?

I still think it's pertinent for a newcomer, feel free to amend the commits if you need to!

PR #1249, opened 2021-03-15
feat: pass filename argument when sending attachments

AntoineAugusti:
## Features

- Accept an optional new parameter on the API side, `filename`
- Pass this argument to the document download API, to reuse it when the email will be sent later or to pass it to the document download frontend See https://github.com/cds-snc/notification-document-download-api/pull/28
- Use the `filename` argument, if present, when attaching files to emails. Previously, attachments were named after the personalisation key in the template.

Trello card: https://trello.com/c/BVIe44l5/443-let-clients-name-the-files-they-are-attaching-in-the-api


tests/app/delivery/test_send_to_providers.py
```diff
@@ -866,6 +867,62 @@ def test_notification_raises_sets_notification_to_virus_found_if_mlwr_score_abov
...
+    filename,
+    expected_filename,
+):
+    template = sample_email_template(notify_db, notify_db_session, content="Here is your ((file))")
+    personalisation = {
+        "file": {
+            "document": {
+                "direct_file_url": "http://foo.bar/direct_file_url",
+                "url": "http://foo.bar/url",
+            },
```
AntoineAugusti, line 889: Mimicks what will be returned by the document download API

app/delivery/send_to_providers.py
```diff
@@ -129,7 +129,11 @@ def send_email_to_provider(notification):
                     buffer = response.read()
                     mime_type = magic.from_buffer(buffer, mime=True)
                     if mime_type == 'application/pdf':
```
AntoineAugusti, line 131: This logic will need to be updated when you'll work on choosing if services want to attach OR link to documents

dorianjp: thanks

dorianjp (approved):
looks like a feature!

PR #1252, opened 2021-03-17
feat: log iat value to ease debugging with JWT tokens

AntoineAugusti:
JWT tokens are only valid for 30 seconds. When clients send requests to the Notify API, they can get a clock error, but it's a bit too generic to be helpful.

> Error: Your system clock must be accurate to within 30 seconds

This PR adds some logging, to be able to hunt down specific values in AWS log groups to have more evidence to help clients.

Trello card: https://trello.com/c/Mbkumte7/376-log-jwt-iat-field-value-to-ease-debugging-with-clock-errors

### DIFF
--- a/app/authentication/auth.py
+++ b/app/authentication/auth.py
@@ -1,5 +1,10 @@
 from flask import request, _request_ctx_stack, current_app, g
-from notifications_python_client.authentication import decode_jwt_token, get_token_issuer
+from notifications_python_client.authentication import (
+    decode_jwt_token,
+    decode_token,
+    epoch_seconds,
+    get_token_issuer,
+)
 from notifications_python_client.errors import TokenDecodeError, TokenExpiredError, TokenIssuerError
 from notifications_utils import request_helper
 from sqlalchemy.exc import DataError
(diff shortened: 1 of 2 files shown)


app/authentication/auth.py
```diff
@@ -102,6 +107,10 @@ def requires_auth():
         except TokenDecodeError:
             continue
         except TokenExpiredError:
+            decoded_token = decode_token(auth_token)
```
AntoineAugusti, line 110: Need to use `decode_token` which does not check the signature/does not throw an exception to extract the `iat` field.

`decode_jwt_token`, used a few lines before, throws an exception.

See https://github.com/alphagov/notifications-python-client/blob/ec19d4dd08e2037df867aa5a7171849183f0b6e3/notifications_python_client/authentication.py#L139-L145

QuanNguyen-CDS, line 110: `decode_token` can raise `DecodeError` if the json.loads(...) from the decoded text throws, it can also throw a `TypeError`. Are we OK in handling these exceptions in the parent caller chain?

AntoineAugusti: Caught a potential `DecodeError` in the last commit. I don't think running `decode_token` every time in the `try` statement is needed, would like to run this only if this rare error happens.

QuanNguyen-CDS (approved):

PR #1255, opened 2021-03-26
Adding zendesk create service integration

QuanNguyen-CDS:
# Summary | Résumé
[Trello](https://trello.com/c/MkP7JBIs)
feat: zendesk create service integration
refactor: zendesk_sell to be more DRY including name splitting
refactor: error flows for zendesk_sell, removed NotImplementedError raise
feat: added 3 new methods, upsert contact, delete contact, upsert deal
refactor: testing for test_rest for creation of new service
feat: adding more test cases for upsert contact, delete contact and upsert deal

1) ran the unit test case `service/test_rest.py` `test_zendesk_sell.py`
2) through the contact us form create a demo request, validate that zendesk lead gets created (regression testing)
3) create a service, validate that the service gets created within the service dashboard, navigate to zendesk sell and validate:
- a contact is created with firstname, lastname, email and mobile number
- a deal is created linking back to the newly created contact
- the deal contains service name as the deal name
- the deal is in Trial stage


app/clients/zendesk_sell.py
```diff
@@ -55,21 +60,44 @@ def _generate_lead_data(contact: ContactRequest) -> Dict[str, Union[str, List[st
...
+        assert len(user.name) and len(user.email_address), 'Name or email are mandatory field'
+
+        first_name, last_name = ZenDeskSell._name_split(user.name)
+        return {
+            'data': {
+                'last_name': last_name,
+                'first_name': first_name,
+                'email': user.email_address,
+                'mobile': user.mobile_number,
+            }
```
AntoineAugusti, line 76: Would it be possible to add the Notify ID here? Maybe in a custom field?

app/clients/zendesk_sell.py
```diff
@@ -55,21 +60,44 @@ def _generate_lead_data(contact: ContactRequest) -> Dict[str, Union[str, List[st
...
+        # validation based upon api mandatory fields
+        assert len(user.name) and len(user.email_address), 'Name or email are mandatory field'
+
+        first_name, last_name = ZenDeskSell._name_split(user.name)
+        return {
+            'data': {
+                'last_name': last_name,
+                'first_name': first_name,
+                'email': user.email_address,
+                'mobile': user.mobile_number,
```
AntoineAugusti, line 75: Does the phone number need to be transformed, maybe using the international format?

app/clients/zendesk_sell.py
```diff
@@ -82,10 +110,113 @@ def send_lead(self, contact: ContactRequest) -> int:
...
+            current_app.logger.warning('Did not upsert contact to zendesk')
+            return None, False
+
+        try:
+            response = requests.post(
+                url=urljoin(self.api_url, f'/v2/contacts/upsert?email={user.email_address}'),
+                data=json.dumps(ZenDeskSell._generate_contact_data(user)),
+                headers={'Accept': 'application/json', 'Content-Type': 'application/json'},
+                auth=BearerAuth(token=self.token),
+                timeout=5
```
AntoineAugusti, line 129: WDYT about refactoring these HTTP calls a bit? Headers, auth, timeout, checking if the API URL is set etc are duplicated quite a few times now. You could have a private method accepting the HTTP method, URI and data maybe? Could look at https://github.com/alphagov/notifications-python-client/blob/master/notifications_python_client/base.py for some inspiration.

app/clients/zendesk_sell.py
```diff
@@ -82,10 +110,113 @@ def send_lead(self, contact: ContactRequest) -> int:
...
+            )
+            response.raise_for_status()
+        except requests.RequestException:
+            current_app.logger.warning('Failed to create zendesk sell contact')
+            return None, False
+
+        # response validation
+        try:
+            resp_data = json.loads(response.text)
```
AntoineAugusti, line 138: You can use `response.json()` here and elsewhere

app/clients/zendesk_sell.py
```diff
@@ -82,10 +110,113 @@ def send_lead(self, contact: ContactRequest) -> int:
...
+            assert 'data' in resp_data, 'Missing "data" field in response'
+            assert 'id' in resp_data['data'], 'Missing "id" field in response'
+            assert 'created_at' in resp_data['data'], 'Missing "created_at" field in response'
+            assert 'updated_at' in resp_data['data'], 'Missing "updated_at" field in response'
+
```
AntoineAugusti, line 143: Do you need all these assertions? The return statement will raise if those fields don't exist

app/clients/zendesk_sell.py
```diff
@@ -82,10 +110,113 @@ def send_lead(self, contact: ContactRequest) -> int:
...
+    def send_create_service(self, service: Service, user: User) -> bool:
+        # Upsert a contact (create/update). Only when this is successful does the software upsert a deal
+        # and link the deal to the contact.
+        # If upsert deal fails go back and delete the contact ONLY if it never existed before
+        contact_id, is_created = self.upsert_contact(user)
+        if not contact_id:
+            return False
+
+        # 11826762 is a zendesk number to signify "Created Trial"
+        deal_id = self.upsert_deal(contact_id, service, 11826762)
```
AntoineAugusti, line 209: Can you add a class constant for the magic number? It'll be more explicit and you could remove a comment.

app/clients/zendesk_sell.py
```diff
@@ -82,10 +110,113 @@ def send_lead(self, contact: ContactRequest) -> int:
...
+        if not deal_id and is_created:
+            # best effort here
+            self.upsert_contact(contact_id)
```
AntoineAugusti, line 212: Should be `delete_contact` here I think

app/clients/zendesk_sell.py
```diff
@@ -55,21 +60,44 @@ def _generate_lead_data(contact: ContactRequest) -> Dict[str, Union[str, List[st
...
+        }
+
+    @staticmethod
+    def _generate_deal_data(contact_id: int, service: Service, stage_id: int) -> Dict[str, Union[str, List[str], Dict]]:
+        return {
+            'data': {
+                'contact_id': contact_id,
+                'name': service.name,
+                'stage_id': stage_id,
+            }
```
AntoineAugusti, line 86: Could we have an external ID? Services can change their name.

tests/app/clients/test_zendesk_sell.py
```diff
@@ -1,9 +1,12 @@
+import json
```
AntoineAugusti, line 1: Can you add a test for `send_create_service`?

tests/app/service/test_rest.py
```diff
@@ -1,4 +1,5 @@
 import json
```
AntoineAugusti, line 1: For this test file I see that you can just mock the Zendesk client and assert it was called with the expected arguments. The other file already makes sure that HTTP requests are sent as expected.

AntoineAugusti (changes requested):

app/clients/zendesk_sell.py
QuanNguyen-CDS: My opinion is that we want to match the Zendesk Integration directly to the data that we have within our Notify database. I'd like to keep the data as a 1-to-1 mapping as much as possible.

If the better format is to transform it to an international number format I'd also recommend also modifying the mobile number at form inject to the international format within our database. WDYT?

app/clients/zendesk_sell.py
QuanNguyen-CDS: Yeah that's a good suggestion, I was really worried about the lifecycle of a deal. There are situations where if the user changes the service name that we wouldn't be able to track it within zendesk as the deal name would match.

Which then moves to another integration point, do we want to also send an update on [service name -> deal name] if the service name changes at any point in time? Currently to my knowledge we don't have that and the lifecycle of deal isn't tracked past the service creation.

app/clients/zendesk_sell.py
QuanNguyen-CDS: Yeah that's a good suggestion, I was really worried about the lifecycle of a deal. There are situations where if the user changes the service name that we wouldn't be able to track it within zendesk as the deal name wouldn't match.

Which then moves to another integration point, do we want to also send an update on [service name -> deal name] if the service name changes at any point in time? Currently to my knowledge we don't have that and the lifecycle of deal isn't tracked past the service creation.

app/clients/zendesk_sell.py
QuanNguyen-CDS: yes absolutely, DRY.

app/clients/zendesk_sell.py
QuanNguyen-CDS: I didn't know how to do this at first, but it seems you can add a custom field via the Zendesk UI. I've added this and have updated the code

app/clients/zendesk_sell.py
QuanNguyen-CDS: nope, removing

app/clients/zendesk_sell.py
QuanNguyen-CDS: sorry thanks for picking this up - i need unit testing around this

tests/app/clients/test_zendesk_sell.py
QuanNguyen-CDS: test cases added

tests/app/service/test_rest.py
QuanNguyen-CDS: modified tests to just mock the method `send_create_service`

PR #1256, opened 2021-03-31
Use US Toll Free Number When Sending SMS Messages to the US

RenDavidDimen:
# Summary | Résumé

Adding logic to have messages to US phone numbers sent from a US Toll Free Number in accordance with future legal updates.

[Special requirements for sending SMS messages to US destinations - Amazon Simple Notification Service](https://docs.aws.amazon.com/sns/latest/dg/channels-sms-us-requirements.html)

---

### DIFF
--- a/app/clients/sms/aws_sns.py
+++ b/app/clients/sms/aws_sns.py
@@ -1,4 +1,5 @@
 import re
+from flask import current_app
 from time import monotonic
 
 import boto3
@@ -59,7 +60,7 @@ class AwsSnsClient(SmsClient):
                 }
 
             # If the number is US based, we must use a US Toll Free number to send the message
-            if phonenumbers.region_code_for_number == 'US':
+            if phonenumbers.region_code_for_number(to) == 'US':
                 client = self._us_client
                 attributes['AWS.MM.SMS.OriginationNumber'] = {
                     'DataType': 'String',
--- a/app/clients/sms/aws_sns.py
+++ b/app/clients/sms/aws_sns.py
@@ -58,6 +58,14 @@ class AwsSnsClient(SmsClient):
                     'StringValue': sender
                 }
 
+            # If the number is US based, we must use a US Toll Free number to send the message
+            if phonenumbers.region_code_for_number == 'US':
+                client = self._us_client
+                attributes['AWS.MM.SMS.OriginationNumber'] = {
+                    'DataType': 'String',
+                    'StringValue': current_app.config['AWS_US_TOLL_FREE_NUMBER']
+                }
+
             try:
                 start_time = monotonic()
                 response = client.publish(
--- a/app/clients/sms/aws_sns.py
+++ b/app/clients/sms/aws_sns.py
@@ -1,5 +1,4 @@
 import re
-from flask import current_app
 from time import monotonic
 
 import boto3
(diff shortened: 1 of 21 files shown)


app/config.py
```diff
@@ -151,6 +151,7 @@ class Config(object):
     AWS_SES_ACCESS_KEY = os.getenv('AWS_SES_ACCESS_KEY')
     AWS_SES_SECRET_KEY = os.getenv('AWS_SES_SECRET_KEY')
     AWS_PINPOINT_REGION = os.getenv('AWS_PINPOINT_REGION', "us-west-2")
+    AWS_US_TOLL_FREE_NUMBER = os.getenv('AWS_US_TOLL_FREE_NUMBER', "us-west-2")
```
RenDavidDimen, line 154: Currently unsure if specifying the region here is necessary.

I also need to update K8s to hold the information here.

tests/app/clients/test_aws_sns.py
```diff
@@ -51,3 +52,27 @@ def test_send_sms_with_long_code_successful_returns_aws_sns_response(notify_api,
             'AWS.MM.SMS.OriginationNumber': {'DataType': 'String', 'StringValue': sender},
         }
     )
+
+
+def test_send_sms_to_us_number_successful_returns_aws_sns_response(notify_api, mocker):
```
AntoineAugusti, line 57: Could you add an additional test demonstrating that if we pass a custom sender, the US toll-free number is used for US phone numbers and not the passed sender?

Thanks!

app/clients/sms/aws_sns.py
```diff
@@ -58,6 +58,15 @@ def send_sms(self, to, content, reference, multi=True, sender=None):
                     'StringValue': sender
                 }
 
+            # If the number is US based, we must use a US Toll Free number to send the message
+            country = phonenumbers.region_code_for_number(phonenumbers.parse(to))
```
AntoineAugusti, line 62: Are you able to reuse `to` or `match` or another attribute to avoid parsing twice the phone number?

AntoineAugusti (changes requested):

app/clients/sms/aws_sns.py
RenDavidDimen: It appears that `region_code_for_number` requires a `numobj` which can be acquired by parsing a string.

From my understanding the way to get there is parsing `to` which is a string acquired by parsing `match`.

The `match.number.national_number` returns an `int` which is not accepted by `region_code_for_number`

Because of this, I believe we must go from `match.number -> to -> region_code_for_number(to)` for the method to work as intended. But it's also possible that I have missed something!

RenDavidDimen: Changed parameter sent to `region_code_for_number` to `match.number` as it fulfills the `numobj` type requirement

tests/app/clients/test_aws_sns.py
RenDavidDimen: Added test case!

app/config.py
RenDavidDimen: Removed region here so as to simply pull the Toll Free Number

AntoineAugusti (approved):
LGTM 📞 

Can merge to staging and production, env variables have been set

PR #1257, opened 2021-04-01
fix: handle hand crafted Notify user agents

AntoineAugusti:
Fix a possible 500 error when people submit a custom user agent, starting with `notify` but with an unexpected format.

See occurence on Slack: https://gcdigital.slack.com/archives/CNWA63606/p1617214953097200

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -278,7 +278,7 @@ def process_user_agent(user_agent_string):
         return 'unknown'
 
     m = re.search(
-        r'(?P<name>notify.*)/(?P<version>\d+.\d+.\d+)',
+        r'^(?P<name>notify.*)/(?P<version>\d+.\d+.\d+)$',
         user_agent_string,
         re.IGNORECASE
     )
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -275,9 +275,12 @@ def create_random_identifier():
 def process_user_agent(user_agent_string):
     if user_agent_string and user_agent_string.lower().startswith("notify"):
         components = user_agent_string.split("/")
-        client_name = components[0].lower()
-        client_version = components[1].replace(".", "-")
-        return "{}.{}".format(client_name, client_version)
+        try:
+            client_name = components[0].lower()
+            client_version = components[1].replace(".", "-")
+            return "{}.{}".format(client_name, client_version)
+        except IndexError:
+            return "unknown"
     elif user_agent_string and not user_agent_string.lower().startswith("notify"):
         return "non-notify-user-agent"
     else:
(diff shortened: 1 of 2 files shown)


app/__init__.py
```diff
@@ -275,9 +275,12 @@ def create_random_identifier():
 def process_user_agent(user_agent_string):
     if user_agent_string and user_agent_string.lower().startswith("notify"):
         components = user_agent_string.split("/")
-        client_name = components[0].lower()
-        client_version = components[1].replace(".", "-")
-        return "{}.{}".format(client_name, client_version)
+        try:
+            client_name = components[0].lower()
+            client_version = components[1].replace(".", "-")
+            return "{}.{}".format(client_name, client_version)
```
AntoineAugusti, line 281: This code was meant to handle this case

```py
process_user_agent("NOTIFY-API-PYTHON-CLIENT/3.0.0")
# "notify-api-python-client.3-0-0"
```

It logs libs usage on StatsD

QuanNguyen-CDS: I like this approach, if I can make a suggestion of using regex and making the matching a bit more strict. something like this:
`
def process_user_agent(agent_str: str):
    try:
        m = re.search(r'(?P<name>notify.*)/(?P<version>\d*.\d*.\d)', agent_str, re.IGNORECASE)
        return f'{m.group("name")}.{m.group("version").replace(".", "-")}' if m else 'non-notify-user-agent'
    except TypeError:
        return 'unknown'
`

AntoineAugusti: Refactored to regex in next commit.

tests/app/test_user_agent_processing.py
```diff
@@ -1,13 +1,19 @@
...
-    assert "notify-api-python-client.3-0-0" == process_user_agent("NOTIFY-API-PYTHON-CLIENT/3.0.0")
 
-
-def test_can_handle_non_notify_api_user_agent():
-    assert "non-notify-user-agent" == process_user_agent("Mozilla/5.0 (iPad; U; CPU OS 3_2_1 like Mac OS X; en-us) AppleWebKit/531.21.10 (KHTML, like Gecko) Mobile/7B405")  # noqa
-
-
-def test_handles_null_user_agent():
-    assert "unknown" == process_user_agent(None)
+@pytest.mark.parametrize(
```
QuanNguyen-CDS, line 6: i like the tests!

QuanNguyen-CDS (approved):
LGTM.
Just a small datapoint, if we want to make it even more strict we can switch `search` with `match` if we want to detect if `notify` is at the start of `user_agent_string` instead of somewhere inside `user_agent_string`. A small nuisance and the pattern is quite explicit and doesn't block the approval

PR #1258, opened 2021-04-06
feat: delete daily limit cache when updating limits

AntoineAugusti:
Remove the daily limit cache key from Redis when a service's message limit value is updated.

At the moment, if a service goes above its message limit and we update the message limit value, the cache may be used for up to 1h before expiring. This is not a great experience because if we change limits we want that to be reflected right rather than telling clients to wait for up to 1h.

https://github.com/cds-snc/notification-api/blob/3db95eb950de58faeecefe0818c95c7872863d1f/app/notifications/validators.py#L38-L44

You can check for other places where `daily_limit_cache_key` is used in this codebase to understand how it works.

Trello: https://trello.com/c/7VIM7U6r/419-delete-daily-limit-cache-when-updating-a-services-message-limit

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -252,7 +252,9 @@ def update_service(service_id):
     fetched_service = dao_fetch_service_by_id(service_id)
     # Capture the status change here as Marshmallow changes this later
     service_going_live = fetched_service.restricted and not req_json.get('restricted', True)
-    message_limit_changed = fetched_service.message_limit != req_json.get('message_limit', None)
+    message_limit_changed = (
+        fetched_service.message_limit != req_json.get('message_limit', fetched_service.message_limit)
+    )
     current_data = dict(service_schema.dump(fetched_service).data.items())
 
     current_data.update(request.get_json())
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -7,12 +7,14 @@ from flask import (
     current_app,
     Blueprint
 )
+from notifications_utils.clients.redis import daily_limit_cache_key
 from notifications_utils.letter_timings import letter_can_be_cancelled
 from notifications_utils.timezones import convert_utc_to_local_timezone
 from sqlalchemy import func
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm.exc import NoResultFound
 
+from app import redis_store
 from app.clients.zendesk_sell import ZenDeskSell
 from app.config import QueueNames
 from app.dao import fact_notification_status_dao, notifications_dao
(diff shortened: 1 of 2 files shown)


AntoineAugusti:
cc @dorianjp, we talked about this on Slack recently, you may be interested to see this PR

app/service/rest.py
```diff
@@ -262,10 +264,11 @@ def update_service(service_id):
     if 'letter_branding' in req_json:
         letter_branding_id = req_json['letter_branding']
         service.letter_branding = None if not letter_branding_id else LetterBranding.query.get(letter_branding_id)
-    dao_update_service(service)
 
-    if 'default_email_is_french' in req_json:
-        service.default_email_is_french = req_json['default_email_is_french']
```
AntoineAugusti, line 268: This piece was not needed. A test proves that it still works https://github.com/cds-snc/notification-api/blob/c32ec8ea4aacf51d87da41031031f543d96261e9/tests/app/service/test_rest.py#L712-L719

app/service/rest.py
```diff
@@ -262,10 +264,11 @@ def update_service(service_id):
...
+    if 'message_limit' in req_json:
+        redis_store.delete(daily_limit_cache_key(service_id))
```
QuanNguyen-CDS, line 269: will there be any time where the value is the same and now what's happening is the cache is getting deleted for no reason?
i.e. should we check the new value of `message_limit` for equality before deleting the cache?

AntoineAugusti: Good question! The [admin code](https://github.com/cds-snc/notification-admin/blob/master/app/notify_client/service_api_client.py) passes the `message_limit` explicitly when updating a service so it seems safe to assume that if this key is here, the value has changed.

Updating a service does not happen often no matter what. It's done from service settings by service owners or a platform admin

AntoineAugusti: Improved the logic in the next commit

app/service/rest.py
```diff
@@ -250,6 +252,7 @@ def update_service(service_id):
     fetched_service = dao_fetch_service_by_id(service_id)
     # Capture the status change here as Marshmallow changes this later
     service_going_live = fetched_service.restricted and not req_json.get('restricted', True)
+    message_limit_changed = fetched_service.message_limit != req_json.get('message_limit', None)
```
QuanNguyen-CDS, line 255: are there situations where 'message_limit' is never passed down?
in those situations
`fetched_service.message_limit` will be some value while req_json.get('message_limit', None) will return `None`
In that case the cache will be deleted.

Suggestion:
`message_limit_changed = fetched_service.message_limit != req_json.get('message_limit') if 'message_limit' in req_json else False`

AntoineAugusti: Right, forgot this case. Thanks for catching it! Sorry about that.

QuanNguyen-CDS (approved):
nice, thank you - looks good

AntoineAugusti:
Thanks for your review, making sure it's robust

PR #1259, opened 2021-04-07
feat: save stats for attachments

AntoineAugusti:
Save stats when attachments are sent.

Store:
- number of attachments in the notification
- number of attachments sent on Notify
- number of attachments sent per service
- number of attachments sent per template
- number of attachments sent per sending method
- number of attachments sent per file type
- number of attachments sent per file size (1mb bucket size)

Uses MIME type and file size from https://github.com/cds-snc/notification-document-download-api/pull/35

Trello: https://trello.com/c/LJ3Ae6Of/422-store-stats-about-attachments

### DIFF
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -5,7 +5,7 @@ import werkzeug
 from flask import request, jsonify, current_app, abort
 from notifications_utils.recipients import try_validate_and_format_phone_number
 
-from app import api_user, authenticated_service, notify_celery, document_download_client
+from app import api_user, authenticated_service, notify_celery, document_download_client, statsd_client
 from app.celery.letters_pdf_tasks import create_letters_pdf, process_virus_scan_passed
 from app.celery.research_mode_tasks import create_fake_letter_response_file
 from app.clients.document_download import DocumentDownloadError
@@ -191,7 +191,12 @@ def process_sms_or_email_notification(*, form, notification_type, api_key, templ
     # Do not persist or send notification to the queue if it is a simulated recipient
     simulated = simulated_recipient(send_to, notification_type)
 
-    personalisation = process_document_uploads(form.get('personalisation'), service, simulated=simulated)
+    personalisation = process_document_uploads(
+        form.get('personalisation'),
+        service,
+        simulated,
+        template.id
+    )
 
     notification = persist_notification(
         template_id=template.id,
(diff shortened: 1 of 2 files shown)


app/v2/notifications/post_notifications.py
```diff
@@ -244,9 +249,32 @@ def process_document_uploads(personalisation_data, service, simulated=False):
...
 
 
+def save_stats_for_attachments(files_data, service_id, template_id):
+    nb_files = len(files_data)
+    statsd_client.incr(f"attachments.nb-attachments.count-{nb_files}")
+    statsd_client.incr('attachments.nb-attachments', count=nb_files)
+    statsd_client.incr(f"attachments.services.{service_id}", count=nb_files)
+    statsd_client.incr(f"attachments.templates.{template_id}", count=nb_files)
+
+    for document in [f['document'] for f in files_data]:
```
QuanNguyen-CDS, line 269: Is there a potential for raise `KeyError`, does the caller up to the REST endpoint able to handle this?

tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -36,6 +37,27 @@
 )
 
 
+def document_download_response(override={}):
+    # See response from
+    # https://github.com/cds-snc/notification-document-download-api/blob/master/app/upload/views.py
+    base = {
```
QuanNguyen-CDS, line 43: the [link above ](https://github.com/cds-snc/notification-document-download-api/blob/master/app/upload/views.py#L41) doesn't show a response of `mime_type`, `file_size` and `file_extension` unless I'm missing something

AntoineAugusti: Not yet because the PR https://github.com/cds-snc/notification-document-download-api/pull/35 hasn't been merged yet, but they will be there after

app/v2/notifications/post_notifications.py
AntoineAugusti: No, a potential `KeyError` is not caught. If some keys are not there that's a real error. We control responses from the document download API and it should match the output from `document_download_response` in tests. Or you can look at the document API HTTP endpoint as well

AntoineAugusti: We've got a similar assumption about keys here as well. This is covered by tests

https://github.com/cds-snc/notification-api/blob/46f9ce23d0acc8ff52fb65fe96ac72ea3f5a40bc/app/delivery/send_to_providers.py#L108-L142

QuanNguyen-CDS (approved):
Good

PR #1260, opened 2021-04-08
fix: appropriate mime type for emails with attachments

AntoineAugusti:
Reword the SES logic to craft emails with attachments.

The logic:

- If sending a TXT email without attachments:
  => Multipart mixed
- If sending a TXT + HTML email without attachments:
  => Multipart alternative
- If sending a TXT + HTML email with attachments
=>  Multipart Mixed (enclosing)
      - Multipart alternative (TXT + HTML)
      - Attachment(s)

See AWS doc: https://docs.aws.amazon.com/ses/latest/DeveloperGuide/send-email-raw.html#send-email-raw-mime
StackOverflow: https://stackoverflow.com/a/64174056

I'm sorry I didn't update tests but unfortunately test coverage is not good for this file: https://github.com/cds-snc/notification-api/blob/master/tests/app/clients/test_aws_ses.py. We don't test what's passed to AWS at the moment.

PR #1261, opened 2021-04-08
feat: write tests for AWS SES

AntoineAugusti:
Follow up of https://github.com/cds-snc/notification-api/pull/1260

Refactor + writing tests for 2 main cases:
- TXT + HTML, without attachments
- TXT + HTML + attachments

### DIFF
--- a/tests/app/clients/test_aws_ses.py
+++ b/tests/app/clients/test_aws_ses.py
@@ -1,4 +1,6 @@
+import re
 from base64 import b64encode
+from textwrap import dedent
 
 import botocore
 import pytest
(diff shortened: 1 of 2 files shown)


tests/app/clients/test_aws_ses.py
```diff
@@ -80,6 +82,111 @@ def test_send_email_handles_reply_to_address(notify_api, mocker, reply_to_addres
...
+            body='email body',
+            html_body='<p>email body</p>',
+            reply_to_address='reply@example.com',
+        )
+
+    boto_mock.send_raw_email.assert_called_once()
+    raw_message = boto_mock.send_raw_email.call_args.kwargs['RawMessage']['Data']
+
+    regex = dedent(r"""
+        Content-Type: multipart\/alternative; boundary="===============(?P<boundary>.+)=="
```
QuanNguyen-CDS, line 103: is there a reason why we are capturing name groups?

QuanNguyen-CDS, line 103: is there any worry about multiline regex compatibility across platforms? just a bit worried about this. No concrete example of it being wrong.

QuanNguyen-CDS (approved):
Comment are inline and minor concerns, not blocking the merge

tests/app/clients/test_aws_ses.py
AntoineAugusti: Yes, those name groups are here to prove that envelopes have been set as expected. Some things need to be contained within things and groups are boundaries https://github.com/cds-snc/notification-api/blob/9ca52c3c0d44f745b4097b9a6d72cf854b6e3431/app/clients/email/aws_ses.py#L95-L106

AntoineAugusti: It passed at least on Linux and macOS, let's see if we get reports for Windows

PR #1262, opened 2021-04-12
feat: add bulk process type and route bulk notifications 

AntoineAugusti:
## What / The story

Some services send bulk notifications using Notify, through the API. The most famous one is "Get updates on Covid-19". Those templates clog up our queue and create a delay, up to 5-10 minutes, for other emails while they're sending (takes 20-30 minutes to complete).

## Why / The RoadMap link

Delivering emails slowly is a poor user experience for other services. We've got an SLO for email deliveries (90% under 10s and 99% under 60s) we're not meeting because of that.

## What I did

- Added a new template process type: `bulk`
- Routed notifications to the appropriate queue according to the template process
- Updated tests to account for this new template process type

Trello: https://trello.com/c/pPYbRqYo/428-reduce-email-delays-by-routing-bulk-notifications-to-another-queue


jimleroyer (approved):
:shipit:

PR #1263, opened 2021-04-13
feat: add content-type header for direct attachments

AntoineAugusti:
Add a `Content-Type` header when sending direct attachments with AWS SES.

It's recommend in [the documentation](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/send-email-raw.html#send-email-raw-mime) ("File attachments" section). Previously it was not set or AWS SES added a generic header like `Content-Type: application/octet-stream`.

Trello: https://trello.com/c/XzhQ42Jx/440-specify-content-type-when-sending-direct-attachments

### DIFF
--- a/tests/app/conftest.py
+++ b/tests/app/conftest.py
@@ -1246,3 +1246,24 @@ def app_statsd(mocker):
 
 def datetime_in_past(days=0, seconds=0):
     return datetime.now(tz=pytz.utc) - timedelta(days=days, seconds=seconds)
+
+
+def document_download_response(override={}):
+    # See response from
+    # https://github.com/cds-snc/notification-document-download-api/blob/master/app/upload/views.py
+    base = {
+        'id': 'document-id',
+        'direct_file_url': 'http://direct-file-url.localdomain',
+        'url': 'http://frontend-url.localdomain',
+        'mlwr_sid': 'mlwr-sid',
+        'filename': 'filename',
+        'sending_method': 'sending_method',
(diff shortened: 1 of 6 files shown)


tests/app/conftest.py
```diff
@@ -1246,3 +1246,24 @@ def app_statsd(mocker):
 
 def datetime_in_past(days=0, seconds=0):
     return datetime.now(tz=pytz.utc) - timedelta(days=days, seconds=seconds)
+
+
+def document_download_response(override={}):
```
AntoineAugusti, line 1251: It was in `tests/app/v2/notifications/test_post_notifications.py` before. Moved it here because it's now used in 2 places

tests/app/delivery/test_send_to_providers.py
```diff
@@ -912,7 +908,7 @@ def test_notification_document_with_pdf_attachment(
     if filename_attribute_present:
         request_mock.assert_called_once_with('http://foo.bar/direct_file_url')
         urlopen_mock.assert_called_once_with('request_mock')
-        attachments = [{'data': 'request_content', 'name': expected_filename}]
+        attachments = [{'data': 'request_content', 'name': expected_filename, 'mime_type': 'application/pdf'}]
```
AntoineAugusti, line 911: Other tests are refactors, this is the real interesting test for this class

QuanNguyen-CDS (approved):
LGTM

PR #1264, opened 2021-04-13
Remove mention of Travis on healthcheck

AntoineAugusti:
We do not use Travis, remove those fields from the Makefile and the healthcheck page https://api.notification.alpha.canada.ca.

They are creating confusion for developers.


QuanNguyen-CDS (approved):
LGTM

PR #1265, opened 2021-04-14
feat: save provider response for tech failures on SNS

AntoineAugusti:
Add a new column to the `notifications` table to store the raw provider response. We don't have enough notification statuses to cover the wide range of possible failures for SNS SMS. Store a more detailed response for technical failures coming from SNS. The detailed provider response will be displayed on the admin UI.

It's safe to add columns on PostgreSQL 11+: https://leopard.in.ua/2016/09/20/safe-and-unsafe-operations-postgresql

## What's included

- Saving the provider response coming from SNS SMS when this is a technical failure
- Adding a new field to delivery callbacks
- Making sure that the provider response field is included when getting a notification status from the API

## Not yet

Those will come shortly after
- Admin PR
- Documentation PR

Trello: https://trello.com/c/IUiDX5bk/443-communicate-sms-opt-outs-to-services-via-dashboard-error


migrations/versions/0317_add_provider_response.py
```diff
@@ -0,0 +1,21 @@
...
+
+revision = '0317_add_provider_response'
+down_revision = '0316_now_live_template'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('notifications', sa.Column('provider_response', sa.Text(), nullable=True))
```
AntoineAugusti, line 17: About the choice of `text`

> There is no performance difference among these three types, apart from increased storage space when using the blank-padded type, and a few extra CPU cycles to check the length when storing into a length-constrained column. While character(n) has performance advantages in some other database systems, there is no such advantage in PostgreSQL; in fact character(n) is usually the slowest of the three because of its additional storage costs. In most situations text or character varying should be used instead.

https://www.postgresql.org/docs/11/datatype-character.html

QuanNguyen-CDS, line 17: does this mean that this table will be filled up with the same text over and over again from the provider? I guess it's overly complicated to suggest an enum/choice and serialize it as an int into the database?
I guess this status, e.g. NOTIFICATION_TECHNICAL_FAILURE is used everywhere else as well, so that's the problem...

QuanNguyen-CDS (approved):
Looks OK

migrations/versions/0317_add_provider_response.py
AntoineAugusti: This column will be filled for SMS with a technical failure status. We get less than 200 errors/month at the moment and this table only holds 7 days of data. As you pointed out, other status fields like the status store strings already which I guess is fine considering how cheap database storage is nowadays.

PR #1267, opened 2021-04-15
feat: catch other exceptions for base64 decode

AntoineAugusti:
Fixing this because we had 500 errors in production.

According to the documentation, https://docs.python.org/3/library/base64.html

> A binascii.Error exception is raised if s is incorrectly padded.

Catching other exceptions as well.


jimleroyer (approved):
LGTM

PR #1268, opened 2021-04-16
feat: give warning to users when they are approaching sending limits

AntoineAugusti:
As a user, I want to be alerted before I reach my daily sending limit, so that I can adjust accordingly / request a higher limit.

Send an email to team members of a service when:
- a service reaches 80% of its daily sending limit
- a service reaches its daily sending limit

Trello: https://trello.com/c/o0xXaKAD/444-give-warning-to-users-when-they-are-approaching-sending-limits

Screen Shot 2021-04-16 at 14 19 30


AntoineAugusti:
@QuanNguyen-CDS New Redis keys,

- nearing-{daily_limit_cache_key(service.id)}
- over-{daily_limit_cache_key(service.id)}

will be moved to [the Redis folder in the utils repo](https://github.com/cds-snc/notification-utils/blob/master/notifications_utils/clients/redis/__init__.py) just after. Those keys will likely be used on the admin in the future to display warnings so we need them in a shared space.

I'll remove the duplication here after.

app/notifications/validators.py
```diff
@@ -40,31 +42,67 @@ def check_service_over_api_rate_limit(service, api_key):
...
     check_service_over_api_rate_limit(service, api_key)
     check_service_over_daily_message_limit(api_key.key_type, service)
 
 
+def warn_about_daily_message_limit(service, messages_sent):
+    nearing_daily_message_limit = messages_sent >= .8 * service.message_limit
+    over_daily_message_limit = messages_sent >= service.message_limit
+
+    current_time = datetime.utcnow().isoformat()
+    cache_expiration = int(timedelta(days=1).total_seconds())
```
QuanNguyen-CDS, line 63: I think in a larger, high traffic system we should time bound this to the end of the day, but I think it's ok for today.

QuanNguyen-CDS (approved):
Looks pretty good - some minor comments but doesn't block approval

PR #1269, opened 2021-04-19
feat: new StatsD metrics to compute SLOs

AntoineAugusti:
Add new StatsD metrics to help compute SLOs. Some SLOs were only computable using the Notify database, make sure that those metrics are available on CloudWatch to make it easy for us to consolidate metrics and create a CloudWatch dashboards for SLOs.

Adds:
- Delay to send emails (multiple categories with/without attachments are according to process type (normal/bulk/priority)
- Delay to send SMS (multiple categories according to process type)
- Delay for processed jobs (do those start on time)

[See SLOs](https://docs.google.com/spreadsheets/d/1fU-FJ7THfWEqNhbpQ4r22MQipFwC7SP851ZQnOf68cM/edit)

Trello: https://trello.com/c/GFHzZGgz/453-measure-slos-for-the-first-time

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -25,6 +25,7 @@ from app import (
     DATETIME_FORMAT,
     encryption,
     notify_celery,
+    statsd_client,
 )
 from app.aws import s3
 from app.celery import (  # noqa: F401
@@ -104,6 +105,9 @@ def process_job(job_id, sender_id=None):
     job.job_status = JOB_STATUS_IN_PROGRESS
     job.processing_started = start
     dao_update_job(job)
+
+    # Record StatsD stats to compute SLOs
+    statsd_client.timing_with_dates('job.processing-start-delay', job.processing_started, job.scheduled_for)
 
     db_template = dao_get_template_by_id(job.template_id, job.template_version)
 
--- a/app/delivery/send_to_providers.py
+++ b/app/delivery/send_to_providers.py
@@ -53,10 +53,10 @@ def send_sms_to_provider(notification):
             notification.reply_to_text
         )
 
-        template_model = dao_get_template_by_id(notification.template_id, notification.template_version)
+        template_dict = dao_get_template_by_id(notification.template_id, notification.template_version).__dict__
 
         template = SMSMessageTemplate(
-            template_model.__dict__,
+            template_dict,
             values=notification.personalisation,
             prefix=service.name,
             show_prefix=service.prefix_sms,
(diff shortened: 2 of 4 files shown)


app/celery/tasks.py
```diff
@@ -105,6 +106,9 @@ def process_job(job_id, sender_id=None):
     job.processing_started = start
     dao_update_job(job)
 
+    # Record StatsD stats to compute SLOs
+    statsd_client.timing_with_dates('job.processing-start-delay', job.processing_started, job.scheduled_for)
```
QuanNguyen-CDS, line 110: Slightly confused here, `timing_with_dates` as a prototype of (stat, start, end, rate=1) however logging it here isn't actually the end of the job.
Should we be logging this after the `job_complete` ?

app/delivery/send_to_providers.py
```diff
@@ -84,8 +84,11 @@ def send_sms_to_provider(notification):
                 notification.billable_units = template.fragment_count
                 update_notification_to_sending(notification, provider)
 
-        delta_milliseconds = (datetime.utcnow() - notification.created_at).total_seconds() * 1000
-        statsd_client.timing("sms.total-time", delta_milliseconds)
+        # Record StatsD stats to compute SLOs
+        statsd_client.timing_with_dates("sms.total-time", notification.sent_at, notification.created_at)
```
QuanNguyen-CDS, line 88: this is also confusing to me, prototype of `timing_with_dates` is (stat, start, end, rate=1), shouldn't this be
`start = notification.created_at` and `end = time.now`

app/celery/tasks.py
AntoineAugusti: Our SLO is about the start of the job, not the end.

> Scheduled sends start on the time they were scheduled for.

Jobs can take a long time to complete if we're sending 40k notifications. What's important is that if people said they want those notifications to be delivered at 10am, we should start to send those at 10am

app/delivery/send_to_providers.py
AntoineAugusti: Start and end are reversed on utils, but I promise that the first parameter should be in the newer than the second parameter to have a positive value.

https://github.com/cds-snc/notification-utils/blob/9d4336b7578c4a5b16fe9979015d3f6ac6d83aa4/notifications_utils/clients/statsd/statsd_client.py#L82-L85

AntoineAugusti: Happy to submit a PR there

app/celery/tasks.py
QuanNguyen-CDS: ok go it, `processing-start-delay` - that makes sense

app/delivery/send_to_providers.py
AntoineAugusti: Submitted a utils PR https://github.com/cds-snc/notification-utils/pull/77

QuanNguyen-CDS (approved):
LGTM

PR #1271, opened 2021-04-19
fix: appropriate start date when recording stats for jobs

AntoineAugusti:
Fix a bug introduced in #1269, `scheduled_for` can be null if a job is not scheduled, which happens often.


jimleroyer (approved):
🚢  it

PR #1276, opened 2021-04-23
feat: add tests when generating callbacks

AntoineAugusti:
We had a Celery error in production surfacing this bug. The bug's root cause is a typo, but it was not caught by tests.

Fixing the bug and adding tests as a result.


QuanNguyen-CDS (approved):
looks OK

PR #1277, opened 2021-04-23
fix: adjust log level when email address is invalid

AntoineAugusti:
Logging an exception counts as an error, while those things are "normal".

Happens in production at the moment when an email address contains accents, SES does not support email addresses with accents.

See https://gcdigital.slack.com/archives/CNWA63606/p1619182249228800

### DIFF
--- a/app/celery/provider_tasks.py
+++ b/app/celery/provider_tasks.py
@@ -58,7 +58,9 @@ def deliver_email(self, notification_id):
             raise NoResultFound()
         send_to_providers.send_email_to_provider(notification)
     except InvalidEmailError as e:
-        current_app.logger.exception(e)
+        current_app.logger.info(
+            f"Cannot send notification {notification_id}, got an invalid email address: {str(e)}."
+        )
         update_notification_status_by_id(notification_id, 'technical-failure')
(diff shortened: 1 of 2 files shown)


app/celery/provider_tasks.py
```diff
@@ -58,7 +58,9 @@ def deliver_email(self, notification_id):
             raise NoResultFound()
         send_to_providers.send_email_to_provider(notification)
     except InvalidEmailError as e:
-        current_app.logger.exception(e)
+        current_app.logger.info(
+            f"Cannot send notification {notification_id}, got an invalid email address: {str(e)}."
```
QuanNguyen-CDS, line 62: does this stack trace have enough information to debug the problem (email address, etc.)?

AntoineAugusti: Yup, the source `InvalidEmailError` at the end of the message includes the email address and the message

https://github.com/cds-snc/notification-api/blob/a42942bb59a13cbdba074f5f380cd5d50e340d38/app/clients/email/aws_ses.py#L106-L110

QuanNguyen-CDS (approved):

PR #1280, opened 2021-04-28
feat: handle invalid iat value for JWT tokens

AntoineAugusti:
Fix a 500 error in production we just had where the `iat` value is not an integer.

### DIFF
--- a/tests/app/authentication/test_authentication.py
+++ b/tests/app/authentication/test_authentication.py
@@ -83,6 +83,28 @@ def test_auth_should_not_allow_request_with_no_iat(client, sample_api_key):
     assert exc.value.short_message == 'Invalid token: signature, api token not found'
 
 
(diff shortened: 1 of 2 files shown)


tests/app/authentication/test_authentication.py
```diff
@@ -83,6 +83,28 @@ def test_auth_should_not_allow_request_with_no_iat(client, sample_api_key):
     assert exc.value.short_message == 'Invalid token: signature, api token not found'
 
 
+@pytest.mark.parametrize('is_admin', [True, False], ids=['client', 'admin request'])
+def test_auth_should_not_allow_request_with_invalid_iat(client, sample_api_key, is_admin):
```
AntoineAugusti, line 87: Did not come up with a test like this, tests just before/after look similar

QuanNguyen-CDS (approved):
Looks Good

PR #1283, opened 2021-04-30
Bump Python dependencies

AntoineAugusti:
Took a look at https://github.com/alphagov/notifications-api/pull/3073 to facilitate this upgrade.

I'm not sure how to address LGTM or Snyk violations, if you know, happy to hear that.

PR #1285, opened 2021-05-03
fix: update pip in Dockerfile

AntoineAugusti:
[CI broke](https://github.com/cds-snc/notification-api/runs/2492286786?check_suite_focus=true) after merging #1283.

It couldn't install the `cryptography` package. The recommended way to fix this is to upgrade pip, which we do before running tests as well. https://github.com/cds-snc/notification-api/blob/2c8e2d2bfbd9c1c9c3da76606bf41dd131f85198/.github/workflows/test.yaml#L30

Did that and built the image locally without troubles


jimleroyer (approved):
:shipit:

PR #1290, opened 2021-05-06
notification emails to be pushed to another celery worker

QuanNguyen-CDS:
# Summary | Résumé
[Trello](https://trello.com/c/CSB8Cyy8/466-limit-sms-to-1-sec-per-phone-number-instead-of-1-sec-notify-wide)
feat: moving email to another celery worker instance
note: mfa email still go through the normal celery worker
fix: updated celery polling to sqs to 0.5 seconds

Testing setup:
- created a simple rest endpoint [flask] and it's able to do a modest 800 rps - tested with apache bench (ab)
- modified the awssesclient to send requests to this rest endpoint
- created a python script to push emails to notification-api, ran 3 instances to simulate load
Tested different settings locally while monitoring SQS queues on staging/aws
Tested with --pool=gevent, prefetch-multiplier[10,100,124], polling_interval [1,0.5,0.3], acks_late

Note: if we agree to move forward on this PR, it will not get committed until the notification-manifest PR gets created and approved (to be tested on staging)


AntoineAugusti (commented):
Could you share more details why you think it's worth tweaking `polling_interval`, `acks_late` and `concurrency`?

QuanNguyen-CDS:
> Could you share more details why you think it's worth tweaking `polling_interval`, `acks_late` and `concurrency`?

Yes, of course.

Poll Interval:
Setting up the rest server, i was able to also setup a manual metric for Requests Per Second. Running the email task locally and queuing hundreds of requests within SQS, It was observed that there were fluctuations within the processing of tasks. I'd describe it as "bursty". I had hypothesized that the celery tasks were sitting idle until the next poll cycle.
When I increased the poll cycle to 0.5 seconds. The "burstiness" observations went away.

Concurrency:
The sending of emails is highly specialized; A burst of emails thousands of emails in a short period of time. 
Additionally, the task itself is I/O bound:
- fetching of the notification from the database
- fetching the template from the database
- sending the data for attachments
- sending the email itself using the provider client
- and then another database update on notification
The purpose of the increase in concurrency is to try to take I/O bound tasks and spread them out. If the tasks were CPU bound then we would want to minimize the concurrency to the number of CPUs allocated. However we want to account for the task waiting for I/O actions to complete.

Ack Late:
watching this video: https://www.youtube.com/watch?v=ceJ-vy7fvus @ 13:52 gave me ideas to play around with -Ofair, prefetch_multiplier and ack_late. It seems they came to the conclusions of these settings while running into problems running celery tasks, slow and fast tasks. I played around with the prefetch multiplier but didn't have any real effects on the performance of the system. Reading the description for ack late seems to make sense to enable it: 

https://docs.celeryproject.org/en/stable/userguide/tasks.html
https://docs.celeryproject.org/en/stable/faq.html#faq-acks-late-vs-retry

The description 
_If your task is idempotent you can set the acks_late option to have the worker acknowledge the message after the task returns instead._

_The acks_late setting would be used when you need the task to be executed again if the worker (for some reason) crashes mid-execution. It’s important to note that the worker isn’t known to crash, and if it does it’s usually an unrecoverable error that requires human intervention (bug in the worker, or task code)._

sastels:
@jimleroyer @Moro-Code we've been doing a lot of work around this lately - is this still relevant?

jimleroyer:
@sastels Yes I'd say we should keep this. It seems like the tweaking can squeeze more performance. We could test thoroughly with out new load tests (which we didn't have yet when this PR was created).

jzbahrai:
@jimleroyer can we shut this?

jimleroyer:
Let's keep this please, there are good micro-optimizations that can help us that Quan did but we never merged in.

PR #1298, opened 2021-05-18
feat: notify service users when updating their daily limit

AntoineAugusti:
As a platform admin, I want to keep users posted when updating their daily limit of a live service. Send an email to service users when the daily limit of a live service is updated.

Trello: https://trello.com/c/3j0OXCit/514-send-email-to-users-when-a-service-daily-limit-is-updated


AntoineAugusti:
@anikbrazeau Adding you to review the content as shared on Slack

app/service/rest.py
```diff
@@ -274,6 +274,17 @@ def update_service(service_id):
...
+            send_notification_to_service_users(
+                service_id=service_id,
+                template_id=current_app.config['DAILY_LIMIT_UPDATED_TEMPLATE_ID'],
+                personalisation={
+                    'service_name': current_data['name'],
+                    'message_limit_en': '{:,}'.format(current_data['message_limit']),
+                    'message_limit_fr': '{:,}'.format(current_data['message_limit']).replace(',', ' ')
+                },
+                include_user_fields=['name']
+            )
```
jimleroyer, line 287: I would place this new logic in a new function to keep the already long one as short as we can.

tests/app/service/test_rest.py
```diff
@@ -2226,6 +2227,50 @@ def test_update_service_updating_daily_limit_clears_redis_cache(
...
 
 
+@pytest.mark.parametrize('is_trial, current_limit, new_limit, expected_call', [
+    (False, 1_000, 1_000, False),
+    (False, 1_000, 2_000, True),
+    (True, 50, 2_000, False),
+    (False, 1_000, 50, True),
+    (False, 1_000, None, False),
+    (True, 1_000, None, False),
+])
```
jimleroyer, line 2237: Great cases 👍

app/service/rest.py
AntoineAugusti: Done in the next commit, thanks for the suggestion

jimleroyer (approved):
:shipit:

PR #1302, opened 2021-06-02
fix: set NOTIFY_ENVIRONMENT in staging

AntoineAugusti:
Rework #1301

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -472,7 +472,7 @@ class Production(Config):
 
 
 class Staging(Production):
-    pass
+    NOTIFY_ENVIRONMENT = "staging"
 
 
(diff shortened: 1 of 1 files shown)


app/config.py
```diff
@@ -472,7 +472,7 @@ class Production(Config):
 
 
 class Staging(Production):
-    pass
+    NOTIFY_ENVIRONMENT = "staging"
```
AntoineAugusti, line 475: If you look at lines above, `NOTIFY_ENVIRONMENT` was set to a constant in production and not using the env variable. 😢

jimleroyer (approved):

PR #1303, opened 2021-06-03
Run black on codebase

AntoineAugusti:
Same thing than done in https://github.com/cds-snc/notification-admin/pull/1049

The first commit is only configuration file (easy to review) while the second one runs `black` on the codebase


scripts/run_tests.sh
```diff
@@ -23,8 +23,14 @@ function display_result {
 make test-requirements
 display_result $? 1 "Requirements check"
 
+black ./app ./tests --check
```
AntoineAugusti: Changed to `black --config pyproject.toml --check .` after

scripts/run_tests.sh
```diff
@@ -23,8 +23,14 @@ function display_result {
...
+display_result $? 2 "Code style check (Black)"
+
 flake8 .
-display_result $? 1 "Code style check"
+display_result $? 1 "Code style check (flake8)"
+
+isort --check-only ./app ./tests
```
AntoineAugusti: Changed to `isort --check-only .` after to leverage the config file

scripts/run_tests.sh
smcmurtry: Nice!

scripts/run_tests.sh
smcmurtry: interesting - I guess we never had an isort check on this repo?

smcmurtry (approved):
Looks good Antoine!

scripts/run_tests.sh
AntoineAugusti: Yeah, `isort` is an addition to the API!

PR #1307, opened 2021-06-04
fix: appropriate status for sms in reports

AntoineAugusti:
## Steps to Reproduce
1) Sent an SMS to a CA number
2) Go to the list of notifications in the UI
3) Status show "Sent internationally" instead of "Sent"

Example URL https://notification.canada.ca/services/de598a67-4bf1-4ea0-b515-8014c36b4734/download-notifications.csv?message_type=sms&status=sending%2Cdelivered%2Cfailed

## Expected Behaviour
Status is "Sent"

Trello: https://trello.com/c/VWLjKJwr/678-reports-show-sent-sent-internationally-for-sms

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1670,7 +1670,7 @@ class Notification(db.Model):
                 "delivered": "Delivered",
                 "sending": "Sending",
                 "created": "Sending",
-                "sent": "Sent internationally",
+                "sent": "Sent",
             },
             "letter": {
                 "technical-failure": "Technical failure",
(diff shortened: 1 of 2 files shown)


app/models.py
```diff
@@ -1670,7 +1670,7 @@ def formatted_status(self):
                 "delivered": "Delivered",
                 "sending": "Sending",
                 "created": "Sending",
-                "sent": "Sent internationally",
```
AntoineAugusti, line 1673: I believe it was set to this value because in the past, without delivery receipts, the status was `sending` and not `sent`. Anyway this does not make sense anymore since we can have `sent` to `delivered`

jimleroyer (approved):

PR #1296, opened 2021-05-14
Added route notification to bulk queue for spreadsheets larger than 1,000 rows

jimleroyer:
# Summary | Résumé

[Route notifications sent using spreadsheets to the bulk queue](https://trello.com/c/rZEFQezH/472-route-notifications-sent-using-spreadsheets-to-the-bulk-queue)

The feature covers both **email** and **SMS** notifications but **not letters**.

The `CSV_BULK_REDIRECT_THRESHOLD` key configuration was added to make the threshold of spreadsheet rows configurable.

# Test instructions | Instructions pour tester la modification

Unit tests covers the case where spreadsheet is smaller or larger than the CSV threshold to redirect to bulk queue. 

# Reviewer checklist | Liste de vérification du réviseur

This is a suggested checklist of questions reviewers might ask during their
review | Voici une suggestion de liste de vérification comprenant des questions
que les réviseurs pourraient poser pendant leur examen :

- [ ] Does this meet a user need? | Est-ce que ça répond à un besoin utilisateur?
- [ ] Is it accessible? | Est-ce que c’est accessible?
- [ ] Is it translated between both offical languages? | Est-ce dans les deux
      langues officielles?
- [ ] Is the code maintainable? | Est-ce que le code peut être maintenu?
- [ ] Have you tested it? | L’avez-vous testé?
- [ ] Are there automated tests? | Y a-t-il des tests automatisés?
- [ ] Does this cause automated test coverage to drop? | Est-ce que ça entraîne
      une baisse de la quantité de code couvert par les tests automatisés?
- [ ] Does this break existing functionality? | Est-ce que ça brise une
      fonctionnalité existante?
- [ ] Should this be split into smaller PRs to decrease change risk? | Est-ce
      que ça devrait être divisé en de plus petites demandes de tirage (« pull
      requests ») afin de réduire le risque lié aux modifications?
- [ ] Does this change the privacy policy? | Est-ce que ça entraîne une
      modification de la politique de confidentialité?
- [ ] Does this introduce any security concerns? | Est-ce que ça introduit des
      préoccupations liées à la sécurité?
- [ ] Does this significantly alter performance? | Est-ce que ça modifie de
      façon importante la performance?
- [ ] What is the risk level of using added dependencies? | Quel est le degré de
      risque d’utiliser des dépendances ajoutées?
- [ ] Should any documentation be updated as a result of this? (i.e. README
      setup, etc.) | Faudra-t-il mettre à jour la documentation à la suite de ce
      changement (fichier README, etc.)?

PR #1308, opened 2021-06-05
fix: always route SMS to throttled queue when sender is set

AntoineAugusti:
## Current behaviour
- I have a service with a custom SMS sender
- I upload a spreadsheet sending more than 200 SMS
- Those SMS are routed to the usual SMS queue and are sent faster than 1 SMS per second

## Expected behaviour
SMS are throttled at 1 SMS per second no matter what, to respect telecommunication rate limits.

## Description

Make sure that SMS with a custom sender are always routed to the throttled SMS SQS queue, even if another queue was requested.

This makes sure that we don't go over rate limits for each long code number.

We had a problem in production because a client sent a large number of SMS through a spreadsheet and they were routed to the bulk queue thanks to the recent work done in https://github.com/cds-snc/notification-api/pull/1296.


tests/app/celery/test_tasks.py
```diff
@@ -584,6 +584,32 @@ def test_should_route_save_sms_task_to_bulk_on_large_csv_file(notify_db, notify_
     assert mocked_deliver_sms.called
 
 
+def test_should_route_save_sms_task_to_throttled_queue_on_large_csv_file_if_custom_sms_sender(
+    notify_db, notify_db_session, mocker
+):
+    service = create_service_with_defined_sms_sender(sms_sender_value="3433061234")
```
AntoineAugusti, line 590: Setting a default SMS sender with a custom number

tests/app/celery/test_tasks.py
```diff
@@ -584,6 +584,32 @@ def test_should_route_save_sms_task_to_bulk_on_large_csv_file(notify_db, notify_
...
+        notification_id,
+        encryption.encrypt(notification),
+    )
+
+    persisted_notification = Notification.query.one()
+    provider_tasks.deliver_throttled_sms.apply_async.assert_called_once_with(
+        [str(persisted_notification.id)], queue="send-throttled-sms-tasks"
+    )
+    mocked_deliver_sms.assert_not_called()
+    mocked_deliver_throttled_sms.assert_called_once()
```
AntoineAugusti, line 610: Making sure this is routed to the appropriate queue

tests/app/notifications/test_process_notification.py
```diff
@@ -289,7 +289,7 @@ def test_persist_notification_increments_cache_if_key_exists(sample_template, sa
             "sms",
             "normal",
             "+14383898585",
-            "research-mode-tasks",
+            "send-throttled-sms-tasks",
```
AntoineAugusti, line 292: Does not care that this is research mode and routed to the throttled queue

tests/app/notifications/test_process_notification.py
```diff
@@ -356,7 +356,7 @@ def test_persist_notification_increments_cache_if_key_exists(sample_template, sa
             "sms",
             "normal",
             "+14383898585",
-            "notify-internal-tasks",
+            "send-throttled-sms-tasks",
```
AntoineAugusti, line 359: Does not care that a custom queue was requested, still routed to the throttled queue

app/notifications/process_notifications.py
```diff
@@ -125,8 +125,7 @@ def send_notification_to_queue(notification, research_mode, queue=None):
         deliver_task = provider_tasks.deliver_sms
         if notification.sends_with_custom_number():
             deliver_task = provider_tasks.deliver_throttled_sms
-            if not queue:
-                queue = QueueNames.SEND_THROTTLED_SMS
+            queue = QueueNames.SEND_THROTTLED_SMS
```
AntoineAugusti, line 128: Removing the `if` is the actual fix

maxneuvians (approved):
LGTM - nice work!

PR #1305, opened 2021-06-04
API bulk sending

AntoineAugusti:
This PR adds a new API endpoint to send notifications in bulk, up to 50k recipients at a time per template. Bulk notifications can be scheduled up to 4 days in the future. This replicates what's possible to do through the admin already.

The HTTP controller uses a similar logic than the admin:
- validating the payload
- uploading a CSV-like payload to S3
- saving a `Job` to the database, which will be processed by Celery later on (saving notifications to the database and sending those)

Because this uses jobs, bulk requests show up already in the UI. But we [will need to make a few tweaks](https://gcdigital.slack.com/archives/C023Y2HBK5M/p1623252377018300).

## Later on
- changes to the admin UI to account for bulk jobs from the admin or the API
- API documentation in EN & FR

## Tests

Tested a fair amount of payloads locally, on top of an extensive test suite. Validating/storing a payload of 50k recipients takes 6s on my machine, which I think is acceptable.

<details>
 <summary>Python code to test locally</summary>

```py
import requests
from datetime import datetime, timedelta
import csv
from io import StringIO
import itertools


def rows_to_csv(rows):
    output = StringIO()
    writer = csv.writer(output)
    writer.writerows(rows)
    return output.getvalue()


def job_line(nb):
    return list(itertools.repeat(["antoine.augusti+test@cds-snc.ca", "Antoine 🇫🇷"], nb))


API_KEY = ""

response = requests.post(
    "http://localhost:6011/v2/notifications/bulk",
    json={
        "name": f"My bulk name {datetime.utcnow().isoformat()}",
        "template_id": "659a214f-dfec-4882-9242-fea0bd502a09",
        "csv": rows_to_csv([["email address", "name"], *job_line(3)]),
        "scheduled_for": (datetime.utcnow() + timedelta(hours=12)).isoformat(),
    },
    headers={"Authorization": f"ApiKey-v1 {API_KEY[-36:]}"},
)
print(response.json())

```
</details>

Trello: https://trello.com/c/IOgSfVro/554-api-endpoint-for-a-gc-developer-to-pass-csv-data

PR #1312, opened 2021-06-17
Fix sender ID for bulk API

AntoineAugusti:
Follow up of https://github.com/cds-snc/notification-api/pull/1305

The `sender_id` was not properly set. I was using previously a method `get_reply_to_text` which was used by another controller method and assumed it was returning a UUID. I was wrong.

I began to write tests for sender IDs in `test_post_bulk_creates_job_and_dispatches_celery_task`, using a Pytest param `use_sender_id`, but actually never used this parameter in my test case 😭 

This PR fixes how `sender_id` is set on `Job`s and adds a bunch of tests proving that it works as intended.

### DIFF
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -1,6 +1,7 @@
 import base64
 import csv
 import functools
+import uuid
 from io import StringIO
 
 import werkzeug
@@ -129,6 +130,8 @@ def post_bulk():
     check_service_has_permission(template.template_type, authenticated_service.permissions)
 
     remaining_messages = authenticated_service.message_limit - fetch_todays_total_message_count(authenticated_service.id)
+
+    form["validated_sender_id"] = validate_sender_id(template, form.get("reply_to_id"))
 
     try:
         if form.get("rows"):
@@ -406,6 +409,29 @@ def process_precompiled_letter_notifications(*, letter_data, api_key, template, 
     return notification
 
 
+def validate_sender_id(template, reply_to_id):
+    notification_type = template.template_type
+
+    if notification_type == EMAIL_TYPE:
+        service_email_reply_to_id = reply_to_id
+        check_service_email_reply_to_id(
+            str(authenticated_service.id),
+            service_email_reply_to_id,
+            notification_type,
+        )
+        return service_email_reply_to_id
+    elif notification_type == SMS_TYPE:
+        service_sms_sender_id = reply_to_id
+        check_service_sms_sender_id(
+            str(authenticated_service.id),
+            service_sms_sender_id,
+            notification_type,
+        )
+        return service_sms_sender_id
+    else:
+        raise NotImplementedError("validate_sender_id only handles emails and text messages")
+
+
 def get_reply_to_text(notification_type, form, template, form_field=None):
     reply_to = None
     if notification_type == EMAIL_TYPE:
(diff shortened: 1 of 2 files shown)


tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -1555,8 +1627,9 @@ def test_post_bulk_flags_rows_with_errors(client, notify_db, notify_db_session,
 @pytest.mark.parametrize("data_type", ["rows", "csv"])
 @pytest.mark.parametrize("is_scheduled", [True, False])
 @pytest.mark.parametrize("use_sender_id", [True, False])
```
AntoineAugusti, line 1629: Was declared already, but not actually used in the test case previously. That was the mistake, otherwise we would have seen the bug earlier.

app/v2/notifications/post_notifications.py
```diff
@@ -406,6 +409,29 @@ def process_precompiled_letter_notifications(*, letter_data, api_key, template,
     return notification
 
 
+def validate_sender_id(template, reply_to_id):
```
AntoineAugusti, line 412: This is similar to `get_reply_to_text` just below, but it returns a UUID instead of the reply value (for emails) or sender for SMS

jimleroyer: A bit similar to `check_reply_to` in the `validators.py` file too if I get this right. Why do we hide the reply-to address behind a UUID? to hide the address itself when possible?

tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -1214,6 +1214,78 @@ def test_post_bulk_with_invalid_data_arguments(
...
+    )
+
+    assert response.status_code == 400
+    error_json = json.loads(response.get_data(as_text=True))
+    assert error_json["errors"] == [
+        {
+            "error": "ValidationError",
+            "message": "reply_to_id is not a valid UUID",
+        }
+    ]
```
jimleroyer, line 1238: 👍

tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -1214,6 +1214,78 @@ def test_post_bulk_with_invalid_data_arguments(
...
+    )
+
+    assert response.status_code == 400
+    error_json = json.loads(response.get_data(as_text=True))
+    assert error_json["errors"] == [
+        {
+            "error": "BadRequestError",
+            "message": f"sms_sender_id {fake_uuid} does not exist in database for service id {sms_code_template.service_id}",
+        }
+    ]
```
jimleroyer, line 1286: Good tests to increase coverage!

tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -1568,6 +1641,11 @@ def test_post_bulk_creates_job_and_dispatches_celery_task(
     if is_scheduled:
         scheduled_for = datetime.utcnow() + timedelta(days=1)
         data["scheduled_for"] = scheduled_for.isoformat()
+    if has_default_reply_to:
+        create_reply_to_email(sample_email_template.service, "test@test.com")
+    if use_sender_id:
+        reply_to_email = create_reply_to_email(sample_email_template.service, "custom@test.com", is_default=False)
+        data["reply_to_id"] = reply_to_email.id
```
jimleroyer, line 1648: It's nice to read the tests sometimes and getting more what we support as features.

jimleroyer (approved):
Good stuff. Have a few questions, mostly around my understanding of the code.

app/v2/notifications/post_notifications.py
AntoineAugusti: This is mostly to follow what's done already on other API endpoints, when sending an email for example. You pass UUID for templates, reply to etc.

The UI is clear about the email address and its UUID. You can also update a reply to and keep the same email address.
Screen Shot 2021-06-22 at 08 22 06

PR #1314, opened 2021-06-23
fix: have PR bot use github token

Moro-Code:
# Summary | Résumé

# ARCHIVE 

Instead of using a personal token. Github now provides the `GITHUB_TOKEN` to use the github bot. So I switched the workflow to use the this for the PR bot. Of course the only way to test this is to do a PR...

# NEW 

Created github app called Notify PR Bot. Installed it on our repos and modified the workflow to authenticate with it.

Lets see if this PR works 

N/A

Does it work ?

      langues officielles?
      fonctionnalité existante?

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -64,6 +64,22 @@ jobs:
         kubectl set image deployment.apps/celery celery=$DOCKER_SLUG:${GITHUB_SHA::7} -n=notification-canada-ca --kubeconfig=$HOME/.kube/config
         kubectl set image deployment.apps/celery-beat celery-beat=$DOCKER_SLUG:${GITHUB_SHA::7} -n=notification-canada-ca --kubeconfig=$HOME/.kube/config
         kubectl set image deployment.apps/celery-sms celery-sms=$DOCKER_SLUG:${GITHUB_SHA::7} -n=notification-canada-ca --kubeconfig=$HOME/.kube/config
+
+    - name: Obtain a GitHub App Installation Access Token
+      id: githubAppAuth
+      run: |
+        TOKEN="$(npx obtain-github-app-installation-access-token ci ${{ secrets.TOKEN }})"
+        echo "::add-mask::$TOKEN"
+        echo "::set-output name=token::$TOKEN"
+
+    - name: Test the obtained token
+      run: |
+        curl -X POST -H 'Content-Type: application/json' \
+          -d '{"context":"test","state":"success"}' \
+          "https://api.github.com/repos/$GITHUB_REPOSITORY/statuses/$GITHUB_SHA?access_token=$GITHUB_TOKEN"
+      env:
+        GITHUB_TOKEN: ${{ steps.githubAppAuth.outputs.token }}
+
     - uses: cds-snc/notification-pr-bot@master
       env:
-        TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        TOKEN: ${{ steps.githubAppAuth.outputs.token }}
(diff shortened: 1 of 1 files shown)


maxneuvians:
The PR bot opens PR in repo's that are other than their own. I do not think the GITHUB_TOKEN will have permissions to do that - but you can give it a try (https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token).

.github/workflows/docker.yaml
```diff
@@ -66,4 +66,4 @@ jobs:
         kubectl set image deployment.apps/celery-sms celery-sms=$DOCKER_SLUG:${GITHUB_SHA::7} -n=notification-canada-ca --kubeconfig=$HOME/.kube/config
     - uses: cds-snc/notification-pr-bot@master
       env:
-        TOKEN: ${{ secrets.TOKEN }}
+        TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jimleroyer, line 69: We should contact one of the organization's GitHub admin, to know what's the default permissions on GITHUB_TOKEN per repository and through the organization. If my understanding is correct, we need to know of the token's default permissions. It might be restricted to repository or through the whole organization. 

It might not be something desirable too from an admin perspective as we'd have permissions to write pull requests in all repositories under CDS, if my understanding is correct. But maybe we can make our case for PR bot usages? We are not the only team with one.

https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token

Moro-Code:
@jimleroyer we would just need to update the value for the token here. I have shared this with you

.github/workflows/docker.yaml
```diff
@@ -64,6 +64,22 @@ jobs:
...
+        echo "::add-mask::$TOKEN"
+        echo "::set-output name=token::$TOKEN"
+
+    - name: Test the obtained token
+      run: |
+        curl -X POST -H 'Content-Type: application/json' \
+          -d '{"context":"test","state":"success"}' \
+          "https://api.github.com/repos/$GITHUB_REPOSITORY/statuses/$GITHUB_SHA?access_token=$GITHUB_TOKEN"
+      env:
+        GITHUB_TOKEN: ${{ steps.githubAppAuth.outputs.token }}
```
jimleroyer, line 81: Did you test the commands replicated by these steps locally?

.github/workflows/docker.yaml
```diff
@@ -64,6 +64,22 @@ jobs:
         kubectl set image deployment.apps/celery celery=$DOCKER_SLUG:${GITHUB_SHA::7} -n=notification-canada-ca --kubeconfig=$HOME/.kube/config
         kubectl set image deployment.apps/celery-beat celery-beat=$DOCKER_SLUG:${GITHUB_SHA::7} -n=notification-canada-ca --kubeconfig=$HOME/.kube/config
         kubectl set image deployment.apps/celery-sms celery-sms=$DOCKER_SLUG:${GITHUB_SHA::7} -n=notification-canada-ca --kubeconfig=$HOME/.kube/config
+
+    - name: Obtain a GitHub App Installation Access Token
+      id: githubAppAuth
+      run: |
+        TOKEN="$(npx obtain-github-app-installation-access-token ci ${{ secrets.TOKEN }})"
```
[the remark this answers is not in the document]
Moro-Code: No it's a token. I shared access with you on last pass

.github/workflows/docker.yaml
Moro-Code: Probably a good idea

AntoineAugusti:
Note that if you're going this route you need to do a similar change on *all other repos*: admin, API, document download API/frontend and documentation. They all have a similar `docker.yml` script

jimleroyer (approved):
:shipit:

PR #1320, opened 2021-07-07
Adding mypy to CI

smcmurtry:
# Summary | Résumé

This PR adds mypy to the CI script in notification-api, similar to what we have in notification-admin: https://github.com/cds-snc/notification-admin/blob/master/scripts/run_tests.sh#L37

~I will convert this from a draft once I fix (or ignore!) the remaining errors mypy is throwing.~ Done! 

Ready for review.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -6,7 +6,7 @@ import uuid
 from time import monotonic
 
 from dotenv import load_dotenv
-from flask import _request_ctx_stack, g, jsonify, make_response, request
+from flask import _request_ctx_stack, g, jsonify, make_response, request # type: ignore
 from flask_marshmallow import Marshmallow
 from flask_migrate import Migrate
 from notifications_utils import logging, request_helper
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -6,7 +6,7 @@ import uuid
 from time import monotonic
 
 from dotenv import load_dotenv
-from flask import _request_ctx_stack, g, jsonify, make_response, request # type: ignore
+from flask import _request_ctx_stack, g, jsonify, make_response, request  # type: ignore
 from flask_marshmallow import Marshmallow
 from flask_migrate import Migrate
 from notifications_utils import logging, request_helper
--- a/app/authentication/bearer_auth.py
+++ b/app/authentication/bearer_auth.py
@@ -7,6 +7,6 @@ class BearerAuth(requests.auth.AuthBase):
     def __init__(self, token: str):
         self.token = token
 
-    def __call__(self, r: requests.Request):
+    def __call__(self, r):
         r.headers["Authorization"] = "Bearer " + self.token
         return r
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -596,7 +596,7 @@ def process_incomplete_job(job_id):
     job_complete(job, resumed=True)
 
 
-def queue_to_use(notifications_count: int) -> str:
+def queue_to_use(notifications_count: int):
     """Determine which queue to use depending on given parameters.
 
     We only check one rule at the moment: if the CSV file is big enough,
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -1,6 +1,7 @@
 import json
 from collections import defaultdict, namedtuple
 from datetime import datetime
+from typing import Optional
 
 from flask import current_app
 from notifications_utils.columns import Row
@@ -596,7 +597,7 @@ def process_incomplete_job(job_id):
     job_complete(job, resumed=True)
 
 
-def queue_to_use(notifications_count: int):
+def queue_to_use(notifications_count: int) -> Optional[str]:
     """Determine which queue to use depending on given parameters.
 
     We only check one rule at the moment: if the CSV file is big enough,
--- a/app/dbsetup.py
+++ b/app/dbsetup.py
@@ -1,7 +1,7 @@
 from functools import cached_property, partial
 from typing import Optional
 
-from flask import _app_ctx_stack # type: ignore
+from flask import _app_ctx_stack  # type: ignore
 from flask_sqlalchemy import BaseQuery, SignallingSession, SQLAlchemy, get_state
 from sqlalchemy import orm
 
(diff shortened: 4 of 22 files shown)


app/models.py
```diff
@@ -76,7 +77,7 @@ def update_from_original(self, original):
                 current_app.logger.debug("{} has no column {} to copy from".format(original, c.name))
 
 
-class User(db.Model):
+class User(db.Model):  # type: ignore
```
smcmurtry, line 80: This file gave me the most trouble. `db.Model` is a type rather than a variable (or something like that?) so MyPy complains about using it as a base class, even though this seems to be the recommended way to use `flask-sqlalchemy`: https://flask-sqlalchemy.palletsprojects.com/en/2.x/models/

I tried the suggested work around (https://github.com/dropbox/sqlalchemy-stubs/issues/76#issuecomment-595839159) but that didn't work. I tried ignoring the whole file, but that caused problems in other files because the db models no longer had type definitions. So I am resorting to adding comments that tell mypy to ignore individual lines.

app/models.py
```diff
@@ -76,7 +78,10 @@ def update_from_original(self, original):
                 current_app.logger.debug("{} has no column {} to copy from".format(original, c.name))
 
 
-class User(db.Model):
+BaseModel: DefaultMeta = db.Model
```
smcmurtry, line 81: This file gave me some trouble. `db.Model` is a type rather than a variable (or something like that?) so MyPy complains about using it as a base class, even though this seems to be the recommended way to use flask-sqlalchemy: https://flask-sqlalchemy.palletsprojects.com/en/2.x/models/

I tried the work around suggested here: dropbox/sqlalchemy-stubs#76 (comment) and it worked!

requirements_for_test.txt
```diff
@@ -18,3 +18,5 @@ rfc3987==1.3.8
 jinja2-cli[yaml]==0.6.0
 black==21.5b2
 locust==1.5.3
+mypy==0.812
+sqlalchemy-stubs==0.4
```
smcmurtry, line 22: type definitions for `sqlalchemy`

app/__init__.py
```diff
@@ -6,7 +6,7 @@
 from time import monotonic
 
 from dotenv import load_dotenv
-from flask import _request_ctx_stack, g, jsonify, make_response, request
+from flask import _request_ctx_stack, g, jsonify, make_response, request  # type: ignore
```
smcmurtry, line 9: `_request_ctx_stack` is missing from the type definitions for flask for some reason

mypy.ini
```diff
@@ -0,0 +1,92 @@
+[mypy]
+python_version = 3.9
+
+[mypy-pytest.*]
+ignore_missing_imports = True
```
smcmurtry, line 5: Ignoring imports that we don't have type definitions for

app/dbsetup.py
```diff
@@ -12,7 +13,7 @@ class ExplicitRoutingSession(SignallingSession):
     then the `reader` bind will get returned instead.
     """
 
-    _name = None
+    _name: Optional[str] = None
```
jimleroyer, line 16: 🕺

tests/app/v2/notifications/test_notification_schemas.py
```diff
@@ -26,7 +24,7 @@
 }
 
 
-@pytest.mark.parametrize("input", [valid_get_json, valid_get_with_optionals_json])
+@pytest.mark.parametrize("input", [{}, valid_get_with_optionals_json])
```
jimleroyer, line 27: Hmm while the `valid_get_json` was just an empty dictionary, labeling it as such with a variable name might be helpful to indicate that this was a valid JSON. I'll let you decide though, I am not familiar with the test code but I do like verbose naming that might clear out what's the goal of the `input`.

jimleroyer (approved):
Thanks for bringing stronger type safety to the API.

app/authentication/bearer_auth.py
```diff
@@ -7,6 +7,6 @@ class BearerAuth(requests.auth.AuthBase):
     def __init__(self, token: str):
         self.token = token
 
-    def __call__(self, r: requests.Request):
+    def __call__(self, r):
```
Moro-Code, line 10: just curious why a type annotation was removed

app/celery/tasks.py
```diff
@@ -596,7 +596,7 @@ def process_incomplete_job(job_id):
     job_complete(job, resumed=True)
 
 
-def queue_to_use(notifications_count: int) -> str:
+def queue_to_use(notifications_count: int):
```
Moro-Code, line 599: Same thing here... were these type annotations wrong ?

app/schemas.py
```diff
@@ -62,7 +63,7 @@ def _validate_datetime_not_in_past(dte, msg="Date cannot be in the past"):
         raise ValidationError(msg)
 
 
-class BaseSchema(ma.SQLAlchemyAutoSchema):
+class BaseSchema(ma.SQLAlchemyAutoSchema):  # type: ignore
```
Moro-Code, line 66: Should SQLAlchemyAutoScheme be imported as well like Schema ?

Moro-Code (commented):
This is awesome work ! I think overtime as we integrate type annotations more and more into the code base this will help tremendously reduce bugs and make code suggestions much better !

tests/app/v2/notifications/test_notification_schemas.py
smcmurtry: Good point! I will revert this change and add a type definition

app/authentication/bearer_auth.py
smcmurtry: `requests.Request` was not the correct type - `requests.auth.AuthBase` required a parent class of `requests.Request` for the type

app/celery/tasks.py
smcmurtry: This function could return `None` so the `str` was not the correct return type. I will change this so the return type is `Optional[str]`.

app/schemas.py
smcmurtry: I tried that but it caused a lot of tests to fail. I think we are initiallizing some things on our instance of Marshmallow that need to be available in these classes.

PR #1326, opened 2021-07-20
Update to the latest version of notifications-utils

smcmurtry:
# Summary | Résumé

Here are the release notes: https://github.com/cds-snc/notification-utils/releases/tag/44.0.0

To test, I ran this locally, sent email containing language tags, and then verified that the email I received had the correct html.


sastels (approved):
Works on my iPhone!

PR #1329, opened 2021-07-21
Stop passing around sender_id

sastels:
# Summary | Résumé

https://trello.com/c/Ijn2D73k/576-bulk-api-follow-up-improve-how-sender-id-is-passed

We moved `sender_id` into the job object, so we don't need to pass it around as a kwarg.

Notes:
Currently `process_job()` is only called in
- `run_scheduled_jobs()`
- `create_job()`
- `create_bulk_job()`

Note that it is never called with a `sender_id`. 

In `process_job()` we just use `sender_id` to send to `process_row()`. Since it gets the job, we can pull out the `sender_id` there instead of passing it in.

`process_row()` is called in
- `process_job()`
- `process_incomplete_job()`

In `process_incomplete_job()`, `sender_id` is not set when `process_row()` is called.

So we can stop passing `sender_id` around and set it in `process_row()`

Send some emails and make sure they go through?

In a given job, `save_email()` and `save_sms()` are doing the same database reads each time they're called. We could probably just pull out stuff in `process_job()` and pass it down through the encrypted object. For example, the `reply_to_text` would be easy to move. I think it's better to get into this in a separate PR though.
 
      langues officielles?
      fonctionnalité existante?

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -81,10 +81,9 @@ from app.utils import get_csv_max_rows
 
 @notify_celery.task(name="process-job")
 @statsd(namespace="tasks")
-def process_job(job_id, sender_id=None):
+def process_job(job_id):
     start = datetime.utcnow()
     job = dao_get_job_by_id(job_id)
-    sender_id = str(job.sender_id) if job.sender_id else sender_id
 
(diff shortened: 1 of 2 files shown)


jimleroyer (approved):
LGTM

PR #1337, opened 2021-07-30
Feat/url allowlist file attachment

mohamed-cds:
This PR is to add additional checks when providing URL's to requests urlOpen as well as enforcing https URLs when sending notifications with links. _**This was previous reviewed but the PR got lost while i was away, so starting over.**_

#### Why these changes are being suggested
Validating URL's is a best practise when writing secure code. Although this piece of code itself has a low probability of user controlled data reaching urlopen, code can be copied or refactored leading to user control of the URL parameter in the future. 

Rejecting requests that don't begin with "s3" will effectively reduce the change of any future security risks.

#### References
[CWE-74](https://cwe.mitre.org/data/definitions/74.html) - Improper Neutralization of Special Elements in Output Used by a Downstream Component ('Injection')

#### Example CVE with the same underlying issue
CVE-2019-18348

Send a notification with a file attachment to validate this feature is still operating correctly

      langues officielles?
      fonctionnalité existante?


app/delivery/send_to_providers.py
```diff
@@ -132,9 +132,11 @@ def send_email_to_provider(notification):
 
             if sending_method == "attach":
                 try:
-
+                    # Prevent URL patterns like file:// ftp:// that may lead to security vulnerabilities
+                    if not personalisation_data[key]["document"]["direct_file_url"].lower().startswith(("http", "s3")):
+                        raise ValueError("Invalid URL provided")
```
jimleroyer, line 137: My quick take on it is that we might want to validate this earlier if a `direct_file_url` is non secure. Applying the validation logic in that part of the code would raise an exception in a celery task. This error would not get back to the user. The celery task will get retried until the retries run out by the task and potentially throw celery errors alerts. 

Or do we want to have some defensive programming in here "just in case"?

It is certainly a good validation to have overall. Let me dig more into the code but it's my current understanding, I am not sure if we apply or not that validation already on the user front, we need to check.

tests/app/delivery/test_send_to_providers.py
```diff
@@ -888,6 +888,50 @@ def test_notification_document_with_pdf_attachment(
...
+    }
+    if filename_attribute_present:
+        personalisation["file"]["document"]["filename"] = filename
+        personalisation["file"]["document"]["sending_method"] = "attach"
+    else:
+        personalisation["file"]["document"]["sending_method"] = "link"
+
+    db_notification = create_notification(template=template, personalisation=personalisation)
+
+    # See https://stackoverflow.com/a/34929900
```
jimleroyer, line 924: I would provide a summary of the issue here along with the SO link.

app/delivery/send_to_providers.py
mohamed-cds: i'd leave this here just in-case. The test should flag this if we refactor.

If we can also track this upstream that's even better. Can users ever provide a URL themselves when setting notificaions? or will this always contain a s3 bucket url pointing to whatever file they uploaded? If a user can provide the URL directly then more work will have to be done upstream for sure.

tests/app/delivery/test_send_to_providers.py
mohamed-cds: Copied this down from test_notification_document_with_pdf_attachment. Not sure what the underlying issue, just that i seem to need it to get the test to work :)

app/delivery/send_to_providers.py
```diff
@@ -131,10 +131,12 @@ def send_email_to_provider(notification):
                     raise MalwarePendingException
 
             if sending_method == "attach":
+                # Prevent URL patterns like file:// ftp:// that may lead to security vulnerabilities
+                if not personalisation_data[key]["document"]["direct_file_url"].lower().startswith(("http", "s3")):
+                    raise ValueError("Invalid URL provided")
```
Moro-Code, line 136: We should have an except statement for ValueError in `deliver_sms` and `deliver_email` such that sending the notification is not retried. I'm happy to discuss

mohamed-cds: added

app/delivery/send_to_providers.py
jimleroyer: Hmm the `direct_file_url` entry might always target a S3 bucket if I read the code right, when the sending method of the notification is `attach`. But Notify does allow for linking (when sending method is `link`) and that would rather be on the `url` entry in the same dictionary. Realistically, then, I would expect the checked URL in this to be HTTPS and target S3 (to double check via some local debugging).

The `url` entry through provided by the user could be non HTTPS and could be a risk for people receiving the email. I don't [see anything in the documentation](https://documentation.notification.canada.ca/en/send.html#sending-a-file-by-email) that requires clients to provide a HTTPS link though. What do you think?

tests/app/delivery/test_send_to_providers.py
jimleroyer: Damn...

app/delivery/send_to_providers.py
mohamed-cds: I like it, that's even more well defined. I'll update this PR to restrict `url` to "https" and `direct_file_url` to "s3".

We can then update our docs to mention that https links are only allowed. This may cause problems for any GC departments still using http, but might also be an incentive to migrate to https

mohamed-cds:
Replaced by #1344

PR #1341, opened 2021-08-05
Removing the "Code already sent, wait 10 seconds" page

smcmurtry:
# Summary | Résumé

Trello: https://trello.com/c/LwlEusgQ/639-bug-cannot-create-an-account-with-2fa-stuck-in-a-loop

Before: https://www.youtube.com/watch?v=IYtNStxSfCE

After: https://www.youtube.com/watch?v=XCf495Lwukc

This change is to make a 2nd request for a 2FA code within 10 seconds return a 204 instead of a 400. This will prevent users from seeing the "Code already sent, wait 10 seconds" page. And it will keep the behaviour where a 2nd code is not sent in this 10 second window to address the original intent described by Max: https://github.com/cds-snc/notification-api/pull/1341#discussion_r683859199


app/user/rest.py
```diff
@@ -269,12 +269,11 @@ def verify_user_code(user_id):
 def send_user_2fa_code(user_id, code_type):
     user_to_send_to = get_user_by_id(user_id=user_id)
 
-    if verify_within_time(user_to_send_to, age=timedelta(seconds=10)) >= 1:
-        raise InvalidRequest("Code already sent, wait 10 seconds", status_code=400)
-
```
jimleroyer, line 274: Digging into github's commit history to know of the original intent of this, I saw that @maxneuvians actually added this. I am not sure if he will remember the reason but it could be good to sync with him to know why we thought this was necessary. The [change comes from here](https://github.com/cds-snc/notification-api/commit/b5c184a81522b3010730d2e72867745c700f4248).

maxneuvians: https://github.com/cds-snc/notification-api/pull/164 - basically people were jamming the login button and because there is a delay in SMS messages coming to phones they would get multiple ones and would never know which one is correct.

smcmurtry: Yes, so my change would prevent a new code from being sent, without raising an error

PR #1346, opened 2021-08-10
feat: add an entrypoint for lambda

maxneuvians:
Adds a lambda entrypoint for the container:

Can be build this 
```
docker build -t XXXXX.dkr.ecr.ca-central-1.amazonaws.com/notify/api:test --build-arg GIT_SHA=abcd1234 -f  ci/Dockerfile.lambda .
```


smcmurtry (approved):
Looks good to me! This change doesn't effect our current deployment pipeline, and the [awsgi](https://github.com/youngrok/awsgi) dependency is only about 1 MB. So I think the downsides are minimal and it will let us continue to test the lambda function approach.

PR #1348, opened 2021-08-10
Add a continuous deployment step for the lambda image

smcmurtry:
# Summary | Résumé

Most of this is copied from https://github.com/cds-snc/scan-websites/blob/main/.github/workflows/build_and_push.yml. All I did was copy in the name of our ECR repository in staging (`notify/api`) and set the branch name to "master".

This uses a new env variable that I haven't added yet: our aws account number (`AWS_ACCOUNT`).

I'm sure I missed some things so will keep this a draft for now.

### DIFF
--- a/.github/workflows/build_and_push.yml
+++ b/.github/workflows/build_and_push.yml
@@ -8,7 +8,7 @@ on:
 
 env:
   GITHUB_SHA: ${{ github.sha }}
-  REGISTRY: ${{ secrets.AWS_ACCOUNT }}.dkr.ecr.ca-central-1.amazonaws.com/notify/api
+  REGISTRY: ${{ secrets.AWS_ACCOUNT }}.dkr.ecr.ca-central-1.amazonaws.com/notify
 
 jobs:
   changes:
@@ -71,9 +71,3 @@ jobs:
           aws lambda update-function-code \
             --function-name ${{ matrix.image }} \
             --image-uri $REGISTRY/${{ matrix.image }}:latest
-
-      - name: Migrate Database
-        if: ${{ matrix.image == 'api' }}
-        run: |
-          source .github/workflows/scripts/migrate.sh
-          migrate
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_push.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Logout of Amazon ECR
+        run: docker logout ${{ steps.login-ecr.outputs.registry }}
+
+      - name: Deploy lambda
+        run: |
+          aws lambda update-function-code \
+            --function-name ${{ matrix.image }} \
+            --image-uri $REGISTRY/${{ matrix.image }}:latest
+
+      - name: Migrate Database
```
maxneuvians, line 75: I would remove this step as you don't have a script that actually does the migration :)

.github/workflows/build_and_push.yml
```diff
@@ -0,0 +1,79 @@
...
+name: Build and Push Container to ECR, deploy to lambda
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [master]
+
+env:
+  GITHUB_SHA: ${{ github.sha }}
+  REGISTRY: ${{ secrets.AWS_ACCOUNT }}.dkr.ecr.ca-central-1.amazonaws.com/notify/api
```
maxneuvians, line 11: i would leave this a `.../notify` because on line 73 you are adding a name for the image

maxneuvians (changes requested):

.github/workflows/build_and_push.yml
```diff
@@ -0,0 +1,79 @@
...
+      images: ${{ steps.filter.outputs.changes }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
+            api: 'app/**'
```
smcmurtry, line 26: This action should run when any of the api code changes. I changed this from `api: 'api/**'` since our api code is in the app folder. Does the key name here (`api`) get used for anything?

.github/workflows/build_and_push.yml
smcmurtry: whoops!

maxneuvians (approved):
LGTM

PR #1349, opened 2021-08-11
Added an argument so docker can find Dockerfile.lambda

smcmurtry:
# Summary | Résumé

This is to fix the failing workflow: https://github.com/cds-snc/notification-api/runs/3303137409

### DIFF
--- a/.github/workflows/build_and_push.yml
+++ b/.github/workflows/build_and_push.yml
@@ -39,7 +39,6 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Build container
-        working-directory: ./${{ matrix.image }}
         run: |
           docker build \
           --build-arg git_sha=$GITHUB_SHA \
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_push.yml
```diff
@@ -44,7 +44,8 @@ jobs:
           docker build \
           --build-arg git_sha=$GITHUB_SHA \
           -t $REGISTRY/${{ matrix.image }}:$GITHUB_SHA \
-          -t $REGISTRY/${{ matrix.image }}:latest .
+          -t $REGISTRY/${{ matrix.image }}:latest . \
+          -f ci/Dockerfile.lambda
```
maxneuvians, line 48: line 42 will put you in the `api` directory as defined on line 26, which doesn't exist. I would recommend removing line 42.

PR #1354, opened 2021-08-19
Added pull request template

jimleroyer:
# Summary | Résumé

Created a pull request template that is more aligned with Notify's need. I removed a few superfluous items in the reviewer's checklist that should be determined earlier in the development process, removed the out of scope and help requests sections as we rarely fill these and we raise these in Slack earlier as well usually.

There is a new section around release plan for that one feature so that we don't miss necessary extra steps. It should not be filled in usually but it's important for the time when it will need to, as notes for the person doing releases.

      langues officielles?
      fonctionnalité existante?


pull_request_template.md
```diff
@@ -0,0 +1,68 @@
...
+      modification de la politique de confidentialité?
+- [ ] Does this introduce any security concerns? | Est-ce que ça introduit des
+      préoccupations liées à la sécurité?
+- [ ] Does this significantly alter performance? | Est-ce que ça modifie de
+      façon importante la performance?
+- [ ] What is the risk level of using added dependencies? | Quel est le degré de
+      risque d’utiliser des dépendances ajoutées?
+- [ ] Should any documentation be updated as a result of this? (i.e. README
+      setup, etc.) | Faudra-t-il mettre à jour la documentation à la suite de ce
+      changement (fichier README, etc.)?
```
sastels, line 68: Should we cut down this reviewer section even more? If no one ever goes through and checks off boxes, it's a bit useless. For example, for api we can probably cut out a11y and translations. Or, for translations, something like "are you changing a string that needs to be translated in admin?"

I guess, I feel that most of these things should be thought of (and probably checked off or commented on) by the author. If this is basically the CDS template then maybe this discussion has already been had and settled.

sastels (approved):
Looks fine, minor comments / rant.

pull_request_template.md
jimleroyer: Yeah you are right, we could do a more targeted pull request template depending on the code base / repository. I'll remove these two items that you mentioned. 

Regarding the CDS standard, it's good, but not good enough if we don't make use of it. We might care for it a bit more if we build it ourselves.. and we might realize later that we'll need to include back some items that was initially there. Let's walk the path and see where it goes.

PR #1370, opened 2021-09-07
[WIP] Define load tests for one off, bulk  email and SMS

iokpala:
[Trello card](https://trello.com/c/xkZD3br3/618-define-and-load-test-plans-api-bulk-api-in-staging)

# Summary | Résumé

This PR seeks to define load tests to test the one off email endpoint[1].

Prior to this, an example load tests already exists. This effort expands the one off email endpoint load test to cover attachment and link.

This choice is influenced by the already existing smoke tests for one off email.

Locust insists on have a task defined[2], which it uses to simulate the to run the task, this can be weighted which yields probable selection based on the weight.

The developers understand, would be that of isolated tasks that would run on their own greenlet(thread), hence this has led to the definition of separate class in order to achieve the aforementioned.

[1] \<domain>/v2/notifications/email
[2] https://docs.locust.io/en/stable/writing-a-locustfile.html

> workon notifications-api
>  locust --headless -u 1000 -r 100 -f tests-perf/locust/locust-notifications.py

- [ ] Load testing can be executed

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -21,3 +21,5 @@ AWS_ACCESS_KEY_ID=
 AWS_SECRET_ACCESS_KEY=
 
 AWS_PINPOINT_REGION=us-west-2
+
+LOAD_TEST_DOMAIN=https://api.staging.notification.cdssandbox.xyz
--- a/.env.example
+++ b/.env.example
@@ -21,5 +21,3 @@ AWS_ACCESS_KEY_ID=
 AWS_SECRET_ACCESS_KEY=
 
 AWS_PINPOINT_REGION=us-west-2
-
-LOAD_TEST_DOMAIN=https://api.staging.notification.cdssandbox.xyz
--- a/tests-perf/locust/locust-notifications.py
+++ b/tests-perf/locust/locust-notifications.py
@@ -63,7 +63,7 @@ class NotifyApiUser(HttpUser):
 
         self.client.post("/v2/notifications/bulk", json=self.json, headers=self.headers)
 
-    @task(12)
+    @task(16)
     def send_sms_notifications(self):
         self.json = {
             "phone_number": 12897682684,
--- a/tests-perf/locust/locust-notifications.py
+++ b/tests-perf/locust/locust-notifications.py
@@ -34,7 +34,7 @@ class NotifyApiUser(HttpUser):
 
         self.headers = {"Authorization": os.getenv("TEST_AUTH_HEADER")}
         self.email = os.getenv("LOAD_TEST_EMAIL", "success@simulator.amazonses.com")
-        self.phone_number = os.getenv("LOAD_TEST_PHONE_NUMBER", "12897682684")
+        self.phone_number = os.getenv("LOAD_TEST_PHONE_NUMBER", "16132532222")
         self.template_group = NotifyApiUserTemplateGroup(
             bulk_email_id=os.getenv("LOAD_TEST_BULK_EMAIL_TEMPLATE_ID", "5ebee3b7-63c0-4052-a8cb-387b818df627"),
             email_id=os.getenv("LOAD_TEST_EMAIL_TEMPLATE_ID", "a59b313d-8de2-4973-ac2f-66de7ec0b239"),
--- a/tests-perf/locust/locust-notifications.py
+++ b/tests-perf/locust/locust-notifications.py
@@ -1,11 +1,12 @@
 """ locust-notifications.py
-   isort:skip_file
+    isort:skip_file
 """
 # flake8: noqa
 
 import os
 import sys
 from datetime import datetime
+from dataclasses import make_dataclass
 
 sys.path.append(os.path.abspath(os.path.join("..", "tests_smoke")))
 
@@ -14,6 +15,13 @@ from locust import HttpUser, constant_pacing, task
 from tests_smoke.smoke.common import job_line, rows_to_csv  # type: ignore
 
 load_dotenv()
+NotifyApiUserTemplateGroup = make_dataclass('NotifyApiUserTemplateGroup', [
+    'bulk_email_id',
+    'email_id',
+    'email_with_attachment_id',
+    'email_with_link_id',
+    'sms_id',
+])
 
 
 class NotifyApiUser(HttpUser):
@@ -25,26 +33,32 @@ class NotifyApiUser(HttpUser):
         super(NotifyApiUser, self).__init__(*args, **kwargs)
 
         self.headers = {"Authorization": os.getenv("TEST_AUTH_HEADER")}
-        self.email = "success@simulator.amazonses.com"
-        self.template_id = "5ebee3b7-63c0-4052-a8cb-387b818df627"
+        self.email = os.getenv("LOAD_TEST_EMAIL", "success@simulator.amazonses.com")
+        self.phone_number = os.getenv("LOAD_TEST_PHONE_NUMBER", "12897682684")
+        self.template_group = NotifyApiUserTemplateGroup(
+            bulk_email_id=os.getenv("LOAD_TEST_BULK_EMAIL_TEMPLATE_ID", "5ebee3b7-63c0-4052-a8cb-387b818df627"),
+            email_id=os.getenv("LOAD_TEST_EMAIL_TEMPLATE_ID", "a59b313d-8de2-4973-ac2f-66de7ec0b239"),
+            email_with_attachment_id=os.getenv("LOAD_TEST_EMAIL_WITH_ATTACHMENT_TEMPLATE_ID", "a59b313d-8de2-4973-ac2f-66de7ec0b239"),
+            email_with_link_id=os.getenv("LOAD_TEST_EMAIL_WITH_LINK_TEMPLATE_ID", "5ebee3b7-63c0-4052-a8cb-387b818df627"),
+            sms_id=os.getenv("LOAD_TEST_SMS_TEMPLATE_ID", "83d01f06-a818-4134-bd69-ce90a2949280"),
+        )
 
     @task(1)
     def send_email_notifications(self):
-        json = self.__email_json("a59b313d-8de2-4973-ac2f-66de7ec0b239")
+        json = self.__email_json(self.template_group.email_id)
 
         self.client.post("/v2/notifications/email", json=json, headers=self.headers)
 
     @task(2)
     def send_email_with_attachment_notifications(self):
         personalisation = {
-            "application_file": "arbitrary text for this variable",
             "attached_file": {
                 "file": "Q29udGVudCBvZiBBdHRhY2hlZCBmaWxl",
                 "filename": "attached_file.txt",
                 "sending_method": "attach",
             }
         }
-        json = self.__email_json(self.template_id, personalisation)
+        json = self.__email_json(self.template_group.email_with_attachment_id, personalisation)
 
         self.client.post("/v2/notifications/email", json=json, headers=self.headers)
 
@@ -57,7 +71,7 @@ class NotifyApiUser(HttpUser):
                 "sending_method": "link",
             }
         }
-        json = self.__email_json(self.template_id, personalisation)
+        json = self.__email_json(self.template_group.email_with_link_id, personalisation)
 
         self.client.post("/v2/notifications/email", json=json, headers=self.headers)
 
@@ -65,7 +79,7 @@ class NotifyApiUser(HttpUser):
     def send_bulk_notifications(self):
         json = {
             "name": f"My bulk name {datetime.utcnow().isoformat()}",
-            "template_id": self.template_id,
+            "template_id": self.template_group.bulk_email_id,
             "csv": rows_to_csv([["email address", "application_file"], *job_line(self.email, 2)])
         }
 
@@ -74,8 +88,8 @@ class NotifyApiUser(HttpUser):
     @task(16)
     def send_sms_notifications(self):
         json = {
-            "phone_number": "12897682684",
-            "template_id": "83d01f06-a818-4134-bd69-ce90a2949280"
+            "phone_number": self.phone_number,
+            "template_id": self.template_group.sms_id
         }
 
         self.client.post("/v2/notifications/sms", json=json, headers=self.headers)
--- a/tests-perf/locust/locust-notifications.py
+++ b/tests-perf/locust/locust-notifications.py
@@ -19,63 +19,64 @@ class NotifyApiUser(HttpUser):
         super(NotifyApiUser, self).__init__(*args, **kwargs)
 
         self.headers = {"Authorization": AUTH_HEADER}
-        self.template_id = "5ebee3b7-63c0-4052-a8cb-387b818df627"
         self.email = "success@simulator.amazonses.com"
-        self.json = {
-            "email_address": self.email,
-            "template_id": self.template_id,
-            "personalisation": {},
-        }
+        self.template_id = "5ebee3b7-63c0-4052-a8cb-387b818df627"
 
     @task(1)
     def send_email_notifications(self):
-        self.json = {
-            "email_address": self.email,
-            "template_id": "a59b313d-8de2-4973-ac2f-66de7ec0b239",
-            "personalisation": {},
-        }
+        json = self.__email_json("a59b313d-8de2-4973-ac2f-66de7ec0b239")
 
-        self.client.post("/v2/notifications/email", json=self.json, headers=self.headers)
+        self.client.post("/v2/notifications/email", json=json, headers=self.headers)
 
     @task(2)
     def send_email_with_attachment_notifications(self):
-        self.json["personalisation"] = {
+        personalisation = {
             "application_file": {
                 "file": "Q29udGVudCBvZiBBdHRhY2hlZCBmaWxl",
                 "filename": "attached_file.txt",
                 "sending_method": "attach",
             }
         }
+        json = self.__email_json(self.template_id, personalisation)
 
-        self.client.post("/v2/notifications/email", json=self.json, headers=self.headers)
+        self.client.post("/v2/notifications/email", json=json, headers=self.headers)
 
     @task(4)
     def send_email_with_link_notifications(self):
-        self.json["personalisation"] = {
+        personalisation = {
             "application_file": {
                 "file": "Q29udGVudCBvZiBBdHRhY2hlZCBmaWxl",
                 "filename": "attached_file.txt",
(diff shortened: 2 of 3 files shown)


tests-perf/locust/locust-notifications.py
```diff
@@ -12,13 +12,50 @@ class NotifyApiUser(HttpUser):
...
+
+class OneOffEmailWithFileAttachmentMessage(NotifyApiUser):
+    @tag("one-off-email-attached")
+    @task
+    def send_one_off_email_with_attachment_notifications(self):
+        self.json["personalisation"] = {
+            "application_file": {
+                "file": "Q29udGVudCBvZiBBdHRhY2hlZCBmaWxl",
+                "filename": "attached_file.txt",
+                "sending_method": "attach",
```
iokpala, line 43: Followed the schema definitions [1] and the smoke tests.. still getting 400 bad requests 

[1] https://github.com/cds-snc/notification-api/blob/master/app/schema_validation/definitions.py#L22-L42

tests-perf/locust/locust-notifications.py
```diff
@@ -12,13 +12,50 @@ class NotifyApiUser(HttpUser):
...
+
+        self.headers = {"Authorization": AUTH_HEADER}
+        self.json = {
             "email_address": "success@simulator.amazonses.com",
             "template_id": "9c17633c-126a-4ad3-ad2f-b14c3a85314a",
             "personalisation": {"colour": "Fulvous"},
         }
 
-        self.client.post("/v2/notifications/email", json=json, headers=headers)
+    @task
```
jimleroyer, line 25: If we have the same configuration for all user classes, we can have all `task` annotations in the same class and put different weight in each to determine which one would get used, [per the documentation.](https://docs.locust.io/en/stable/writing-a-locustfile.html#task-decorator)

jimleroyer:
Do we plan to add the SMS coverage in a different PR or in this one later?

iokpala:
@jimleroyer SMS can be part of this PR..

tests-perf/locust/locust-notifications.py
iokpala: That was defined in error and part of debugging.. I remember seeing the example as part of the smoke tests.. https://github.com/cds-snc/notification-api/blob/4dcb11eb1f893b4bfa4f2142a7ff8b7b0f462d32/tests_smoke/smoke/test_api_one_off.py#L38

tests-perf/locust/locust-notifications.py
```diff
@@ -12,13 +12,50 @@ class NotifyApiUser(HttpUser):
...
+    def task(self):
+        pass
+
+class OneOffEmailMessage(NotifyApiUser):
```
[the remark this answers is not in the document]
iokpala: Ok will make that change

tests-perf/locust/locust-notifications.py
```diff
@@ -12,13 +12,50 @@ class NotifyApiUser(HttpUser):
...
+                "file": "Q29udGVudCBvZiBBdHRhY2hlZCBmaWxl",
+                "filename": "attached_file.txt",
+                "sending_method": "attach",
+            }
+        }
+
+        self.client.post("/v2/notifications/email", json=self.json, headers=self.headers)
+
+class OneOffEmailWithLinkMessage(NotifyApiUser):
+    @tag("one-off-email-link")
```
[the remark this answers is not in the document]
iokpala: Whilst exploring the documentation for locust[1] for tag, i suppose it appeared useful to target individual tasks

[1] https://docs.locust.io/en/stable/writing-a-locustfile.html#tag-decorator

tests-perf/locust/locust-notifications.py
iokpala: @jimleroyer I have updated the templated id to `5ebee3b7-63c0-4052-a8cb-387b818df627`..

jimleroyer: Ok that one is on me. The services need to have a switch turned on in the admin settings to be  able to send attachments. It was disabled for this service:
image

Can you try again?

iokpala: Still getting the same 400 bad request error

tests-perf/locust/locust-notifications.py
```diff
@@ -12,13 +12,40 @@ class NotifyApiUser(HttpUser):
...
+                "sending_method": "attach",
+            }
+        }
+
+        self.client.post("/v2/notifications/email", json=self.json, headers=self.headers)
+
+    @task(4)
+    def send_email_with_link_notifications(self):
+        self.json["personalisation"] = {
+            "var": {
```
jimleroyer, line 44: This `var` would fail here as the template would need it. `application_file` seems like the proper variable name to use.

tests-perf/locust/locust-notifications.py
jimleroyer: I was able to make it work for me locally. I changed the code a bit to get rid of `self.json` and inline explicitly each content message. I also observed that there is an inexistent `var` used somewhere that would need to be switched to `application_file`.

tests-perf/locust/locust-notifications.py
iokpala: Ok will trying changing bar for application_file.. Cheers

tests-perf/locust/locust-notifications.py
```diff
@@ -1,24 +1,81 @@
...
 
-    @task
-    def send_notifications(self):
-        headers = {"Authorization": AUTH_HEADER}
+    def __init__(self, *args, **kwargs):
+        super(NotifyApiUser, self).__init__(*args, **kwargs)
+
+        self.headers = {"Authorization": os.getenv("TEST_AUTH_HEADER")}
+        self.email = "success@simulator.amazonses.com"
+        self.template_id = "5ebee3b7-63c0-4052-a8cb-387b818df627"
```
sastels, line 22: might be a good idea to move these template id's to env vars

tests-perf/locust/locust-notifications.py
```diff
@@ -1,24 +1,81 @@
...
+        json = {
+            "name": f"My bulk name {datetime.utcnow().isoformat()}",
+            "template_id": self.template_id,
+            "csv": rows_to_csv([["email address", "application_file"], *job_line(self.email, 2)])
+        }
+
+        self.client.post("/v2/notifications/bulk", json=json, headers=self.headers)
+
+    @task(16)
+    def send_sms_notifications(self):
```
sastels, line 68: Do we need to load test sending sms on a regular basis? They cost $0.01 each, and sending thousands a day will add up.

tests-perf/locust/locust-notifications.py
iokpala: Yeah that could work.. It would require having them defined as template groups.. Example: EMAIL_TEMPLATE, EMAIL_WITH_ATTACHMENT_TEMPLATE, SMS_TEMPLATE etc etc??

tests-perf/locust/locust-notifications.py
iokpala: Hello Steve, I gathered that using this number 12897682684 is free

sastels: no, it still costs to send to it.

tests-perf/locust/locust-notifications.py
sastels: yeah, maybe prefaced by `LOAD_TEST_` or something, ie `LOAD_TEST_EMAIL_TEMPLATE`, etc.

sastels (changes requested):
Looks good! only comments
- I think we should pull out those id's into variables, 
- maybe think about whether we should be load testing SMS? This PR doesn't put it into CI, so this last issue is probably for a future discussion.

iokpala:
>Looks good! only comments
@sastels Cheers
>I think we should pull out those id's into variables,

Will look into getting this done.. However it would mean grouping the environment variables.

>maybe think about whether we should be load testing SMS? 

I am not sure.. I defined tests on SMS based on the requirements as defined in the trello card 

>This PR doesn't put it into CI, so this last issue is probably for a future discussion.

This isn't meant for CI.. the next PR will be focused on this

tests-perf/locust/locust-notifications.py
```diff
@@ -1,24 +1,88 @@
...
+
+    @task(2)
+    def send_email_with_attachment_notifications(self):
+        personalisation = {
+            "application_file": "arbitrary text for this variable",
+            "attached_file": {
+                "file": "Q29udGVudCBvZiBBdHRhY2hlZCBmaWxl",
+                "filename": "attached_file.txt",
+                "sending_method": "attach",
+            }
```
jimleroyer, line 45: Hmm I didn't expect the `personalisation` to be under this format for attached sending method, from the documentation at least. I need to revisit this.

tests-perf/locust/locust-notifications.py
```diff
@@ -1,24 +1,88 @@
...
+    host = os.getenv("LOAD_TEST_DOMAIN", "https://api.staging.notification.cdssandbox.xyz")
+
+    def __init__(self, *args, **kwargs):
+        super(NotifyApiUser, self).__init__(*args, **kwargs)
+
+        self.headers = {"Authorization": os.getenv("TEST_AUTH_HEADER")}
+        self.email = "success@simulator.amazonses.com"
+        self.template_id = "5ebee3b7-63c0-4052-a8cb-387b818df627"
+
+    @task(1)
```
jimleroyer, line 31: I would give more weight to this task given it's the one that is the most used.

tests-perf/locust/locust-notifications.py
```diff
@@ -1,24 +1,88 @@
...
+                "file": "Q29udGVudCBvZiBBdHRhY2hlZCBmaWxl",
+                "filename": "attached_file.txt",
+                "sending_method": "link",
+            }
+        }
+        json = self.__email_json(self.template_id, personalisation)
+
+        self.client.post("/v2/notifications/email", json=json, headers=self.headers)
+
+    @task(8)
```
jimleroyer, line 64: This could have less weight as a task, 1 or 2 I think.

tests-perf/locust/locust-notifications.py
jimleroyer: We can use a number as listed [in the smoke testing section of the documentation](https://documentation.notification.canada.ca/en/testing.html#smoke-testing), that won't really get sent. One could argue that we should test the whole execution path but it's the most we can do without incurring any cost.

.env.example
```diff
@@ -21,3 +21,5 @@ AWS_ACCESS_KEY_ID=
 AWS_SECRET_ACCESS_KEY=
 
 AWS_PINPOINT_REGION=us-west-2
+
+LOAD_TEST_DOMAIN=https://api.staging.notification.cdssandbox.xyz
```
jimleroyer, line 25: This would not be necessary I think.. as the `.env` file that we read are actually in the *locust* folder and not at the root. This won't get used by the main Flask application, right?

iokpala: Yes you are right.. It is defined there as an example.

tests-perf/locust/locust-notifications.py
iokpala: Thanks looking forward to your findings

tests-perf/locust/locust-notifications.py
iokpala: This[1] suggests to my understanding that `@task(1)` has the highest chance of getting called

[1] https://docs.locust.io/en/stable/writing-a-locustfile.html#task-decorator

tests-perf/locust/locust-notifications.py
iokpala: Ok will update it

tests-perf/locust/locust-notifications.py
sastels: I believe it's the opposite - the higher the number, the more it's called. For example, when I ran the code with 100 users it was mostly the sms task (weight 16)
```
Name                                  # reqs     
--------------------------------------------
 POST /v2/notifications/bulk             23
 POST /v2/notifications/email            19
 POST /v2/notifications/sms              58
--------------------------------------------
 Aggregated                             100
```
I'd say weight the `/email` at something like 10 and the rest at 1?

sastels (changes requested):
This looks awesome! I think we just need to sort out those task weights and we'll be ready to get this merged! 🎉

.env.example
sastels: After this merges we should have a PR that reorgs/refactors the code a bit so that the smoke tests and locust tests can share env vars (ie templates). But that should definitely not be done in this PR.

iokpala:
@sastels With the weight in reverse.. here are the results:
NB: Done over 3 circles.. fairly consistent.

```
Name                                                          # reqs      # fails  |     Avg     Min     Max  Median  |   req/s failures/s
--------------------------------------------------------------------------------------------------------------------------------------------
 POST /v2/notifications/bulk                                        5     0(0.00%)  |     616     432     864     630  |    0.50    0.00
 POST /v2/notifications/email                                      91     0(0.00%)  |     515     243    1264     500  |    8.25    0.00
 POST /v2/notifications/sms                                         4     0(0.00%)  |     466     435     498     460  |    0.12    0.00
--------------------------------------------------------------------------------------------------------------------------------------------
 Aggregated                                                       100     0(0.00%)  |     518     243    1264     500  |    8.88    0.00

 Name                                                          # reqs      # fails  |     Avg     Min     Max  Median  |   req/s failures/s
--------------------------------------------------------------------------------------------------------------------------------------------
 POST /v2/notifications/bulk                                        5     0(0.00%)  |     616     432     864     630  |    0.50    0.00
 POST /v2/notifications/email                                      91     0(0.00%)  |     515     243    1264     500  |    8.25    0.00
 POST /v2/notifications/sms                                         4     0(0.00%)  |     466     435     498     460  |    0.12    0.00
--------------------------------------------------------------------------------------------------------------------------------------------
 Aggregated                                                       100     0(0.00%)  |     518     243    1264     500  |    8.88    0.00

 Name                                                          # reqs      # fails  |     Avg     Min     Max  Median  |   req/s failures/s
--------------------------------------------------------------------------------------------------------------------------------------------
 POST /v2/notifications/bulk                                        5     0(0.00%)  |     616     432     864     630  |    0.50    0.00
 POST /v2/notifications/email                                      91     0(0.00%)  |     515     243    1264     500  |    8.25    0.00
 POST /v2/notifications/sms                                         4     0(0.00%)  |     466     435     498     460  |    0.12    0.00
--------------------------------------------------------------------------------------------------------------------------------------------

```

tests-perf/locust/locust-notifications.py
```diff
@@ -1,24 +1,102 @@
...
-            "personalisation": {"colour": "Fulvous"},
+            "name": f"My bulk name {datetime.utcnow().isoformat()}",
+            "template_id": self.template_group.bulk_email_id,
+            "csv": rows_to_csv([["email address", "application_file"], *job_line(self.email, 2)])
         }
 
-        self.client.post("/v2/notifications/email", json=json, headers=headers)
+        self.client.post("/v2/notifications/bulk", json=json, headers=self.headers)
+
+    @task(1)
```
jimleroyer, line 88: Can we give a bigger weight to the SMS? something bigger than email with files but a bit lower than email with no attachment test?

tests-perf/locust/locust-notifications.py
```diff
@@ -1,24 +1,102 @@
...
     wait_time = constant_pacing(60)
-    host = "https://api.staging.notification.cdssandbox.xyz"
+    host = os.getenv("LOAD_TEST_DOMAIN", "https://api.staging.notification.cdssandbox.xyz")
+
+    def __init__(self, *args, **kwargs):
+        super(NotifyApiUser, self).__init__(*args, **kwargs)
+
+        self.headers = {"Authorization": os.getenv("TEST_AUTH_HEADER")}
+        self.email = os.getenv("LOAD_TEST_EMAIL", "success@simulator.amazonses.com")
+        self.phone_number = os.getenv("LOAD_TEST_PHONE_NUMBER", "12897682684")
```
jimleroyer, line 37: That phone number might incur a cost but we can use one listed [on that page](https://documentation.notification.canada.ca/en/testing.html) which will skip the sending part.

tests-perf/locust/locust-notifications.py
jimleroyer: Bump here, is this discussion considered resolved? Are we okay for testing with SMS using the test phone numbers?

tests-perf/locust/locust-notifications.py
iokpala: Increased it to 10.. Here are the results 

```
Name                                                          # reqs      # fails  |     Avg     Min     Max  Median  |   req/s failures/s
--------------------------------------------------------------------------------------------------------------------------------------------
 POST /v2/notifications/bulk                                        5     0(0.00%)  |     614     572     669     620  |    0.47    0.00
 POST /v2/notifications/email                                      68     0(0.00%)  |     466     257     727     450  |    6.43    0.00
 POST /v2/notifications/sms                                        27     0(0.00%)  |     405     213     491     410  |    2.55    0.00
--------------------------------------------------------------------------------------------------------------------------------------------
```

tests-perf/locust/locust-notifications.py
iokpala: This has been changed

tests-perf/locust/locust-notifications.py
sastels: The test # won't cost us anything, though at the same time they won't create a notification and test the backend. IMO we should create real Notify test numbers / addresses that create a notification and do everything except the SES / SMS send (at that point the notifications will be just marked as delivered or failed as desired). This will allow us and our users to fully test our / their systems. This is out of scope of this PR though.

sastels (approved):
Looks good to me! :shipit:

tests-perf/locust/locust-notifications.py
jimleroyer: We discussed this offline. The simulated emails will test throughout AWS backend but the simulated SMS will simply stop at the API. It would be nice to have the SMS to go deeper into the pipeline to have a better performance tests on these but that would requires further changes into Notify. @sastels [created this task](https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/cds-snc/notification-planning/36) to capture necessary changes.

jimleroyer (approved):
:shipit:

PR #1386, opened 2021-10-25
Feature: Execute daily the performance tests

iokpala:
# Related to https://github.com/cds-snc/notification-terraform/pull/305

# Summary | Résumé

To get daily performance tests going, we have identified the need for a dedicated Docker file for this purpose.

This PR defines a new Docker file from which a runtime

Docker container will be created for usage in an AWS lambda.

With are defined the following environment variables:

- LOAD_TEST_PHONE_NUMBER
- LOAD_TEST_EMAIL
- LOAD_TEST_DOMAIN
- LOAD_TEST_SMS_TEMPLATE_ID
- LOAD_TEST_BULK_EMAIL_TEMPLATE_ID
- LOAD_TEST_EMAIL_TEMPLATE_ID
- LOAD_TEST_EMAIL_WITH_ATTACHMENT_TEMPLATE_ID
- LOAD_TEST_EMAIL_WITH_LINK_TEMPLATE_ID
- TEST_AUTH_HEADER

The aforelisted are used in [1], and these have been assigned for staging specific values.

Given the definition within the Docker file, these environment variables can overridden when this container is instantiated.

The performance tests defined in [1], also reuse some code from test_smoke hence its inclusion.

Also within is defined a build, push and deploy steps for performance tests Docker images going forward.

It is triggered only when changes are detected to test-perf and tests_smoke folders/paths[2].

The locust performance tests resuse code define under tests_smoke.

Most of these lines of YAML code is adapted from [3] and follows predefined steps that are very similar to the goals of this commit.

[1] tests-perf/locust/locust-notifications.py
[2] https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onpushpull_requestpaths
[3] .github/workflows/build_and_push.yml

- [ ] Automated workflows with Github

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -1,8 +1,7 @@
 name: Notify Performance / Load Tests
 
-on: []
-  # schedule:
-  #   - cron: '0 0 * * *' # every day at midnight 
+on:
+  workflow_dispatch:
 
 jobs:
   test:
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -1,0 +1,43 @@
+name: Notify Performance / Load Tests
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:11.8
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: perf_test_notification_api
+        ports:
+          - 5432:5432
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+
+    steps:
+      - name: Install libcurl
+        run: sudo apt-get update && sudo apt-get install libssl-dev libcurl4-openssl-dev
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+      - name: Upgrade pip
+        run: python -m pip install --upgrade pip
+      - uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+      - name: Run performance tests
+        run: /bin/bash -c "pip install -r requirements_for_test.txt && locust --headless --config tests-perf/locust/locust.conf -f tests-perf/locust/locust-notifications.py"
+      - name: Notify Slack channel if this performance test job fails
+        if: ${{ failure() && github.ref == 'refs/heads/master' }}
+        run: |
+          json="{'text':'Scheduled CI Performance testing failed: <https://github.com/cds-snc/notification-api/actions|GitHub actions>'}"
+          curl -X POST -H 'Content-type: application/json' --data "$json"  ${{ secrets.SLACK_WEBHOOK }}
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -2,7 +2,7 @@ name: Notify Performance / Load Tests
 
 on:
   schedule:
-    - cron: '0 0 * * *'
+    - cron: '0 0 * * *' # every day at midnight 
 
 jobs:
   test:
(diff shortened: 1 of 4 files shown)


tests-perf/ops/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+RUN python -m pip install wheel
+RUN python -m pip install --upgrade pip
+
+RUN set -ex && mkdir /app
+
+WORKDIR /app
+
+COPY . /app
+
+RUN set -ex && pip3 install -r requirements_for_test.txt
```
CalvinRodo, line 17: If you aren't changing the `requirements_for_test.txt` that often you can speed up build times for this Dockerfile by switching up how you copy files into this container, and taking advantage of layer caching. 

The following will only run pip3 install if the requirements file changes not if any of your python files change which it does now. 

```
COPY requirements_for_test.txt ./

RUN set -ex && pip3 install -r requirements_for_test.txt

COPY . ./
```

You can use the `./` instead of `/app` because WORKDIR changes the current directory to `/app`

https://docs.docker.com/engine/reference/builder/#workdir

Please note this is not a blocker in any way just a possible improvement

tests-perf/ops/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+ENV PYTHONDONTWRITEBYTECODE 1
+
+RUN apk add --no-cache bash build-base git libtool cmake autoconf automake gcc musl-dev postgresql-dev g++ libexecinfo-dev make libffi-dev libmagic libcurl curl-dev rust cargo && rm -rf /var/cache/apk/*
+
+# update pip
+RUN python -m pip install wheel
+RUN python -m pip install --upgrade pip
+
+RUN set -ex && mkdir /app
```
CalvinRodo, line 11: you don't need to create the `/app` directory because `WORKDIR` will create the directory for you if it doesn't exist. 

https://docs.docker.com/engine/reference/builder/#workdir

Also the `set` command is providing no value here and can be removed. 

gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

tests-perf/ops/Dockerfile
CalvinRodo, line 17: `set -ex` provides no value here and can be removed.

gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

.github/workflows/performance.yml
```diff
@@ -0,0 +1,31 @@
...
+          key: ${{ runner.os }}-pip-${{ hashFiles('requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+      - name: Run performance tests
+        run: /bin/bash -c "pip install -r requirements_for_test.txt && locust --headless --config tests-perf/locust/locust.conf -f tests-perf/locust/locust-notifications.py"
+      - name: Notify Slack channel if this performance test job fails
+        if: ${{ failure() && github.ref == 'refs/heads/master' }}
+        run: |
+          json="{'text':'Scheduled CI Performance testing failed: <https://github.com/cds-snc/notification-api/actions|GitHub actions>'}"
+          curl -X POST -H 'Content-type: application/json' --data "$json"  ${{ secrets.SLACK_WEBHOOK }}
```
CalvinRodo, line 31: Just curious why you are running the performance test in the github action here if you are also creating a lambda to run the perf test? 

if you can run it in a github action why go to the trouble of putting a lambda up in AWS. 

wouldn't it be more worthwhile to trigger the lambda here by sending an event to it?

CalvinRodo (commented):
Some non-blocking comments from me

jimleroyer (approved):
Feel free to address Calvin's comments in this PR or in a separate one. But that can be merged as-is too.

iokpala:
@CalvinRodo These are great observations..

I would prefer to merge and then address them in a later PR..

My reason is these changes would require time to ensure no adverse impact to the creation abs execution of this dockerfile

tests-perf/ops/Dockerfile
iokpala: Sure this observation is noted abs will be addressed

.github/workflows/performance.yml
iokpala: This is keeping inline with the acceptance criteria from the ticket.

Ideally we would like to keep both options as we iterate.

PR #1388, opened 2021-10-29
Upgrade to Flask/WSGI v2 + Initial draft for testing SQS and exploration

jimleroyer:
# Summary | Résumé

* Upgraded the Flask and Werkzeug/WSGI library to version 2. This required some testing locally, library compatibilities hair pulling and a very few code migration. It started as I got rid of boto2 which is outdated an decided to bite it as this is a longstanding tech debt and now a security issue.
* Got rid of boto2. We should be careful with it because I've observed that it was tried a few months ago but reverted. I will run locally before to test out if the celery workers are fine.
* Submitted script to explore SQS, especially for its rate limit.
* Reworked VSCode devcontainer files to be up to date. Previous ones stopped work for me anyway so I regenerated from some VSCode templates and finetuned.

One can run the script `test_sqs_queue_rate.py` to setup a test queue, run the test and clean up afterward.

```
> python scripts/test_sqs_queue_rate.py --help
Script to test the SQS rate limit of 10 messages / second.

Uses boto, so relies on correctly set up AWS access keys and tokens.

Usage:
    test_sqs_queue_rate.py <action>

    options are:
    - setup: Create a queue and put messages to fetch.
    - test: Test the queue by fetching the queue in batches.
    - cleanup: Delete the test queue and its messages.

Example:
        test_sqs_queue_rate.py test
```

None.

      fonctionnalité existante?

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -260,7 +260,9 @@ def init_app(app):
 
     @app.errorhandler(WerkzeugHTTPException)
     def werkzeug_exception(e):
-        return make_response(jsonify(result="error", message=e.description), e.code, e.get_headers())
+        # Make sure we will send a JSON content-type by purging error provided one.
+        headers = [h for h in e.get_headers() if h[0] != "Content-Type"]
+        return make_response(jsonify(result="error", message=e.description), e.code, headers)
 
     @app.errorhandler(404)
     def page_not_found(e):
--- a/app/authentication/auth.py
+++ b/app/authentication/auth.py
@@ -120,17 +120,17 @@ def requires_auth():
             decode_jwt_token(auth_token, api_key.secret)
         except TokenDecodeError:
             continue
-        except TokenExpiredError:
+        except TokenExpiredError as tee:
             try:
                 decoded_token = decode_token(auth_token)
             except PyJWTError:
                 continue
             current_app.logger.info(f'JWT: iat value was {decoded_token["iat"]} while server clock is {epoch_seconds()}')
             err_msg = "Error: Your system clock must be accurate to within 30 seconds"
-            raise AuthError(err_msg, 403, service_id=service.id, api_key_id=api_key.id)
+            raise AuthError(err_msg, 403, service_id=service.id, api_key_id=api_key.id) from tee
 
         _auth_with_api_key(api_key, service)
-        return
+        break
     else:
         # service has API keys, but none matching the one the user provided
         raise AuthError("Invalid token: signature, api token not found", 403, service_id=service.id)
@@ -141,8 +141,8 @@ def _auth_by_api_key(auth_token):
         # take last 36 chars of string so that it works even if the full key is provided.
         auth_token = auth_token[-36:]
         api_key = get_api_key_by_secret(auth_token)
-    except NoResultFound:
-        raise AuthError("Invalid token: API key not found", 403)
+    except NoResultFound as nrf:
+        raise AuthError("Invalid token: API key not found", 403) from nrf
     _auth_with_api_key(api_key, api_key.service)
 
 
@@ -157,10 +157,9 @@ def _auth_with_api_key(api_key, service):
     g.service_id = api_key.service_id
     _request_ctx_stack.top.authenticated_service = service
     _request_ctx_stack.top.api_user = api_key
+    user_agent = request.headers.get("User-Agent")
     current_app.logger.info(
-        "API authorised for service {} with api key {}, using client {}".format(
-            service.id, api_key.id, request.headers.get("User-Agent")
-        )
+        f"API authorised for service {service.id} with api key {api_key.id}, using client {user_agent}"
     )
 
 
--- a/app/authentication/auth.py
+++ b/app/authentication/auth.py
@@ -158,9 +158,7 @@ def _auth_with_api_key(api_key, service):
     _request_ctx_stack.top.authenticated_service = service
     _request_ctx_stack.top.api_user = api_key
     user_agent = request.headers.get("User-Agent")
-    current_app.logger.info(
-        f"API authorised for service {service.id} with api key {api_key.id}, using client {user_agent}"
-    )
+    current_app.logger.info(f"API authorised for service {service.id} with api key {api_key.id}, using client {user_agent}")
 
 
 def __get_token_issuer(auth_token):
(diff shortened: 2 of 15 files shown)


requirements-app.txt
```diff
@@ -2,7 +2,8 @@
 # with package version changes made in requirements-app.txt
 
 aws-wsgi==0.2.7
-boto==2.49.0
+boto3==1.19.4
+botocore==1.22.4
```
sastels, line 6: Can you run ` make freeze-requirements` and make sure everything is still good?

jimleroyer: Yes! that is my current goal, with some local testing with api + admin + celery.

jzbahrai (approved):
lgtm

app/__init__.py
```diff
@@ -260,7 +260,9 @@ def exception(error):
 
     @app.errorhandler(WerkzeugHTTPException)
     def werkzeug_exception(e):
-        return make_response(jsonify(result="error", message=e.description), e.code, e.get_headers())
+        # Make sure we will send a JSON content-type by purging error provided one.
+        headers = [h for h in e.get_headers() if h[0] != "Content-Type"]
```
jimleroyer, line 264: The new caught exception comes with a default header for `Content-Type` that aligns to `text`. That screws up our tests as these expect to receive JSON even if that is an error. Hence I stripped the default content type that came with the exception so that the JSON content type set later can take precedence.

app/v2/notifications/get_notifications.py
```diff
@@ -53,7 +53,7 @@ def get_pdf_for_notification(notification_id):
     except Exception:
         raise PDFNotReadyError()
 
-    return send_file(filename_or_fp=BytesIO(pdf_data), mimetype="application/pdf")
+    return send_file(path_or_file=BytesIO(pdf_data), mimetype="application/pdf")
```
jimleroyer, line 56: At the current moment, it is the only breaking change that I have noticed from upgrading Flask and Werkzeug to version 2.

requirements-app.txt
```diff
@@ -2,16 +2,18 @@
...
-celery[sqs]==5.0.5
+celery[sqs]==5.1.2
 docopt==0.6.2
 fido2==0.9.1
 Flask-Bcrypt==0.7.1
 flask-marshmallow==0.14.0
 Flask-Migrate==2.7.0
 git+https://github.com/mitsuhiko/flask-sqlalchemy.git@500e732dd1b975a56ab06a46bd1a20a21e682262#egg=Flask-SQLAlchemy==2.3.2.dev20190108
-Flask==1.1.2
+# Flask-SQLAlchemy==2.5.1 # TODO: Update to latest soon!
```
jimleroyer, line 15: I did some many library upgrades, I wanted to wait for that one because it could have an impact.

requirements.txt
```diff
@@ -4,16 +4,18 @@
 # with package version changes made in requirements-app.txt
 
 aws-wsgi==0.2.7
-boto==2.49.0
+boto3==1.17.58
```
jimleroyer, line 7: It didn't make sense to use boto 2 and boto 3 side by side, so we got rid of the deprecated version!

requirements_for_test.txt
```diff
@@ -17,7 +17,7 @@ rfc3987==1.3.8
 # used for creating manifest file locally
 jinja2-cli[yaml]==0.6.0
 black==21.5b2
-locust==1.5.3
+locust==2.4.3
```
jimleroyer, line 20: @iokpala I hope that won't create compatibility issues and have bad repercussions. I will test with a local run before the merge.

app/__init__.py
mohamed-cds: does this error handler also generate a dynamic error page ? [Removing the Content-Type](https://www.rfc-editor.org/rfc/rfc7231#section-3.1.1.5) usually means browsers will assume `application/octet-stream`, or even worse trying to guess the content type based on response.

Instead of removing, do you have the option of adding `application/json` instead of dropping?

This risk only exists if this response relates to any user interactions and can be sent as a link to logged in users.

app/authentication/auth.py
```diff
@@ -120,17 +120,17 @@ def requires_auth():
...
             except PyJWTError:
                 continue
             current_app.logger.info(f'JWT: iat value was {decoded_token["iat"]} while server clock is {epoch_seconds()}')
             err_msg = "Error: Your system clock must be accurate to within 30 seconds"
-            raise AuthError(err_msg, 403, service_id=service.id, api_key_id=api_key.id)
+            raise AuthError(err_msg, 403, service_id=service.id, api_key_id=api_key.id) from tee
 
         _auth_with_api_key(api_key, service)
-        return
+        break
```
sastels, line 133: Why change this return to a break?

app/__init__.py
jimleroyer: It's not for a dynamic error page but for non-handled exceptions raised by werkzeug. The only way to have a JSON response is to actually remove the text content type from the headers provided by the exception and then passed to the `make_response` function. The call to `jsonify` on the next line will actually set a JSON content type so the response will be JSON (and be aligned to pre 2.0 version of the werkzeug library). 

To recap, the issue was that a the `headers` parameter started to have a `content-type` header which was not present before and in doing so, was overwriting the previously set content type of JSON by `jsonify`. To remove it is to let the `jsonify` function do its work.

app/authentication/auth.py
jimleroyer: This was a Python linter warning, mentioning that `break` would be more canon to the PEP of `else` used with a conditional. I manually tested to see if that works with a `return` as well and it does, but it seems that `break` would be more semantically correct. There should be difference in the end with this change as to how I understand this.

jimleroyer: BTW, I have my linter normally set to `mypy` in VSCode so I don't see these normally but as I worked in a devcontainer and `mypy` wasn't configured, I started to see all of these Python linter warnings/errors. I will probably switch between the two linters more often to bring these small fixes for more canonical Python / best practices.

sastels:
Still sending emails and sms for me 🎉

sastels (approved):
We could have broken this into a couple PRs, but it all looks good and still works 🚀

jimleroyer:
@sastels You are absolutely right! 😅 I didn't plan to have it this way as this was initially just a try to resolve the error from the removal of boto2 dependency. Thank you! 👍

app/__init__.py
mohamed-cds: Thanks @jimleroyer that makes total sense!

PR #1389, opened 2021-11-02
Task/add bulk insertion

jzbahrai:
# Summary | Résumé

Implement the batch saving logic that will save multiple notifications into the database.

1. Added a unit test
2. Manually ran the code to test the insert and save into the DB.

- [x] Is the code maintainable? | Est-ce que le code peut être maintenu?
- [x] Have you tested it? | L’avez-vous testé?
- [x] Are there automated tests? | Y a-t-il des tests automatisés?

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -85,6 +85,30 @@ def dao_create_notification(notification):
     db.session.add(notification)
 
 
+@statsd(namespace="dao")
+@transactional
+def bulk_insert_or_update_notifications(notifications):
+    """
+    Takes a list of models.Notifications and inserts or updates the DB
+    with the list accordingly
+
+    Parameters:
+    ----------
+    notifications: lof models.Notification
+
+    Return:
+    ------
+    None
+    """
+
+    for notification in notifications:
+        if not notification.id:
+            notification.id = create_uuid()
+        if not notification.status:
+            notification.status = NOTIFICATION_CREATED
+    return db.session.bulk_save_objects(notifications)
+
+
 def _decide_permanent_temporary_failure(current_status, status):
     # Firetext will send pending, then send either succes or fail.
     # If we go from pending to delivered we need to set failure type as temporary-failure
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -87,7 +87,7 @@ def dao_create_notification(notification):
 
 @statsd(namespace="dao")
 @transactional
-def bulk_insert_or_update_notifications(notifications):
+def bulk_insert_notifications(notifications):
     """
     Takes a list of models.Notifications and inserts or updates the DB
     with the list accordingly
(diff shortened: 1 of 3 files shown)


tests/app/dao/notification_dao/test_notification_dao.py
```diff
@@ -1670,3 +1671,13 @@ def test_send_method_stats_by_service(sample_service, sample_organisation):
...
+
+
+def test_bulk_insert_notification(sample_template):
+    assert len(Notification.query.all()) == 0
+    n1 = create_notification(sample_template, client_reference="happy", save_notification=False)
+    n2 = create_notification(sample_template, client_reference="sad", save_notification=False)
+    n3 = create_notification(sample_template, client_reference="loud", save_notification=False)
+    bulk_insert_notifications([n1, n2, n3])
+    all_notifications = get_notifications_for_service(sample_template.service_id).items
+    assert len(all_notifications) == 3
```
jimleroyer, line 1683: What happens if a batch insert call is made with zero notifications? or if one faulty notification is inserted (but other two are fine)?

tests/app/db.py
```diff
@@ -239,6 +239,7 @@ def create_notification(
     reply_to_text=None,
     created_by_id=None,
     postage=None,
+    save_notification=True,
```
jimleroyer, line 242: It looks here that the need is more to get a notification rather than to create one. I think there are other functions we could reuse to get a notification without going through this one, is it possible? Might be we can sync to see the existing ones?

jimleroyer (commented):
I have a few questions around the changes.

tests/app/dao/notification_dao/test_notification_dao.py
jzbahrai: What is an example of a faulty notification?

jimleroyer: Can you check how the save work and if you can trigger an error? If some field is missing for example? Then we can test if the whole batch fail or not.

tests/app/db.py
jzbahrai: removed

tests/app/dao/notification_dao/test_notification_dao.py
jzbahrai: based on our offline convo, this will raise.  Added a todo here to update accordingly when we have redis functionality.

app/dao/notifications_dao.py
```diff
@@ -85,6 +85,31 @@ def dao_create_notification(notification):
...
+
+    Return:
+    ------
+    None
+    """
+    for notification in notifications:
+        if not notification.id:
+            notification.id = create_uuid()
+        if not notification.status:
+            notification.status = NOTIFICATION_CREATED
```
iokpala, line 107: Ideal these paths should be tested.. I actually explored this option.. What do you think??

iokpala (approved):
LGTM, barring the paths identified as side effects..

jzbahrai:
@iokpala added the tests as requested. Merging!

PR #1396, opened 2021-11-09
Task/create save notification

jzbahrai:
# Summary | Résumé

Create_notification creates an inmemory
notification object. We make two more
functions - to save the notification to
the DB, and a way to save scheduled notifications
(you can see the first commit to view the change, the second commit is just the test changes)

Regression: low risk as it is all changed only in the test folder

### DIFF
--- a/tests/app/api_key/test_rest.py
+++ b/tests/app/api_key/test_rest.py
@@ -7,6 +7,7 @@ from tests.app.db import (
     create_notification,
     create_service,
     create_template,
+    save_notification,
 )
 
 
@@ -18,7 +19,8 @@ def test_get_api_key_stats_with_sends(admin_request, notify_db, notify_db_sessio
     total_sends = 10
 
     for x in range(total_sends):
-        create_notification(template=template, api_key=api_key)
+        notification = create_notification(template=template, api_key=api_key)
+        save_notification(notification)
 
     api_key_stats = admin_request.get("api_key.get_api_key_stats", api_key_id=api_key.id)["data"]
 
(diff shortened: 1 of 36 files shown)


iokpala (approved):
LGTM

tests/app/api_key/test_rest.py
```diff
@@ -56,10 +58,10 @@ def test_get_api_keys_ranked(admin_request, notify_db, notify_db_session):
     template_email = create_template(service=service, template_type="email")
     total_sends = 10
 
-    create_notification(template=template_email, api_key=api_key_1)
+    save_notification(create_notification(template=template_email, api_key=api_key_1))
     for x in range(total_sends):
-        create_notification(template=template_email, api_key=api_key_1)
-        create_notification(template=template_email, api_key=api_key_2)
+        save_notification(create_notification(template=template_email, api_key=api_key_1))
+        save_notification(create_notification(template=template_email, api_key=api_key_2))
```
jimleroyer, line 64: There could have been just one call to one function that calls these two functions. Here we have two function calls that will repeat over on the test code base, but just one function call in all places would have been fine to keep, and probably preferable to remove some of these duplicates but also to make this refactoring faster. Different ways we could have gone:

In all case: have two separate functions for the two concerns like you have, but have one extra function that would call these two functions.

* Option A: We could have kept the `create_notification` as-is in the code base but be that third functions delegating to the other two. There would have been no need to manually change all other code paths invoking it and call one function instead of one.
* Option B: Do like option A, but rename the `create_notification` if we are not comfortable with that name. Use IDE refactoring tools to expedite the rename. Potentially have to do some manual changes but most of it would be handled automatically for us.

As-is, the PR is fine to me though. It's much better than previous state. 👍

jimleroyer (approved):
I'm good with these changes.

PR #1404, opened 2021-11-22
feat: added zendesk API abstraction and endpoint

asharahmed:
# Summary | Résumé

> 1-3 sentence description of the changed you're proposing, including a link to
> a GitHub Issue # or Trello card if applicable.

**🔔** Added Zendesk API abstraction and updated endpoint to support Zendesk requests. 

---

> Description en 1 à 3 phrases de la modification proposée, avec un lien vers le
> problème (« issue ») GitHub ou la fiche Trello, le cas échéant.

> Sequential steps (1., 2., 3., ...) that describe how to test this change. This
> will help a developer test things out without too much detective work. Also,
> include any environmental setup steps that aren't in the normal README steps
> and/or any time-based elements that this requires.

**🔴** New `env` variables, `ZENDESK_API_KEY` and `ZENDESK_API_URL` (**probably causing the tests to fail?**)

**Would appreciate assistance from reviewer to obtain passing tests**

---

> Étapes consécutives (1., 2., 3., …) qui décrivent la façon de tester la
> modification. Elles aideront les développeurs à faire des tests sans avoir à
> jouer au détective. Veuillez aussi inclure toutes les étapes de configuration
> de l’environnement qui ne font pas partie des étapes normales dans le fichier
> README et tout élément temporel requis.

None.

---

> mentionné.

      fonctionnalité existante?


app/clients/zendesk.py
```diff
@@ -0,0 +1,96 @@
...
+import requests
+from flask import current_app
+from requests.auth import HTTPBasicAuth
+
+from app.user.contact_request import ContactRequest
+
+__all__ = ["Zendesk"]
+
+
+class Zendesk(object):
```
jimleroyer, line 14: Hmm why not reuse the [`ZenDeskSell` class](https://github.com/cds-snc/notification-api/blob/bebd806cad9581101e5cc1d52db2613c1baa1fa9/app/clients/zendesk_sell.py#L16) we already have defined? Is this a different API class than the Zendesk Sell API?

If that is the case, I wonder if we can have a base class for both types of API for Zendesk that would share common configuration and functions?

app/clients/zendesk.py
```diff
@@ -0,0 +1,96 @@
...
+                json=self._generate_ticket(),
+                auth=HTTPBasicAuth(current_app.config["ZENDESK_API_KEY"], "x"),
+                timeout=5,
+            )
+            response.raise_for_status()
+
+            return response.status_code
+        except requests.RequestException as e:
+            content = json.loads(response.content)
+            current_app.logger.warning(f"Failed to create Zendesk ticket: {content['errors']}")
```
jimleroyer, line 90: We'd probably want this to be an error rather than a warning. I realize now that the logging in the `ZenDeskSell` sister class is all warnings too.. that might explain why our recent key auth errors went under the radar but I'd need to confirm the alarms if we have in place too. cc @wmoussa-cds

app/clients/zendesk.py
```diff
@@ -0,0 +1,96 @@
...
+            raise e
+        except NotImplementedError:
+            # There are cases in development when we do not want to send to freshdesk
+            # because configuration is not defined, lets return a 200 OK
+            current_app.logger.warning("Did not send ticket to Zendesk")
+            return 200
```
jimleroyer, line 96: I would get rid of this block. We probably want to know when things are misconfigured. In development, we can use a mock class or special implementation if we need.. the higher layers should handle the `NotImplementedError` error (and probably log it as error/exception).

I see that you followed what `freshdesk.py` did but `zendesk_sell.py` would have better practices IMHO.

app/clients/zendesk.py
```diff
@@ -0,0 +1,96 @@
...
+                auth=HTTPBasicAuth(current_app.config["ZENDESK_API_KEY"], "x"),
+                timeout=5,
+            )
+            response.raise_for_status()
+
+            return response.status_code
+        except requests.RequestException as e:
+            content = json.loads(response.content)
+            current_app.logger.warning(f"Failed to create Zendesk ticket: {content['errors']}")
+            raise e
```
jimleroyer, line 91: If we raise here then we logging it previously is not necessary. I would replicate what the `ZenDeskSell` class does and delegate responsibility of exception handling to the caller entirely, else we will end up with duplicate log entries.

I see that you followed what `freshdesk.py` did but `zendesk_sell.py` would have better practices IMHO.

app/clients/zendesk.py
```diff
@@ -0,0 +1,96 @@
...
+            api_url = current_app.config["ZENDESK_API_URL"]  # Zendesk API URL
+            if not api_url:
+                raise NotImplementedError
+
+            # The API and field definitions are defined here:
+            # https://developer.zendesk.com/rest_api/docs/support/tickets
+            response = requests.post(
+                urljoin(api_url, "/api/v2/tickets"),
+                json=self._generate_ticket(),
+                auth=HTTPBasicAuth(current_app.config["ZENDESK_API_KEY"], "x"),
```
jimleroyer, line 82: Replicating self.token would be nice, like in `ZenDeskSell` class.

app/clients/zendesk.py
```diff
@@ -0,0 +1,96 @@
...
+            "comment": self._generate_description(),
+            "email": self.contact.email_address,
+            "priority": 1,
+            "status": 2,
+            "tags": self.contact.tags,
+        }
+
+    def send_ticket(self) -> int:
+        try:
+            api_url = current_app.config["ZENDESK_API_URL"]  # Zendesk API URL
```
jimleroyer, line 73: This could be a member of the class such as self.api_url like what the `ZenDeskClass` does.

app/user/rest.py
```diff
@@ -492,6 +494,7 @@ def send_branding_request(user_id):
         return jsonify({}), 400
 
     status_code = Freshdesk(contact).send_ticket()
+    status_code = Zendesk(contact).send_ticket()
```
jimleroyer, line 497: 🙌

jimleroyer (commented):
Good start overall! Keep it going!

app/clients/zendesk.py
asharahmed: Yes, I believe `Zendesk` varies from `ZenDeskSell`, I think we could have a parent class for both types of Zendesk APIs

asharahmed:
Currently, configuring `_generate_ticket()` to conform with the [new Zendesk API](https://developer.zendesk.com/api-reference/ticketing/tickets/tickets/#create-ticket). Appears to be why the one test is failing

app/clients/zendesk.py
mohamed-cds: reviewed this and zendesk & zendesk sell have nothing in common. zendesk sell is an acquisition and is still using the [old companies api's](https://developers.getbase.com/docs/rest/articles/first_call). Considering we're only using this class for one call, abstracting and trying to identify a common element isn't practical. Recommending we leave as-is.

app/clients/zendesk.py
mohamed-cds: changed so that error handling is done in one place (caller)

app/clients/zendesk.py
mohamed-cds: changed so that the error handling is done in one place (caller)

mohamed-cds:
i fail at rebasing, so don't mind the million commits now in this pr 😞

app/clients/zendesk.py
jimleroyer: My thinking for reusability was not so much on the business domain level but more on the technical one. There are some shared properties between the two classes such as the base URL and the api key. The `zendesk_sell` module also has a `_send_request` function that has better practices for error handling IMHO. For example, the original `freshdesk` module catch the exception in place, log it and re-throw, which would cause a logging of the same error twice.

app/clients/zendesk.py
```diff
@@ -0,0 +1,89 @@
...
+        response = requests.post(
+            urljoin(self.api_url, "/api/v2/tickets"),
+            json=self._generate_ticket(),
+            auth=HTTPBasicAuth(f"{self.contact.email_address}/token", self.token),
+            timeout=5,
+        )
+        response.raise_for_status()
+        if response.status_code != 201:
+            current_app.logger.error(f"Zendesk api error; response code: {response.status_code}")
+            raise requests.HTTPError
```
jimleroyer, line 89: Can we raise an error that will pack in the status code and let the caller do the error handling? It would drive to better separation of concerns that way, especially as the caller will catch the exception anyway and log it (but currently the error information is split in two different errors when there is no restriction for that to happen).

app/user/rest.py
```diff
@@ -461,6 +462,12 @@ def send_contact_request(user_id):
 
     if contact.is_demo_request():
         return jsonify({}), 204
+
+    try:
+        Zendesk(contact).send_ticket()
+    except Exception as e:
+        current_app.logger.exception(e)
```
jimleroyer, line 469: We could log the response code here as well with the logged exception.

app/clients/zendesk.py
```diff
@@ -0,0 +1,89 @@
...
+
+        # The API and field definitions are defined here:
+        # https://developer.zendesk.com/rest_api/docs/support/tickets
+        response = requests.post(
+            urljoin(self.api_url, "/api/v2/tickets"),
+            json=self._generate_ticket(),
+            auth=HTTPBasicAuth(f"{self.contact.email_address}/token", self.token),
+            timeout=5,
+        )
+        response.raise_for_status()
```
jimleroyer, line 86: The error handling code following this line will not occur if we call `raise_for_status`, i.e. they would do the same logic approximately (without the  logging), [per documentation](https://docs.python-requests.org/en/latest/api/#requests.Response.raise_for_status).

app/user/rest.py
```diff
@@ -491,6 +498,10 @@ def send_branding_request(user_id):
         current_app.logger.error(e)
         return jsonify({}), 400
 
+    status_code = Zendesk(contact).send_ticket()
```
jimleroyer, line 501: We'd need to catch the exception here like in lines [468-469](https://github.com/cds-snc/notification-api/pull/1404/files#diff-856770ce1eee6ec9509abce6fde06d0f795d88514f04c6ab6518697a9e3bcc08R468).

app/user/rest.py
```diff
@@ -491,6 +498,10 @@ def send_branding_request(user_id):
...
+    if status_code != 201:
+        current_app.logger.error(f"Zendesk error response code: {status_code}")
```
jimleroyer, line 503: The `send_ticket` function does not return status code and that would not be the proper way to catch the error. We'd rather need to catch the exception and log along with the status code.

jimleroyer (changes requested):
Thank you for the contribution again.

I made a few comments, some code that might not work as-is.

I would also expect unit tests against the `zendesk` module similar to `test_freshdesk`.

app/user/rest.py
mohamed-cds: Nice find, i totally missed this when i refactored error handling. Will add the try/catch

jimleroyer: That's a lot of changes we went through so quite understandable. Our many 👀 are here to help. 👍

app/clients/zendesk.py
mohamed-cds: The main reason i recommended not creating a common class and mirroring `zendesk_sell` is that we only have one POST request at this time. I think it comes down to a preference on when we want to optimize and i believe it's premature for zendesk since the only shared functionality is the base url & api key which all API's have, in addition to having different authentication mechanisms.

If we add additional functionality then i believe refactoring `Zendesk` to use a single `_send_request` makes sense since you don't have to worry about how to authenticate, etc. I've also changed how the error handling is done and what i have is very similar to `zendesk_sell` with the main difference being that the request is done in a single function + throwing a custom exception

app/clients/zendesk.py
mohamed-cds: My thinking was to catch scenario's where the http status code is not between 400 & 600 but thinking this over makes me think this is will never happen so it's pointless to have both. I'll remove the raise_for_status and add the response code to the error logging

app/clients/zendesk.py
mohamed-cds: 💯 Made the change

app/user/rest.py
mohamed-cds: Zendesk error handling changed to send back the details via exception, so we should be good now

app/clients/zendesk.py
jimleroyer: The main reuse I would have liked is around a common `_send_request` where the error handling could have been the same, so there was more than base url and api key for reusability. I am okay with some duplicate logic though and given we'll decommission Freshdesk soon.

jimleroyer (approved):
LGTM! 👏

wmoussa-cds:
🎉

PR #1408, opened 2021-11-26
Removed env vars from perf-test dockerfile.

jimleroyer:
# Summary | Résumé

* Removed settings of environment variables within the perf-test Dockerfile setup. These should be unnecessary now as we set these env vars through Terraform passing it to the ECS task definition.
* Removed default Notify resources' IDs from Locust code. These can be configured separately on a locust configuration file for the user and we will minimize exposing test IDs not normally accessible to the public (i.e. it's not bad but not ideal and we can configure this in some other way).

* Run the docker image by passing by the env vars through the command line parameters, as this is what the ECS task will do.

This should not break anything as the perf test is already broken as we revoked past referred API key in the header.

      fonctionnalité existante?

### DIFF
--- a/tests-perf/ops/Dockerfile
+++ b/tests-perf/ops/Dockerfile
@@ -19,6 +19,4 @@ RUN echo "fs.file-max = 100000" >> /etc/sysctl.conf
 
 ARG GIT_SHA
 
-ENV GIT_SHA=${GIT_SHA}
-
 ENTRYPOINT [ "bin/execute_and_publish_performance_test.sh" ]
--- a/tests-perf/ops/Dockerfile
+++ b/tests-perf/ops/Dockerfile
@@ -17,6 +17,4 @@ COPY . /app
 RUN set -ex && pip3 install -r requirements_for_test.txt
 RUN echo "fs.file-max = 100000" >> /etc/sysctl.conf
 
-ARG GIT_SHA
-
 ENTRYPOINT [ "bin/execute_and_publish_performance_test.sh" ]
--- a/tests-perf/ops/Dockerfile
+++ b/tests-perf/ops/Dockerfile
@@ -19,17 +19,6 @@ RUN echo "fs.file-max = 100000" >> /etc/sysctl.conf
 
 ARG GIT_SHA
 
-ENV GIT_SHA=${GIT_SHA} \ 
-    PERF_TEST_PHONE_NUMBER=16132532222 \ 
-    PERF_TEST_EMAIL=success@simulator.amazonses.com \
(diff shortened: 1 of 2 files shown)


tests-perf/ops/Dockerfile
```diff
@@ -19,17 +19,6 @@ RUN echo "fs.file-max = 100000" >> /etc/sysctl.conf
...
-    PERF_TEST_EMAIL=success@simulator.amazonses.com \
-    PERF_TEST_AWS_S3_BUCKET=notify-perfortmance-test-results-staging \ 
-    PERF_TEST_CSV_DIRECTORY_PATH=/tmp/notify_performance_test \
-    PERF_TEST_DOMAIN=https://api.staging.notification.cdssandbox.xyz \
-    PERF_TEST_SMS_TEMPLATE_ID=83d01f06-a818-4134-bd69-ce90a2949280 \
-    PERF_TEST_BULK_EMAIL_TEMPLATE_ID=5ebee3b7-63c0-4052-a8cb-387b818df627 \
-    PERF_TEST_EMAIL_TEMPLATE_ID=a59b313d-8de2-4973-ac2f-66de7ec0b239 \
-    PERF_TEST_EMAIL_WITH_ATTACHMENT_TEMPLATE_ID=a59b313d-8de2-4973-ac2f-66de7ec0b239 \
-    PERF_TEST_EMAIL_WITH_LINK_TEMPLATE_ID=5ebee3b7-63c0-4052-a8cb-387b818df627 \
-    TEST_AUTH_HEADER="ApiKey-v1 c55039fc-c0e1-44db-a14e-b6a669148ec6"
```
sastels, line 33: Hi guys to remember how docker works… do we need an ARG for each of theses and then something like the GIT_SHA=${GIT_SHA}

jimleroyer: Environment variable would be defined [via the `--env` parameter on the command line](https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file).

Regarding the GIT_SHA, I do wonder if we need to keep that one in there. It looks like setting up an env var using an env var. I can remove that too.

sastels: I think it’s moving the vars from what’s passed to docker to what’s in the image? Though… in our case we’re gonna inject them with AWS, not when the image is built. So yeah I think what’s here is right.

sastels (approved):
🚀

tests-perf/ops/Dockerfile
jimleroyer: Yes you are right about the `arg`, I didn't notice. But removing both arg and env setup. 👍

PR #1413, opened 2021-12-03
Add heartbeat to keep lambda alive

jzbahrai:
# Summary | Résumé

The heartbeat file is used to keep the lambda function warm. I have added the send_email function and can add more as needed.

The Dockerfile has also been tested!


heartbeat/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN python -m pip install --upgrade pip
+
+RUN set -ex && mkdir /app
+
+WORKDIR /app
+
+COPY . /app
+
+RUN set -ex && pip3 install -r requirements_for_test.txt
+RUN echo "fs.file-max = 100000" >> /etc/sysctl.conf
```
sastels, line 18: I think this was just for the perf tests where we need tons of open connections, we should be fine without it for the heartbeat.

heartbeat/heartbeat.py
```diff
@@ -0,0 +1,16 @@
...
+Code to keep the lambda function alive.
+"""
+import os
+
+from notifications_python_client.notifications import NotificationsAPIClient
+
+API_KEY = os.getenv("HEARTBEAT_API_KEY")
+BASE_URL = "https://api.notification.canada.ca"
+EMAIL_ADDRESS = "success@simulator.amazonses.com"
+TEMPLATE_ID = "7daa4f22-633b-4a7b-8b20-dfc5f66882bd"
```
sastels, line 11: We should get the `BASE_URL` and `TEMPLATE_ID` from environment as well, so we can use the same code for staging and prod

sastels (changes requested):
Good idea to use the notify client. Just a couple suggestions.

heartbeat/heartbeat.py
sastels: hrm. Should we have two heartbeats for the canary test? explicitly send to both the k8s and lambda apis. probably, especially if there's warm up issues for the lambda. Maye we want `BASE_URL` to be an array that we loop over.

sastels (changes requested):

heartbeat/heartbeat.py
```diff
@@ -0,0 +1,16 @@
...
+
+API_KEY = os.getenv("HEARTBEAT_API_KEY")
+BASE_URL = os.getenv("BASE_URL")
+EMAIL_ADDRESS = "success@simulator.amazonses.com"
+TEMPLATE_ID = os.getenv("TEMPLATE_ID")
+
+notifications_client = NotificationsAPIClient(API_KEY, base_url=BASE_URL)
+
+if __name__ == "__main__":
+    response = notifications_client.send_email_notification(email_address=EMAIL_ADDRESS, template_id=TEMPLATE_ID)
```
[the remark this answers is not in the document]
jzbahrai: changed!

heartbeat/heartbeat.py
```diff
@@ -0,0 +1,18 @@
+"""
+Code to keep the lambda function alive.
+"""
+import os
+
+from notifications_python_client.notifications import NotificationsAPIClient
+
+API_KEY = os.getenv("TF_VAR_heartbeat_api_key")
+# BASE_URL is a list of api urls - k8s and lambda.
+BASE_URL = os.getenv("TF_VAR_heartbeat_base_url")
```
jimleroyer, line 10: If this is an array, it would be nice to have a type declaration, i.e.
```suggestion
BASE_URL: [] = os.getenv("TF_VAR_heartbeat_base_url")
```

It would be nice to have it on all other constants too to be explicit defined too please. 😅 🙏

jimleroyer (changes requested):
Nice work! mostly minor issues except the Dockerfile setup. It would be good to test locally the Dockerfile locally to see if it actually work, I am not sure it will.

heartbeat/heartbeat.py
```diff
@@ -2,15 +2,14 @@
...
-import uuid
+from typing import List
 
 from notifications_python_client.notifications import NotificationsAPIClient
 
-API_KEY: str = os.getenv("TF_VAR_heartbeat_api_key")
-# BASE_URL is a list of api urls - k8s and lambda.
-BASE_URL: [] = os.getenv("TF_VAR_heartbeat_base_url")
+API_KEY: str = os.getenv("TF_VAR_heartbeat_api_key", "")
+BASE_URL: List[str] = os.getenv("TF_VAR_heartbeat_base_url") # type: ignore
```
jimleroyer, line 10: Make more sense this way. 👍

jimleroyer (approved):
:shipit: 
Code style fails but it looks good! Better test locally if you haven't yet as this might take more time if there is a docker issue later.

PR #1416, opened 2021-12-10
Add more heartbeat log

jzbahrai:
# Summary | Résumé

Added better logs, error handing and an exit code.

Unfortunately, cannot check for response code as this is what the python client respond back:
```
{'content': {'body': 'This is the heartbeat test message.\r\n\r\nCeci est le test de battement de coeur pour Notify.', 'from_email': 'gc.notify.gc.notification@notification.canada.ca', 'subject': 'Heartbeat-test'}, 'id': '8e17b953-4b00-4b26-974c-3f153032f9aa', 'reference': None, 'scheduled_for': None, 'template': {'id': '7daa4f22-633b-4a7b-8b20-dfc5f66882bd', 'uri': 'https://api-lambda.notification.canada.ca/services/d6aa2c68-a2d9-4437-ab19-3ae8eb202553/templates/7daa4f22-633b-4a7b-8b20-dfc5f66882bd', 'version': 2}, 'uri': 'https://api-lambda.notification.canada.ca/v2/notifications/8e17b953-4b00-4b26-974c-3f153032f9aa'}```

### DIFF
--- a/heartbeat/heartbeat.py
+++ b/heartbeat/heartbeat.py
@@ -28,7 +28,7 @@ if __name__ == "__main__":
             response = notifications_client.send_email_notification(email_address=EMAIL_ADDRESS, template_id=TEMPLATE_ID)
             print("Email has been sent by {}!".format(base_url))
         except HTTPError as e:
-            print(f"Could not send heartbeat: status={e.status_code}, msg={e._message}")
+            print(f"Could not send heartbeat: status={e.status_code}, msg={e.message}")
             raise
         except Exception:
             raise
--- a/heartbeat/heartbeat.py
+++ b/heartbeat/heartbeat.py
@@ -6,6 +6,7 @@ import os
 import uuid
 from typing import List
 
+from notifications_python_client.errors import HTTPError
 from notifications_python_client.notifications import NotificationsAPIClient
 
 API_KEY: str = os.getenv("heartbeat_api_key", "")
@@ -26,7 +27,12 @@ if __name__ == "__main__":
         try:
             response = notifications_client.send_email_notification(email_address=EMAIL_ADDRESS, template_id=TEMPLATE_ID)
             print("Email has been sent by {}!".format(base_url))
+        except HTTPError as e:
+            print("There was an error")
+            print(e.status_code)
+            print(e._message)
+            raise
         except Exception as e:
-            raise e
+            raise
 
     exit(0)
--- a/heartbeat/heartbeat.py
+++ b/heartbeat/heartbeat.py
@@ -28,9 +28,7 @@ if __name__ == "__main__":
             response = notifications_client.send_email_notification(email_address=EMAIL_ADDRESS, template_id=TEMPLATE_ID)
             print("Email has been sent by {}!".format(base_url))
         except HTTPError as e:
-            print("There was an error")
-            print(e.status_code)
-            print(e.message)
+            print(f"Could not send heartbeat: status={e.status_code}, msg={e._message}")
             raise
         except Exception:
             raise
--- a/heartbeat/heartbeat.py
+++ b/heartbeat/heartbeat.py
@@ -23,5 +23,10 @@ if __name__ == "__main__":
         print("Variable TEMPLATE_ID is missing")
     for base_url in BASE_URL:
         notifications_client = NotificationsAPIClient(API_KEY, base_url=base_url)
-        response = notifications_client.send_email_notification(email_address=EMAIL_ADDRESS, template_id=TEMPLATE_ID)
-        print("Email has been sent!")
+        try:
+            response = notifications_client.send_email_notification(email_address=EMAIL_ADDRESS, template_id=TEMPLATE_ID)
+            print("Email has been sent by {}!".format(base_url))
+        except Exception as e:
+            raise e
+
+    exit(0)
--- a/heartbeat/heartbeat.py
+++ b/heartbeat/heartbeat.py
@@ -30,7 +30,5 @@ if __name__ == "__main__":
         except HTTPError as e:
             print(f"Could not send heartbeat: status={e.status_code}, msg={e.message}")
             raise
-        except Exception:
-            raise
 
     exit(0)
--- a/heartbeat/heartbeat.py
+++ b/heartbeat/heartbeat.py
@@ -30,9 +30,9 @@ if __name__ == "__main__":
         except HTTPError as e:
             print("There was an error")
             print(e.status_code)
-            print(e._message)
+            print(e.message)
             raise
-        except Exception as e:
+        except Exception:
             raise
 
     exit(0)

### CODE REVIEW

File: heartbeat/heartbeat.py [hunk 1]
```diff
@@ -23,5 +23,10 @@
         print("Variable TEMPLATE_ID is missing")
     for base_url in BASE_URL:
         notifications_client = NotificationsAPIClient(API_KEY, base_url=base_url)
-        response = notifications_client.send_email_notification(email_address=EMAIL_ADDRESS, template_id=TEMPLATE_ID)
-        print("Email has been sent!")
+        try:
+            response = notifications_client.send_email_notification(email_address=EMAIL_ADDRESS, template_id=TEMPLATE_ID)
+            print("Email has been sent by {}!".format(base_url))
+        except Exception as e:
+            raise e
```
jimleroyer, line 30: Hmm it would not be necessary to catch the exception if we simply raise it next, right?

jzbahrai: true, I just prefer this stylistically, but I can change it to just `raise` if you prefer

jimleroyer:
> Unfortunately, cannot check for response code as this is what the python client respond back

If no exception is raised, things went OK but if there is an error, an `HTTPError` exception [would be raised along with the error code](https://docs.notifications.service.gov.uk/python.html#error-codes).

heartbeat/heartbeat.py
jimleroyer: We could catch a `HTTPError`, print some error message and exit the process with a non `0` exit code. [Per documentation, we could have the response code when an error occurs](https://docs.notifications.service.gov.uk/python.html#error-codes).

If the exception is not a `HTTPError`, we can let it bubble up the stack and crash the program probably.

sastels, line 30: If we get a 400 or something then it won't raise an exception - we should check out the status_code we get back. https://docs.notifications.service.gov.uk/python.html#error-codes
(question: what happens when we have a bad url, like we do now with api-k8s on prod?)

jimleroyer: A `400` would raise an exception but not `200` you mean, no?

jimleroyer: > I can change it to just raise if you prefer

You don't even need to raise, the exception will occur and bubble up.

sastels:
What's the reply when trying to connect to api-k8s.notification.canada.ca?

jzbahrai:
@sastels 
```
docker run -e heartbeat_template_id=xxx -e heartbeat_api_key="XXX" -e heartbeat_base_url='["https://api-lambda.notification.canada.ca", "https://api.notification.canada.ca"]' notify/heartbeat
Email has been sent by https://api-lambda.notification.canada.ca!
Email has been sent by https://api.notification.canada.ca!

File: heartbeat/heartbeat.py [hunk 2]
```diff
@@ -23,5 +24,15 @@
...
         notifications_client = NotificationsAPIClient(API_KEY, base_url=base_url)
-        response = notifications_client.send_email_notification(email_address=EMAIL_ADDRESS, template_id=TEMPLATE_ID)
-        print("Email has been sent!")
+        try:
+            response = notifications_client.send_email_notification(email_address=EMAIL_ADDRESS, template_id=TEMPLATE_ID)
+            print("Email has been sent by {}!".format(base_url))
+        except HTTPError as e:
+            print("There was an error")
+            print(e.status_code)
+            print(e._message)
```
jimleroyer, line 33: Can we provide a bit more clarity on the error msg?
```
print(f"Could not send heartbeat: status={e.status_code}, msg={e._message}")
```

heartbeat/heartbeat.py
```diff
@@ -23,5 +24,15 @@
...
+            raise
+        except Exception as e:
+            raise
```
jimleroyer, line 36: No need to except and raise: there is no error handling. Omitting this will have the same effect.

heartbeat/heartbeat.py
```diff
@@ -23,5 +24,15 @@
         print("Variable TEMPLATE_ID is missing")
     for base_url in BASE_URL:
         notifications_client = NotificationsAPIClient(API_KEY, base_url=base_url)
-        response = notifications_client.send_email_notification(email_address=EMAIL_ADDRESS, template_id=TEMPLATE_ID)
-        print("Email has been sent!")
+        try:
+            response = notifications_client.send_email_notification(email_address=EMAIL_ADDRESS, template_id=TEMPLATE_ID)
+            print("Email has been sent by {}!".format(base_url))
+        except HTTPError as e:
```
jimleroyer, line 30: Perfect! (although `e` isn't used and linter complains)

heartbeat/heartbeat.py
jzbahrai: done

heartbeat/heartbeat.py
jzbahrai: I guess thats fair, it will just bubble up the stack. I just think this is cleaner and would like to leave it as such.

heartbeat/heartbeat.py
jzbahrai: done

heartbeat/heartbeat.py
jimleroyer: Hmm no.. I will challenge you on that because it is not cleaner. But I would like to know why you think it is? Also, can you refer to me documented best practices, or usage of this style in the Python code base (or Notify)? Stackoverflow articles maybe? 

This is to me against many best practices I know, i.e. 1. don't write error handling code if we don't handle error and 2. write less code.

jzbahrai: That's fair I shouldn't have resolved, my bad.

This is my logic - when looking at a try, except -> you think where is this exception being raised? Logically you look at the `except` in the code but it won't be line 30 if it isn't HTTPError.  So then you go back to line 28 which is raising the exception, though by convention you don't want to think of the `try` raising the exception.

that being said, best practice is to never `except Exception` but catch specific `exceptions`.
Is this redundant code? Yes
Does it make it cleaner and more obvious though? Imo Yes

Regardless, I am not tied to one way to another and am open to changing it even if I prefer the other way.

jimleroyer: I think I get what you are saying (I think): that logic around raising exception is implicit and here you make it explicit. Hence it is more clear. Is that how you see it?

jzbahrai: yup that was a pretty concise summary of what I was trying to say.

jimleroyer: **TL;DR**: I have the impression here that what we might desire is stronger type safety for more explicit control flow.

Okay I see. In a way, I agree and I am in the camp that these things should be explicit (and hence I am a fan of languages with stronger type safety where types and control flow is explicit), but alas, Python is not such a programming language.

I think it's **not** a common practice in the Python community because the non explicit control flow is accepted (or even static typing until recent effort). In other communities like Haskell though, maximum expression of the control flow is desired, but that comes at a price of hitting walls against the compiler and extra types. Python takes a more loose approach and focus on getting things done (sorry Haskell, I still love you).

Consider runtime exceptions: these are dynamic. When a runtime exception occurs, they bubble up the stack and are not necessarily all advertised through the functions/methods signatures or even through all the encountered `try...except` clauses. This is a common criticism of using runtime/dynamic exceptions in these programming languages: they can occur everywhere and control flow is quite dynamic.

An empty error management block is not a common practice in such languages though (or any that I know of): imagine that every time we write a `try`, we will now add an `except Exception... raise` block with no error handling code, on the purpose of _clarity_ (or explicit declaration that something might happen but... 🤞 ). It's a losing game because these are all dynamic runtime exceptions and they can spawn easily everywhere.

Keeping an open mind: let's say we adopt this practice, we decide to go that way of defining explicitly `try... raise Exception`, should we add an empty error management block every time we do a `try...` block in Notify? Should we add this to existing blocks in the code base? all of our repos? Would it be really clearer and more explicit? I'd say it's a known behavior and normal for Python. We do a `try` if we think there is a code point that can handle an error usually, for fine (i.e. `HTTPError`) or large grained errors (i.e. `Exception` and its various problematic children classes). We don't do it **solely** to declare that something might happen.

Being more explicit in Python with dynamic runtime exceptions would mean more code cluttering of empty `except...raise` blocks. I can change my mind on that, but I'd like to see real world examples of this practice and documentation/discussion about it.

Sorry for the long message, this was not a concise summary that time.

sastels (approved):

PR #1435, opened 2022-01-19
Cast the env variable to int for chunk size

jzbahrai:
# Summary | Résumé

The env variable is currently a string, we need to change it to a int.


sastels (approved):
👍

iokpala (approved):
LGTM

PR #1444, opened 2022-01-26
Redis heartbeat worker

sastels:
# Summary | Résumé

code for the heartbeat's worker that takes inflight messages and save / adds to queue

None.

      fonctionnalité existante?

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -81,6 +81,47 @@ from app.service.utils import service_allowed_to_send_to
 from app.utils import get_csv_max_rows
 
 
+@notify_celery.task(name="process-inflight")
+@statsd(namespace="tasks")
+def process_inflight(inflight_name):
+
+    # TODO: get inflight notifications from redis somehow
+    notifications = []
+
+    encrypted_smss: List[Any] = []
+    encrypted_emails: List[Any] = []
+    for notification in notifications:
+        service = dao_fetch_service_by_id(notification["service_id"])
+        template = dao_get_template_by_id(notification["template_id"])
+
+        if service_allowed_to_send_to(notification.recipient, service, KEY_TYPE_NORMAL):
+            encrypted_notification = encryption.encrypt(
+                {
+                    "api_key": notification["api_key_id"],
+                    "template": notification["template_id"],
+                    "to": notification["recipient"],
+                    "personalisation": dict(notification["personalisation"]),
+                }
+            )
+            if template.template_type == SMS_TYPE:
+                encrypted_smss.append(encrypted_notification)
+            else:
+                encrypted_emails.append(encrypted_notification)
+
+    if encrypted_smss:
+        save_smss.apply_async(
+            (str(notification.service_id), encrypted_smss),
+            queue=QueueNames.DATABASE if not service.research_mode else QueueNames.RESEARCH_MODE,
+        )
+    if encrypted_emails:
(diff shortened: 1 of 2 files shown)


app/celery/tasks.py
```diff
@@ -249,9 +258,18 @@ def __sending_limits_for_job_exceeded(service, job: Job, job_id):
...
+        pass
+
+    def acknowledge(self, receipt):
+        pass
+
+
 @notify_celery.task(bind=True, name="save-smss", max_retries=5, default_retry_delay=300)
 @statsd(namespace="tasks")
-def save_smss(self, service_id: str, encrypted_notifications: List[Any]):
+def save_smss(self, service_id: str, encrypted_notifications: List[Any], receipt: str):
```
jzbahrai, line 272: can we update the doc string for receipt and how it is connected to the post vs bulk?

app/celery/tasks.py
```diff
@@ -400,7 +424,7 @@ def save_sms(self, service_id, notification_id, encrypted_notification, sender_i
 
 @notify_celery.task(bind=True, name="save-emails", max_retries=5, default_retry_delay=300)
 @statsd(namespace="tasks")
-def save_emails(self, service_id: str, encrypted_notifications: List[Any]):
+def save_emails(self, service_id: str, encrypted_notifications: List[Any], receipt: str):
```
jzbahrai, line 427: same coment as above

jzbahrai (commented):
this is awesome! thank you steve

app/celery/tasks.py
```diff
@@ -81,6 +81,15 @@
 from app.utils import get_csv_max_rows
 
 
+@notify_celery.task(name="process-inflight")
+@statsd(namespace="tasks")
+def process_inflight(receipt, results, type):
+    if type == SMS_TYPE:
+        save_smss.apply_async((None, results, receipt), queue=QueueNames.DATABASE)
```
jimleroyer, line 88: In order to not overflow the `DATABASE` queue even further with more messages, we might want to use a different queue such as `QueueNames.NOTIFY_CACHE` instead. We already overflow the `DATABASE` queue at the moment as it receives all notification to save down the pipeline.

app/celery/tasks.py
```diff
@@ -81,6 +81,15 @@
...
+    else:
+        save_emails.apply_async((None, results, receipt), queue=QueueNames.DATABASE)
```
jimleroyer, line 90: Same would apply here of course.

app/celery/tasks.py
jimleroyer, line 272: It seems the receipt param could be `None`, right? Could we add a default value to it?

app/celery/tasks.py
jimleroyer, line 427: It seems the receipt param could be `None`, right? Could we add a default value to it?

jimleroyer (commented):
Left a few comments, the one I really care though is the queue we'll use to process inflight with the  celery worker.

app/celery/tasks.py
sastels: yeah, good point, the typing here should be `Optional[str]`

app/celery/tasks.py
sastels: 👍  `Optional[str]` as above

jzbahrai (approved):

PR #1451, opened 2022-01-31
Add suffix to inflight list

jzbahrai:
Have the suffix added to the name of the inflight list.

### DIFF
--- a/app/queue.py
+++ b/app/queue.py
@@ -23,7 +23,7 @@ class Buffer(Enum):
         return f"{self.value}:{suffix}" if suffix else self.value
 
     def get_inflight_name(self, receipt: UUID = uuid4(), suffix: str = "") -> str:
-        return f"{Buffer.IN_FLIGHT.value}:{str(suffix)}{str(receipt)}"
+        return f"{Buffer.IN_FLIGHT.value}:{str(suffix)}:{str(receipt)}"
 
 
 class Queue(ABC):
--- a/app/queue.py
+++ b/app/queue.py
@@ -22,6 +22,9 @@ class Buffer(Enum):
     def name(self, suffix=None):
         return f"{self.value}:{suffix}" if suffix else self.value
 
+    def get_inflight_name(self, receipt: UUID = uuid4(), suffix: str = None) -> str:
+        return f"{Buffer.IN_FLIGHT.value}:{str(suffix)}{str(receipt)}"
+
 
 class Queue(ABC):
     """Queue interface for custom buffer.
@@ -83,8 +86,9 @@ class RedisQueue(Queue):
 
     scripts: Dict[str, Any] = {}
 
-    def __init__(self, inbox_suffix=None) -> None:
-        self._inbox = Buffer.INBOX.name(inbox_suffix)
+    def __init__(self, suffix=None) -> None:
+        self._inbox = Buffer.INBOX.name(suffix)
+        self._suffix = suffix
 
     def init_app(self, redis):
         self._redis_client = redis
@@ -92,16 +96,13 @@ class RedisQueue(Queue):
 
     def poll(self, count=10) -> tuple[UUID, list[str]]:
         receipt = uuid4()
-        in_flight_key = self.get_inflight_name(receipt)
+        in_flight_key = Buffer.IN_FLIGHT.get_inflight_name(receipt, self._suffix)
         results = self.__move_to_inflight(in_flight_key, count)
         return (receipt, results)
 
     def acknowledge(self, receipt: UUID):
-        inflight_name = self.get_inflight_name(receipt)
+        inflight_name = Buffer.IN_FLIGHT.get_inflight_name(receipt, self._suffix)
         self._redis_client.delete(inflight_name)
-
-    def get_inflight_name(self, receipt: UUID = uuid4()) -> str:
-        return f"{Buffer.IN_FLIGHT.value}:{str(receipt)}"
 
     def publish(self, message: str):
         self._redis_client.rpush(self._inbox, message)
--- a/app/queue.py
+++ b/app/queue.py
@@ -19,10 +19,10 @@ class Buffer(Enum):
     INBOX = "inbox"
     IN_FLIGHT = "in-flight"
 
-    def name(self, suffix=None):
+    def inbox_name(self, suffix=None):
         return f"{self.value}:{suffix}" if suffix else self.value
 
-    def get_inflight_name(self, receipt: UUID = uuid4(), suffix: str = None) -> str:
+    def inflight_name(self, receipt: UUID = uuid4(), suffix: str = None) -> str:
         return f"{Buffer.IN_FLIGHT.value}:{str(suffix)}:{str(receipt)}" if suffix else f"{Buffer.IN_FLIGHT.value}:{str(receipt)}"
 
 
@@ -87,7 +87,7 @@ class RedisQueue(Queue):
     scripts: Dict[str, Any] = {}
 
     def __init__(self, suffix=None) -> None:
-        self._inbox = Buffer.INBOX.name(suffix)
+        self._inbox = Buffer.INBOX.inbox_name(suffix)
         self._suffix = suffix
 
     def init_app(self, redis):
@@ -96,12 +96,12 @@ class RedisQueue(Queue):
 
     def poll(self, count=10) -> tuple[UUID, list[str]]:
         receipt = uuid4()
-        in_flight_key = Buffer.IN_FLIGHT.get_inflight_name(receipt, self._suffix)
+        in_flight_key = Buffer.IN_FLIGHT.inflight_name(receipt, self._suffix)
         results = self.__move_to_inflight(in_flight_key, count)
         return (receipt, results)
 
     def acknowledge(self, receipt: UUID):
-        inflight_name = Buffer.IN_FLIGHT.get_inflight_name(receipt, self._suffix)
+        inflight_name = Buffer.IN_FLIGHT.inflight_name(receipt, self._suffix)
         self._redis_client.delete(inflight_name)
 
     def publish(self, message: str):
--- a/app/queue.py
+++ b/app/queue.py
@@ -22,7 +22,7 @@ class Buffer(Enum):
     def name(self, suffix=None):
         return f"{self.value}:{suffix}" if suffix else self.value
 
-    def get_inflight_name(self, receipt: UUID = uuid4(), suffix: str = "") -> str:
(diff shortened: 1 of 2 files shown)


app/queue.py
```diff
@@ -22,6 +22,9 @@ class Buffer(Enum):
     def name(self, suffix=None):
         return f"{self.value}:{suffix}" if suffix else self.value
 
+    def get_inflight_name(self, receipt: UUID = uuid4(), suffix: str = None) -> str:
+        return f"{Buffer.IN_FLIGHT.value}:{str(suffix)}{str(receipt)}"
```
sastels, line 26: if `suffix` is `None` then the String "None" will be part of the name - is that what we want?

jimleroyer: Can we separate each value with a `:`, big data design style so we can clearly see each different segment?

app/queue.py
```diff
@@ -22,6 +22,9 @@ class Buffer(Enum):
     def name(self, suffix=None):
         return f"{self.value}:{suffix}" if suffix else self.value
 
+    def get_inflight_name(self, receipt: UUID = uuid4(), suffix: str = "") -> str:
+        return f"{Buffer.IN_FLIGHT.value}:{str(suffix)}:{str(receipt)}"
```
jimleroyer, line 26: @iokpala There is a chance there will be two consecutive `:` if the `suffix` parameter is empty. Can we please make the parameter default to `None` instead and have a ternary operator (similar to line 23 for the `name` function) so that it removes the extra `:` if suffix is not defined?

iokpala: I noticed that as well..  Will fix it

tests/app/test_queue.py
```diff
@@ -16,15 +16,13 @@ def pmr_redis_config():
 
 redis = create_redis_fixture(scope="function")
 REDIS_ELEMENTS_COUNT = 123
-QNAME_SUFFIX = "test"
```
jimleroyer, line 19: Can we keep this constant but instantiate it to app.config["NOTIFY_ENVIRONMENT"]) instead? That would be easier to read in the test code.

app/queue.py
```diff
@@ -22,6 +22,9 @@ class Buffer(Enum):
     def name(self, suffix=None):
         return f"{self.value}:{suffix}" if suffix else self.value
 
+    def get_inflight_name(self, receipt: UUID = uuid4(), suffix: str = "") -> str:
```
jimleroyer, line 25: Can we please pass in `None` to the `suffix` parameter?

tests/app/test_queue.py
```diff
@@ -75,33 +74,33 @@ def delete_all_inflight(self, redis):
             redis.delete(key)
 
     @pytest.mark.serial
-    def test_put_mesages(self, redis, redis_queue):
+    def test_put_mesages(self, app, redis, redis_queue):
```
jimleroyer, line 77: This is not necessary to define the `app` mock as a parameter because of the `autouse=True` usage [at this line](https://github.com/cds-snc/notification-api/pull/1451/files#diff-c9d36498371981109d1bf2de7b12bb84964b6f4a6b55a03af7bf94ca02781366R22).

tests/app/test_queue.py
```diff
@@ -75,33 +74,33 @@ def delete_all_inflight(self, redis):
...
         element = generate_element()
         redis_queue.publish(element)
-        assert redis.llen(Buffer.INBOX.name(QNAME_SUFFIX)) == 1
+        assert redis.llen(Buffer.INBOX.name(app.config["NOTIFY_ENVIRONMENT"])) == 1
         self.delete_all_list(redis)
 
     @pytest.mark.serial
-    def test_polling_message(self, redis, redis_queue):
+    def test_polling_message(self, app, redis, redis_queue):
```
jimleroyer, line 84: This is not necessary to define the `app` mock as a parameter because of the `autouse=True` usage [at this line](https://github.com/cds-snc/notification-api/pull/1451/files#diff-c9d36498371981109d1bf2de7b12bb84964b6f4a6b55a03af7bf94ca02781366R22).

tests/app/test_queue.py
```diff
@@ -75,33 +74,33 @@ def delete_all_inflight(self, redis):
...
             assert isinstance(elements[0], str)
-            assert redis.llen(Buffer.INBOX.name(QNAME_SUFFIX)) == 0
-            assert redis.llen(redis_queue.get_inflight_name(receipt)) == 1
+            assert redis.llen(Buffer.INBOX.name(app.config["NOTIFY_ENVIRONMENT"])) == 0
+            assert redis.llen(Buffer.IN_FLIGHT.get_inflight_name(receipt, app.config["NOTIFY_ENVIRONMENT"])) == 1
 
     @pytest.mark.serial
     @pytest.mark.parametrize("count", [0, 1, 98, 99, 100, 101, REDIS_ELEMENTS_COUNT, REDIS_ELEMENTS_COUNT + 1, 500])
-    def test_polling_many_messages(self, redis, redis_queue, count):
+    def test_polling_many_messages(self, app, redis, redis_queue, count):
```
jimleroyer, line 94: This is not necessary to define the `app` mock as a parameter because of the `autouse=True` usage [at this line](https://github.com/cds-snc/notification-api/pull/1451/files#diff-c9d36498371981109d1bf2de7b12bb84964b6f4a6b55a03af7bf94ca02781366R22).

tests/app/test_queue.py
```diff
@@ -117,46 +116,46 @@ def test_polling_message_with_custom_inbox_name(self, redis, suffix):
...
             redis_queue.acknowledge(receipt)
             assert redis.llen(Buffer.INBOX.name(suffix)) == 0
-            assert redis.llen(redis_queue.get_inflight_name(receipt)) == 0
+            assert redis.llen(Buffer.IN_FLIGHT.get_inflight_name(receipt, suffix)) == 0
         finally:
             self.delete_all_list(redis)
 
     @pytest.mark.serial
-    def test_polling_with_empty_inbox(self, redis, redis_queue):
+    def test_polling_with_empty_inbox(self, app, redis, redis_queue):
```
jimleroyer, line 128: This is not necessary to define the `app` mock as a parameter because of the `autouse=True` usage [at this line](https://github.com/cds-snc/notification-api/pull/1451/files#diff-c9d36498371981109d1bf2de7b12bb84964b6f4a6b55a03af7bf94ca02781366R22).

tests/app/test_queue.py
```diff
@@ -117,46 +116,46 @@ def test_polling_message_with_custom_inbox_name(self, redis, suffix):
...
         (receipt, elements) = redis_queue.poll(10)
         assert len(elements) == 0
-        assert redis.llen(Buffer.INBOX.name(QNAME_SUFFIX)) == 0
-        assert redis.llen(redis_queue.get_inflight_name(receipt)) == 0
+        assert redis.llen(Buffer.INBOX.name(app.config["NOTIFY_ENVIRONMENT"])) == 0
+        assert redis.llen(Buffer.IN_FLIGHT.get_inflight_name(receipt, app.config["NOTIFY_ENVIRONMENT"])) == 0
 
     @pytest.mark.serial
-    def test_polling_with_zero_count(self, redis, redis_queue):
+    def test_polling_with_zero_count(self, app, redis, redis_queue):
```
jimleroyer, line 136: same here..

tests/app/test_queue.py
```diff
@@ -117,46 +116,46 @@ def test_polling_message_with_custom_inbox_name(self, redis, suffix):
...
             (receipt, elements) = redis_queue.poll(-1)
             assert len(elements) == 0
-            assert redis.llen(Buffer.INBOX.name(QNAME_SUFFIX)) == 1
-            assert redis.llen(redis_queue.get_inflight_name(receipt)) == 0
+            assert redis.llen(Buffer.INBOX.name(app.config["NOTIFY_ENVIRONMENT"])) == 1
+            assert redis.llen(Buffer.IN_FLIGHT.get_inflight_name(receipt, app.config["NOTIFY_ENVIRONMENT"])) == 0
 
     @pytest.mark.serial
-    def test_acknowledged_messages(self, redis, redis_queue):
+    def test_acknowledged_messages(self, app, redis, redis_queue):
```
jimleroyer, line 152: same here..

jimleroyer (commented):
Thank you for making the change for avoiding the double `:` scenario.

Added a few more comments on other changes.

app/queue.py
iokpala: Sone

tests/app/test_queue.py
iokpala: I am sure i understand you here.. `app` has been added so we have access to the app context and config.. so we can `app.config["NOTIFY_ENVIRONMENT"]) `

tests/app/test_queue.py
iokpala: Sorry I dont agree.. `QNAME_SUFFIX ` is surplus to requirements.. hence not DRY

iokpala: the idea of a suffic appear broad, and put in place of the `environment` it seems

tests/app/test_queue.py
iokpala: An example would be helpful.

tests/app/test_queue.py
iokpala: An example would be helpful.

tests/app/test_queue.py
iokpala: An example would be helpful.?

tests/app/test_queue.py
iokpala: If we are to go with DAMP then `suffix` appears vague because its isn't descriptive because the sections of the name appear to follow this pattern `queue:environment:id`

jimleroyer: Ok I think I see the misunderstanding. The suffix does not follow `queue:environment:id` pattern you described. It does not make sense anyway with the examples I provided in my last reply. The suffix can be anything deemed good and descriptive for the usage of the queue, i.e. `sms` or `email`. 

This derives from a requirement that Jumana explained earlier this week in a call where we were all present where we need different queues based on the notification type... but that will expand in the future on the notification priority as well.

jimleroyer: Also, at that point, we are talking about the parameter name `suffix` which is out of scope of the initial talk around the usage of a constant in a test. So I'll consider this closed for now. We can change the parameter name later on if we want to in a separate PR, but for now, I will merge this one based on the recent changes I brought to it. We need to move forward and the scope for this PR were:

* have a way to customize inflight name
* write tests around it and avoid repetition of the double `:`

jimleroyer (approved):
LGTM

PR #1454, opened 2022-02-03
locust test with individual emails

sastels:
# Summary | Résumé

Just do a locust performance test of individual emails, to put as much stress on api and the database

Each user sends 200 emails / minute

Set environment variables
```
PERF_TEST_AUTH_HEADER="apikey-v1 xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
PERF_TEST_EMAIL_TEMPLATE_ID=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
```
(email template with no variables)

For approximately 1000 emails / minute:
```
locust -f ./individual-emails.py --headless --users=5  --run-time=1m
```

None.

      fonctionnalité existante?


tests-perf/individual_emails/individual-emails.py
```diff
@@ -0,0 +1,28 @@
+import os
+
+from dotenv import load_dotenv
+from locust import HttpUser, constant_pacing, task
+
+load_dotenv()
+
+
+class NotifyApiUser(HttpUser):
+    wait_time = constant_pacing(0.3)  # about 200 / minute hopefully
```
jimleroyer, line 10: Not sure if the configuration is right here. Can you tell me more about the desired goal? Why not use constant_throughput of 1 minute for example with a number of users set to 200 for example?

tests-perf/individual_emails/locust.conf
```diff
@@ -0,0 +1,3 @@
+host = https://api-k8s.staging.notification.cdssandbox.xyz
```
jimleroyer, line 1: We only want to hit the kubernetes cluster on that?

jimleroyer (commented):
Nice addition to our set of performance tests. Just have a few questions.

tests-perf/individual_emails/individual-emails.py
sastels: Using 200 users hitting every minute means we get a big spike at the 1 minute marks and nothing in between. I'm trying to more closely emulate what happens in prod, where a service will send constantly at a high rate.

There is a locust pacing function `constant_throughput`, but it isn't available in the version of locust we use here, and we can't bump locust up because we're using an old flask 😞 

It might be useful to move all our performance / smoke testing code to its own repo where we are freer to install packages just needed by the tests.

jimleroyer: That might be another case for upgrading werkzeug? so we can upgrade flask.

jimleroyer: Why not set the constant_pacing to every second though? so we get an easy to calculate and to expect formula and a relatively constant throughput?

sastels:
Thinking that maybe this should also be moved to notification-attic - thoughts?

jimleroyer:
@sastels I'd be good with the move to attic too! and make a clean up in the API folder. We can do it in this PR or later, up to you!

jimleroyer (approved):
LGTM!

sastels:
I shall merge for now - I think there's a card about refactoring tests, I'll add to it about possibly moving to the attic.

PR #1455, opened 2022-02-07
fix: setup of VS Code devcontainer

patheard:
# Summary
Fix the local build of the VS Code devcontainer by:

* updating the starship prompt install;
* setting the default shell to `/bin/bash`; and
* updating the terminal font to fallback to `monospace`.


jimleroyer (approved):

jimleroyer:
Thanks for the improvements!

PR #1461, opened 2022-02-11
feat: add kubectl to devcontainer

patheard:
# Summary
Add kubectl, [kubectx](https://github.com/ahmetb/kubectx/) and command auto-complete to
the devcontainer.  This lets us monitor how the cluster
is behaving during API performance testing.

Also adds the [GitHub Copilot](https://copilot.github.com/) VS Code extension.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -33,7 +33,8 @@
 		"visualstudioexptteam.vscodeintellicode",
 		"yzhang.markdown-all-in-one",
 		"ms-ossdata.vscode-postgresql",
-		"googlecloudtools.cloudcode"
+		"googlecloudtools.cloudcode",
+		"GitHub.copilot"
 	],
 
 	"postCreateCommand": "notify-dev-entrypoint.sh"
--- a/.devcontainer/scripts/notify-dev-entrypoint.sh
+++ b/.devcontainer/scripts/notify-dev-entrypoint.sh
@@ -16,6 +16,12 @@ echo -e "alias l='exa -alh'" >> ~/.bashrc
 echo -e "alias ll='exa -alh@ --git'" >> ~/.bashrc
 echo -e "alias lt='exa -al -T -L 2'" >> ~/.bashrc
 
+# Kubectl aliases and command autocomplete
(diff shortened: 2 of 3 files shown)


.devcontainer/scripts/notify-dev-entrypoint.sh
```diff
@@ -16,6 +16,12 @@ echo -e "alias l='exa -alh'" >> ~/.bashrc
 echo -e "alias ll='exa -alh@ --git'" >> ~/.bashrc
 echo -e "alias lt='exa -al -T -L 2'" >> ~/.bashrc
 
+# Kubectl aliases and command autocomplete
+echo -e "alias k='kubectl'" >> ~/.bashrc
+echo -e "source <(kubectl completion bash)" >> ~/.bashrc
+echo -e "complete -F __start_kubectl k" >> ~/.bashrc
+echo -e "source /usr/share/bash-completion/bash_completion" >> ~/.bashrc
```
jimleroyer, line 23: Could be nice to use oh my zsh instead, dump starship. The former offers so much for auto completion and plugins.

.devcontainer/devcontainer.json
```diff
@@ -33,7 +33,8 @@
 		"visualstudioexptteam.vscodeintellicode",
 		"yzhang.markdown-all-in-one",
 		"ms-ossdata.vscode-postgresql",
-		"googlecloudtools.cloudcode"
+		"googlecloudtools.cloudcode",
+		"GitHub.copilot"
```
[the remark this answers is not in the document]
patheard: Correct - if it proves to be annoying for people without accounts, I more than happy to remove and only install myself.

.devcontainer/scripts/notify-dev-entrypoint.sh
patheard: I'm definitely game for switching to `zsh`, I had just thought there was a `starship` preference.  I'll get that added!

jimleroyer: I like starship as it's cross platform and written in Rust (I contributed a bit to it) -- but oh-my-zsh might play much better for devcontainers and all of its plugins. 🙃

PR #1465, opened 2022-02-16
feat: scaffold custom metrics

mohamed-cds:
Custom Cloudwatch logs
Added cloud watch metrics for different part of the custom queue implementation.

Note: needs to be tested in staging

### DIFF
--- a/app/aws/metrics.py
+++ b/app/aws/metrics.py
@@ -31,5 +31,4 @@ def put_batch_saving_metric(queue: RedisQueue, count: int, metrics: MetricsLogge
     except ClientError as e:
         message = "Error sending CloudWatch Metric: {}".format(e)
         current_app.logger.exception(message)
-        pass
     return
--- a/app/aws/metrics.py
+++ b/app/aws/metrics.py
@@ -23,11 +23,76 @@ Metrics_Config.log_group_name = "BatchSaving"
 
 @metric_scope
 def put_batch_saving_metric(queue: RedisQueue, count: int, metrics: MetricsLogger):
+    """
+    Metric to calculate how many items are put in an INBOX
+
+    Args:
+        queue (RedisQueue): Implementation of queue.RedisQueue for BatchSaving
+        count (int): default: 1, count of an item added to the INBOX.
+        metrics (MetricsLogger): Submit metric to cloudwatch
+    """
     try:
         metrics.set_namespace("BatchSaving")
-        metrics.set_dimensions({"inbox": queue._inbox})
+        metrics.set_dimensions({"list_name": queue._inbox})
         metrics.put_metric("published", count, "Count")
         metrics.set_property("expiry", queue._expire_inflight_after_seconds)
+    except ClientError as e:
+        message = "Error sending CloudWatch Metric: {}".format(e)
+        current_app.logger.exception(message)
+    return
+
+
+@metric_scope
+def put_batch_saving_in_flight_metric(count: int, metrics: MetricsLogger):
+    """
+    Metric to calculate how many inflight lists have been created
+
+    Args:
+        count (int): default: 1, count of an inflight list created.
+        metrics (MetricsLogger): Submit metric to cloudwatch
+    """
+    try:
+        metrics.set_namespace("BatchSaving")
+        metrics.put_metric("inflight", count, "Count")
+        metrics.set_dimensions({"created": True})
+    except ClientError as e:
+        message = "Error sending CloudWatch Metric: {}".format(e)
+        current_app.logger.exception(message)
+    return
+
+
+@metric_scope
+def put_batch_saving_inflight_processed(count: int, metrics: MetricsLogger):
+    """
+    Metric to calculate how many inflight lists have been processed.
+
+    Args:
+        count (int): default: 1, count of an inflight list created.
+        metrics (MetricsLogger): Submit metric to cloudwatch
+    """
+    try:
+        metrics.set_namespace("BatchSaving")
+        metrics.put_metric("inflight", count, "Count")
+        metrics.set_dimensions({"acknowledged": True})
+    except ClientError as e:
+        message = "Error sending CloudWatch Metric: {}".format(e)
+        current_app.logger.exception(message)
+    return
+
+
+@metric_scope
+def put_batch_saving_expiry_metric(count: int, metrics: MetricsLogger):
+    """
+    Metric to calculate how many inflight list have not been processed and instead
+    sent back to the inbox.
+
+    Args:
+        count (int): Number of inlfight lists sent to inbox
+        metrics (MetricsLogger): Submit metric to cloudwatch
+    """
+    try:
+        metrics.set_namespace("BatchSaving")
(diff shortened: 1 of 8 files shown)


app/aws/metrics.py
```diff
@@ -0,0 +1,99 @@
...
+from flask import current_app
+
+from app.config import Config
+
+if TYPE_CHECKING:  # A special Python 3 constant that is assumed to be True by 3rd party static type checkers
+    from aws_embedded_metrics.logger.metrics_logger import MetricsLogger  # type: ignore
+
+    from app.queue import RedisQueue
+
+Metrics_Config = get_config()
```
jimleroyer, line 17: To follow usual variable and lower snake case conventions:

```suggestion
metrics_config = get_config()
```

app/aws/metrics.py
```diff
@@ -0,0 +1,99 @@
...
+Metrics_Config.agent_endpoint = Config.CLOUDWATCH_AGENT_ENDPOINT
+Metrics_Config.service_name = "BatchSaving"
+Metrics_Config.service_type = "Redis"
+Metrics_Config.log_group_name = "BatchSaving"
```
jimleroyer, line 21: Would that run fine if ran in an environment without a cloudwatch agent endpoint? Modules/objects in Notify follow a pattern of bootstrapping configuration via an `init_app` method from the [application startup over here](https://github.com/cds-snc/notification-api/blob/9461b26f9582bdea6fe7c68b3d360597e5eab9ba/app/__init__.py#L60) and an example [with Celery over there](https://github.com/cds-snc/notification-api/blob/ce27097cf39fb9615771e5d581e3abdcf0bc2668/app/celery/celery.py#L38).

app/queue.py
```diff
@@ -105,20 +112,24 @@ def poll(self, count=10) -> tuple[UUID, list[str]]:
...
             current_app.logger.warning(f"Moved inflights {expired} back to inbox {self._inbox}")
 
     def acknowledge(self, receipt: UUID):
         inflight_name = Buffer.IN_FLIGHT.inflight_name(receipt, self._suffix)
         self._redis_client.delete(inflight_name)
+        put_batch_saving_inflight_processed(1)
 
     def publish(self, message: str):
         self._redis_client.rpush(self._inbox, message)
+        put_batch_saving_metric(self, 1)
```
jimleroyer, line 132: Do we have a quota on how many calls we are allowed to send to AWS metrics? Thinking of the case where traffic will be high and many publish might happen per second.

tests/app/test_queue.py
```diff
@@ -203,6 +229,18 @@ def test_expire_inflights(self, redis, redis_queue):
...
+        with self.given_inbox_with_many_indexes(redis, redis_queue):
+            pbsem_mock = mocker.patch("app.queue.put_batch_saving_expiry_metric")
+            Buffer.INBOX.inbox_name(QNAME_SUFFIX)
+            redis.set("not_inflight", "test")
+            redis_queue.poll(10)
+            redis_queue.poll(10)
+            redis_queue.poll(10)
+            time.sleep(2)
+            redis_queue.expire_inflights()
+            assert pbsem_mock.assert_called_with(3) is None
```
jimleroyer, line 242: Do you think we could put the new tests in a different class such as ` TestRedisQueueMetrics`? As it does not test the core feature of the queue but metrics around it? No strong take on that; we'd leverage the classes to test different features.

app/aws/metrics.py
```diff
@@ -0,0 +1,99 @@
...
+from typing import TYPE_CHECKING
+
+from aws_embedded_metrics import metric_scope  # type: ignore
+from aws_embedded_metrics.config import get_config  # type: ignore
+from botocore.exceptions import ClientError
+from flask import current_app
+
+from app.config import Config
+
+if TYPE_CHECKING:  # A special Python 3 constant that is assumed to be True by 3rd party static type checkers
```
jimleroyer, line 12: Interesting usage..

jimleroyer (commented):
Nice work, I love the simplicity of it, how quick that came up and the testing.

requirements-app.txt
```diff
@@ -42,6 +42,8 @@ more-itertools==8.12.0
 awscli-cwlogs>=1.4.6,<1.5
 aws-embedded-metrics==1.0.7
 
+# Putting upgrade on hold due to v2 introducing breaking changes
+Werkzeug==1.0.1
```
jimleroyer, line 46: Yes we should have pinned this before, thank you.

app/queue.py
mohamed-cds: No concerns related to volume. CloudWatch embedded metrics is a two step process. Our app will log to a CloudWatch log group using a specific format. CloudWatch will detect this format and automatically generate the metrics async.

app/aws/metrics.py
jzbahrai: @jimleroyer We might be okay, but we are in the process of testing this to be sure. Do you want us to change the initialization regardless though?

jzbahrai: We tested this locally and there is not issue with invocating the cloud watch metric

jimleroyer: My concern was more if configuration is missing, what happens? Will devs setting it up locally will have to provide the endpoint to run Notify? If not, it's ok. Else I'd suggest we make it easy to disable it if the endpoint is missing.

app/queue.py
jimleroyer: It seems we are allowed 1,000,000 API calls per month. We'll probably make a couple thousands per month (if HC sends 40,000 every business days, that would be 800,000 per month for that specific service and only for the API call going through the publish buffer queue).

I was wondering more about if there is a rate limit on AWS side, how many times per second or minute we can hit them, considering the bursty traffic, it needs to be at least 6,000 a minute.

tests/app/test_queue.py
jzbahrai: done :+1

app/aws/metrics.py
jzbahrai: So looking at the local logs, regardless of the endpoint being configured or not we see the data is submitted but doesn't wait for an ack:
```
[2022-02-21 21:04:00,946: INFO/MainProcess] Received task: heartbeart-inbox-sms[4c2178c2-3b8f-46ae-9cfa-22d5bb66e182]  
[2022-02-21 21:04:01,033: INFO/ForkPoolWorker-2] Parsed agent endpoint (udp) 127.0.0.1:8125
[2022-02-21 21:04:01,033: INFO/ForkPoolWorker-2] Submitted metrics to agent over UDP
```

app/queue.py
mohamed-cds: From what i know that only applies to the standard CloudWatch metrics where you invoke `PutMetrics`.

Embedded metrics are intended for high volumes and i couldn't find any service limits related to the amount of data stored with the exception of a message size of `256Kb`. Majority of the cost with this service is the number of unique metrics you log.

At the moment we're generating `3` unique metrics for Batch savings.

jimleroyer (approved):
LGTM :shipit:

PR #1468, opened 2022-02-17
Rework notifications buffer queue ack on failure 

jimleroyer:
# Summary | Résumé

Rework notifications in following ways:

* Acknowledge notifications from the buffer queue as soon as these have been persisted in the database. At that point, the notifications are in good hands and can be removed from the buffer queues.
* Redirect failed notifications to individual atomic save. Remove from the buffer queue once this has been done.

There are a few minor refactoring such as renaming stuff around.

_I would like to add two more unit tests ideally._

New unit tests should cover the modified behavior and regression tests should cover the existing features.

Make sure the feature flags are properly setup to test in staging and disable in production.

      fonctionnalité existante?


app/celery/tasks.py
```diff
@@ -635,49 +632,57 @@ def update_letter_notifications_to_error(self, notification_references):
...
-                    notif=notification_id,
-                    receipt=receipt,
-                )
+        service = notification["service"]
+        # Sometimes, SQS plays the same message twice. We should be able to catch an IntegrityError, but it seems
+        # SQLAlchemy is throwing a FlushError. So we check if the notification id already exists then do not
+        # send to the retry queue.
+        found = get_notification_by_id(notification_id)
+        if not found and service:
+            forward_msg = "Batch saving: forwarding notification {notif} to individual save from receipt {receipt}.".format(
```
patheard, line 670: At this point, do you want to `break` from the loop so the `else` statement below doesn't execute?

app/celery/tasks.py
```diff
@@ -635,49 +632,57 @@ def update_letter_notifications_to_error(self, notification_references):
...
-                task.retry(queue=QueueNames.RETRY, exc=exc)
-            except task.MaxRetriesExceededError:
-                current_app.logger.error(f"Max retry failed: {retry_msg}")
+            current_app.logger.info(forward_msg)
+            save_fn = save_email if notification_type == EMAIL_TYPE else save_sms
+            save_fn.apply_async(
+                (service.id, notification_id, signed, None),
+                queue=QueueNames.DATABASE if not service.research_mode else QueueNames.RESEARCH_MODE,
+            )
+    else:  # end of the loop, purge the notifications from the buffer queue:
```
patheard, line 680: Should this `receipt` purge always occur at the end of the loop?  If yes, we could remove the `else` from the `for` loop.

app/celery/tasks.py
```diff
@@ -453,15 +451,18 @@ def save_emails(self, service_id: str, signed_notification: List[Any], receipt:
...
-        saved_notifications = persist_notifications(decrypted_notifications)
+        # If the data is not present in the encrypted data then fallback on whats needed for process_job
+        saved_notifications = persist_notifications(verified_notifications)
+        if receipt:
+            email_queue.acknowledge(receipt)
+            current_app.logger.info(f"Batch saving: {receipt} removed from buffer queue.")
     except SQLAlchemyError as e:
-        handle_list_of_exception(self, decrypted_notifications, e, receipt)
+        signed_and_verified = list(zip(signed_notifications, verified_notifications))
+        handle_batch_error_and_forward(self, signed_and_verified, EMAIL_TYPE, e, receipt)
```
patheard, line 465: I like this pattern of sending the failed batch through to process each notification individually and then removing the list from the Redis queue.

patheard (commented):
Looks good - just had a question about the `for/else` pattern in ` handle_batch_error_and_forward`.

app/celery/tasks.py
jimleroyer: I thought we initially did this but I was wrong. It's a bit weird to have an idea but based on a mistake. Steve also proposed the same after the incident happened, so it seems quite natural.

app/celery/tasks.py
jimleroyer: Yes absolutely, I will remove that `else` clause.

app/celery/tasks.py
jimleroyer: Yep!

patheard (commented):
One other comment I missed, which you might already have in mind, would be to add some unit tests around this change in behaviour if possible.   

Other than that I think it's definitely a good candidate for testing to see if it removes the Redis tail eating behaviour.

jimleroyer:
@patheard I added two unit tests for testing the failing scenario where the save could not happen. I would like to add two more for when the save failed but the notifications were actually found in the database (and hence are duplicates). This should be a quick one tomorrow morning as I now can reuse the new ones as skeleton code with proper mocking.

tests/app/celery/test_tasks.py
```diff
@@ -98,8 +98,9 @@ def email_job_with_placeholders(notify_db, notify_db_session, sample_email_templ
     return create_job(template=sample_email_template_with_placeholders)
 
 
+@pytest.mark.usefixtures("notify_db_session")
 class TestBatchSaving:
-    def test_should_save_smss(self, notify_db_session, sample_template_with_placeholders, mocker):
```
patheard, line 102: heh, I like this fixture removal - I find it odd how the pattern is to include it in so many of our unit tests when it is rarely used

patheard (approved):
🎉

tests/app/celery/test_tasks.py
jimleroyer: Yeah there might be a better pattern to teardown the database tables and session than specifying this. Using test classes though allowed me to move this at line 101 instead, so it will be used for all tests with class. 👍

jzbahrai (approved):

PR #1469, opened 2022-02-17
test: locust performance test with bounce rate

patheard:
# Summary
Add a performance test that supports a `PERF_TEST_BOUNCE_RATE`
environment variable (value between 0 and 1) that determines the
percentage of requests that will be delivered to the
`PERF_TEST_EMAIL_BOUNCE` email address.

All other emails will be sent to the `PERF_TEST_EMAIL_SUCCESS`
email address.

# Related
* cds-snc/notification-planning#412


mohamed-cds (approved):
If this doesn't trigger it, we can also try filling up the sms queue with their fail/success equivalents. 🚀 ✅

PR #1474, opened 2022-02-17
Update: Fix string to boolean environment variables conversion with AST

iokpala:
Related to #1473 

# Summary | Résumé

Within is method defined to convert possible boolean strings to boolean

None.

---

> mentionné.

      fonctionnalité existante?


app/config.py
```diff
@@ -20,6 +21,13 @@
     extract_cloudfoundry_config()
 
 
+def str_to_bool(env_val) -> bool:
+    try:
+        return ast.literal_eval(env_val)
+    except Exception:
+        return False
```
jimleroyer, line 28: Thinking here... I think we might benefit from actually letting the exception raise instead of catching it. Wrong configuration value and interpretation for us, failing early could actually be helpful and avoid situations like the one we had this week.

iokpala: I added this because if you recall during mixpanal.. I did allow the exception to be thrown..

jimleroyer: Yeah, MixPanel is a case where we felt we didn't want the application to break if it was missing. In this case, a misconfiguration can be a production risk, so I would like this to fail early. Hope that makes sense!

jimleroyer:
@iokpala Could we apply the same logic in admin? I wonder too if the application will fail with existing configurations, what would be the plan for these?

iokpala:
> @iokpala Could we apply the same logic in admin? 

Sure thing

> I wonder too if the application will fail with existing configurations, what would be the plan for these? 

It won't because this method and it's application on the env variable is in isolation 

> Just wondering if we merge this right now, what would be the potential behavior.


The tests suggests that all will be fine.

app/config.py
jimleroyer: I had a chat with @patheard and @mohamed-cds today, especially around the lambda api and how it might crashes if we don't catch the exception and rely on defaults. The kubernetes pods have a fail safe where new versions won't get deployed if an error is detected at start up but the AWS lambda API won't have the same unfortunately. 

Hence we will revert to catching the exception and rely on a default, considering the unsafe aspect of switching AWS lambdas.

For other components such as the admin running uniquely in the kubernetes pods, the option to raise the exception and not catch it could make sense.

jimleroyer:
> > Just wondering if we merge this right now, what would be the potential behavior.

> The tests suggests that all will be fine.

I don't think we have tests that cover this, i.e. to take current configuration values sitting in `.env` files and seeing how the application works with these. We discovered that some feature flag we thought should be disabled are in fact enabled. I'd like for us to be cautious because it seems we got burned with this in the past but we didn't notice it. These changes should make these more obvious though, it's good stuff. 👍

app/config.py
patheard: I updated this to explicitly test for the String values of `"True"` or `"False"` and return the default value in all other cases.  

My reasoning was that in a case of calling `str_to_bool(os.getenv("NON_EXISTENT_KEY))`, the `str_to_bool` function would receive the value of `None` and `ast.literal_eval` would dutifully return this as `None` rather than the boolean value we're expecting.

jimleroyer: Alright. Can we though perform a `lower` and `strip` operation on the `str_val` parameter so the case does not matter?

patheard: Absolutely - I'll get that added.

tests/app/test_config.py
```diff
@@ -82,3 +82,29 @@ def test_queue_names_all_queues_correct():
...
+    os.environ["FF_REDIS_BATCH_SAVING"] = "false"
+    assert str_to_bool(os.getenv("FF_REDIS_BATCH_SAVING"), True) is True
+
+    os.environ["FF_REDIS_BATCH_SAVING"] = ""
+    assert str_to_bool(os.getenv("FF_REDIS_BATCH_SAVING"), True) is True
+
+    os.environ["FF_REDIS_BATCH_SAVING"] = "mmmm cheese"
+    assert str_to_bool(os.getenv("FF_REDIS_BATCH_SAVING"), True) is True
+
+    assert str_to_bool(os.getenv("NON_EXISTENT_ENV_VAR"), False) is False
```
jimleroyer, line 108: 👍

mohamed-cds (commented):
Have we considered using https://pypi.org/project/environs ? It supports enforcing types and will allow us to have more strict checks for all our envs if we wanted to expand beyond feature flags.

`env.bool('ENABLE_LOGIN', False)`

app/config.py
patheard: Added!

tests/app/test_config.py
```diff
@@ -82,3 +82,35 @@ def test_queue_names_all_queues_correct():
...
+    assert str_to_bool(os.getenv("SOME_OTHER_ENV_VAR"), False) is False
+
+    os.environ["FF_REDIS_BATCH_SAVING"] = "true false"
+    assert str_to_bool(os.getenv("FF_REDIS_BATCH_SAVING"), True) is True
+
+    os.environ["FF_REDIS_BATCH_SAVING"] = ""
+    assert str_to_bool(os.getenv("FF_REDIS_BATCH_SAVING"), True) is True
+
+    assert str_to_bool(os.getenv("NON_EXISTENT_ENV_VAR"), False) is False
+    assert str_to_bool("mmmm cheese", True) is True
```
jimleroyer, line 114: 🧀

jimleroyer (approved):
LGTM!

jimleroyer:
@mohamed-cds We have not AFAIK but that looks nice to investigate and replace if that is a good fit. That PR for now should unblock our release pipeline hopefully.

PR #1475, opened 2022-02-17
Bulk Notification Insert : Add test for duplicate ids

jzbahrai:
# Summary | Résumé

When we have a list of notifications and we use the bulk save functionality, if there are duplicate IDs, we should throw an error.
The code in this PR adds error handling and a test to handle the above scenario.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -17,7 +17,6 @@ from notifications_utils.timezones import (
 )
 from sqlalchemy import asc, desc, func
 from sqlalchemy.dialects.postgresql import insert
-from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import joinedload
 from sqlalchemy.orm.exc import NoResultFound
 from sqlalchemy.sql import functions
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -17,6 +17,7 @@ from notifications_utils.timezones import (
 )
 from sqlalchemy import asc, desc, func
 from sqlalchemy.dialects.postgresql import insert
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import joinedload
 from sqlalchemy.orm.exc import NoResultFound
 from sqlalchemy.sql import functions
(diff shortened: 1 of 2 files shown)


tests/app/dao/notification_dao/test_notification_dao.py
```diff
@@ -1769,13 +1769,25 @@ def test_send_method_stats_by_service(sample_service, sample_organisation):
...
+    def test_bulk_insert_notification_duplicate_ids(self, sample_template):
+        assert len(Notification.query.all()) == 0
+        n1 = create_notification(sample_template, client_reference="happy")
+        n2 = create_notification(sample_template, client_reference="sad")
+        n3 = create_notification(sample_template, client_reference="loud")
+        n1.id = n2.id
+        n1.status = n2.status
+        with pytest.raises(Exception):
+            bulk_insert_notifications([n1, n2, n3])
+        assert len(get_notifications_for_service(sample_template.service_id).items) == 0
```
jimleroyer, line 1793: Perfecto!

app/dao/notifications_dao.py
```diff
@@ -107,7 +108,11 @@ def bulk_insert_notifications(notifications):
             notification.status = NOTIFICATION_CREATED
 
     # TODO: Add error handling (Redis queue?) for failed notifications
-    return db.session.bulk_save_objects(notifications)
+    try:
+        return db.session.bulk_save_objects(notifications)
+    except IntegrityError as e:
+        current_app.logger.error(e)
+        raise
```
jimleroyer, line 115: I don't think to be logged here if we raise it again: the higher layers will catch it and log. If we have it this way, that would log it twice in the logs.

jimleroyer: I am not sure I get it, can you help me understand why we catch an exception that we throw with no additional behavior? This is similar to how the code was before but with more lines.

I think we had this conversation 2-3 months ago. I would ask the same that I asked back then: can you identify best practices in the Python community that encourages this practice?

jimleroyer (approved):
LGTM!

PR #1478, opened 2022-02-21
Feat/scaffold custom metrics

mohamed-cds:
Original reverted due errors caused due to the absence of a asyncio eventloop. Test a work around where a new event loop is created instead of trying to obtain an existing one. 

related: #1465


jzbahrai (approved):
lgtm lets try it!

PR #1480, opened 2022-02-22
fix: reference the correct cloudwatch agent hostname

mohamed-cds:
Changes:
- The `STATSD_HOST` environment variable defined in the manifest project contains the hostname of the CloudWatch agent. This PR will fix the reference to use that value instead of localhost.
- Devcontainer tweaks so that the app exposes it's port so that it can be tested locally with other notification products.


app/queue.py
```diff
@@ -107,9 +107,7 @@ def __init__(self, suffix=None, expire_inflight_after_seconds=300) -> None:
 
         # Create async event loop for CloudWatch metrics
         try:
-            loop = (
-                asyncio.get_event_loop()
-            )  # This will fail to create a new event loop if called outside the main thread. https://bugs.python.org/issue39381
+            asyncio.get_event_loop()  # This will fail to create a new event loop if called outside the main thread. https://bugs.python.org/issue39381
```
jimleroyer, line 110: Hmm I wonder why there was an assignment. Nothing to do with the following one in the except block?

mohamed-cds: this is just to test that an event loop exists. If it fails it'll create one, so i had an unnecessary assignment when i introduced this the first time.

jzbahrai (approved):

PR #1481, opened 2022-02-22
Change the metric namespace to be the same as the Notification namespace

jzbahrai:
# Summary | Résumé
Change metric namespace to be `"NotificationCanadaCa"` vs `BatchSaving`


mohamed-cds (approved):
LGTM 🚀 ✅

PR #1483, opened 2022-02-23
feat: use custom metrics_scope to manage async metrics

mohamed-cds:
Bringing in the metrics_scope from the aws projects so that we can manage the async aspects of the aws embedded metrics.

### DIFF
--- a/app/aws/metrics_scope.py
+++ b/app/aws/metrics_scope.py
@@ -49,16 +49,8 @@ def metric_scope(fn):  # type: ignore
             except Exception as e:
                 raise e
             finally:
-                try:
-                    loop = (
-                        asyncio.get_event_loop()
-                    )  # This will fail to create a new event loop if called outside the main thread. https://bugs.python.org/issue39381
-                except RuntimeError as ex:
-                    if "There is no current event loop in thread" in str(ex):
-                        loop = asyncio.new_event_loop()  # https://github.com/awslabs/aws-embedded-metrics-python/issues/14
-                        asyncio.set_event_loop(loop)
-                    else:
-                        raise
+                loop = asyncio.get_event_loop()
+                loop.run_until_complete(logger.flush())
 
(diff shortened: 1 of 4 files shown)


jzbahrai (approved):

patheard (commented):
So I wouldn't even know how to start adding unit test for the `metric_scope` function, but do you think it's possible?

app/aws/metrics_scope.py
```diff
@@ -0,0 +1,55 @@
...
+from functools import wraps
+
+from aws_embedded_metrics.logger.metrics_logger_factory import (  # type: ignore
+    create_metrics_logger,
+)
+
+
+def metric_scope(fn):  # type: ignore
+
+    if asyncio.iscoroutinefunction(fn):
```
jimleroyer, line 25: I really wonder how much it can detect, I'd like to debug that. Interesting..

app/aws/metrics_scope.py
```diff
@@ -0,0 +1,55 @@
...
+
+        @wraps(fn)
+        def wrapper(*args, **kwargs):  # type: ignore
+            logger = create_metrics_logger()
+            if "metrics" in inspect.signature(fn).parameters:
+                kwargs["metrics"] = logger
+            try:
+                return fn(*args, **kwargs)
+            except Exception as e:
+                raise e
```
jimleroyer, line 50: Why raise the exception if we don't do anything with it?

app/aws/metrics_scope.py
```diff
@@ -0,0 +1,55 @@
...
+            finally:
+                loop = asyncio.new_event_loop()
+                loop.run_until_complete(logger.flush())
```
jimleroyer, line 53: Instantiating a new coroutine outside of a non coroutine function might avoid a conflict with an existing one, but I wonder if we need to avoid multiple and concurrent executions of this code path..

PR #1494, opened 2022-02-28
feat: check if redis poll returned results

mohamed-cds:
Currently metrics are being generated without checking if the redis queue poll returned results. This is resulting in metrics being inaccurate. This PR introduces an evaluation of whether the queue returned data before logging a metric.

**note:** `test_should_send_personalised_template_with_html_enabled` test will fail since validator.w3.org is currently offline and is required by the library that validates html templates.


jimleroyer (approved):
Changes LGTM -- except for the unit test failure we'd need to 👀 into and fix?

PR #1496, opened 2022-03-01
feat: migrate to requests endpoint for zendesk

mohamed-cds:
The zendesk client was originally developed to interact with the zendesk `ticket` api. Unfortunately that endpoint is only available for authenticated users. This went undetected during development since all test emails originated from users with zendesk accounts.

Migrating to use the `requests` api allows for tickets to be associated with users without zendesk accounts.


jimleroyer (approved):
LGTM :shipit: 🚢

PR #1501, opened 2022-03-08
feat: add config logging during app initialize

patheard:
# Summary
This config logging can be used to troubleshoot app init problems and confirm
that future changes to how environment variables are set work as expected.

As part of this change, a list of sensitive config keys has also been
defined to avoid leaking secrets in the app logs.

# Related
* cds-snc/notification-planning#421

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -94,6 +94,10 @@ def create_app(application, config=None):
     register_blueprint(application)
     register_v2_blueprints(application)
 
+    # Log the application configuration
+    config_values = config.get_config(config.get_sensitive_config())
+    application.logger.info(f"Notify config: {config_values}")
+
     # avoid circular imports by importing this file later
     from app.commands import setup_commands
 
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -95,8 +95,7 @@ def create_app(application, config=None):
     register_v2_blueprints(application)
 
     # Log the application configuration
-    config_values = config.get_safe_config()
-    application.logger.info(f"Notify config: {config_values}")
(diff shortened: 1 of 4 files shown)


app/__init__.py
```diff
@@ -94,6 +94,10 @@ def create_app(application, config=None):
     register_blueprint(application)
     register_v2_blueprints(application)
 
+    # Log the application configuration
+    config_values = config.get_config(config.get_sensitive_config())
```
jimleroyer, line 98: Minor improvements: I'd just have a one function call to simplify on the API consumer side, which can include this this exact line wrapped in a function such as `config.get_safe_config())`.

tests/app/test_config.py
```diff
@@ -114,3 +114,26 @@ def test_when_env_value_default_is_used(reload_config):
...
     assert str_to_bool(None, True) is True
     assert str_to_bool(None, False) is False
+
+
+def test_get_config(reload_config):
+    config.Config.ADMIN_BASE_URL = "http://foo.bar"
+    config.Config.AWS_REGION = "dark-side-of-the-moon"
+    logged_config = config.Config.get_config([])
+    assert logged_config["ADMIN_BASE_URL"] == "http://foo.bar"
+    assert logged_config["AWS_REGION"] == "dark-side-of-the-moon"
```
jimleroyer, line 124: 🌑

jimleroyer (approved):
LGTM -- with potential readability improvement

app/__init__.py
patheard: Ah yeah - I like that idea.  Added!

PR #1505, opened 2022-03-10
Task/add local configs

jzbahrai:
# Summary | Résumé

Add vscode localconfig and some configs for .env file


sastels (approved):
A good start, we can refine as needed

PR #1510, opened 2022-03-22
fix: use `export` instead of `eval`

patheard:
# Summary
Update to use `export` directly to export the environment variables
into the Lambda's execution environment.

# Related
* cds-snc/notification-planning#421


mohamed-cds (approved):
the world is a better place with one less eval. 🚀 ✅

PR #1515, opened 2022-04-05
fix: only log when no notification found

patheard:
# Summary
Update the logging added to `get_notification_with_personalisation`
so that it only logs the query filters if a NoResultsFound exception
is raised.

# Related
* #1514
* cds-snc/notification-planning#537


jzbahrai (approved):
lgtm!

PR #1516, opened 2022-04-05
fix: add stack trace when NoResultsFound

patheard:
# Summary 
Update `get_notification_with_personalisation` exception
logging to include the stack trace.

# Results
* #1514


sastels (approved):
Nice.

PR #1518, opened 2022-04-05
fix: improve how the stack is printed

patheard:
# Summary
Update to have the logged stack printed as a string with newlines
rather than a list of strings.

# Related
* #1514


sastels (approved):
Are you just trying to increase your github score?

PR #1523, opened 2022-04-19
Migration: add forced reset template

jzbahrai:
# Summary | Résumé

Add template for a Forced Reset Password email.

### DIFF
--- a/migrations/versions/0419_add_forced_pass_template.py
+++ b/migrations/versions/0419_add_forced_pass_template.py
@@ -24,12 +24,17 @@ service_id = "d6aa2c68-a2d9-4437-ab19-3ae8eb202553"
 
 
 def upgrade():
-    password = hashpw(str(uuid.uuid4()))
     op.get_bind()
     template_insert = """INSERT INTO templates (id, name, template_type, created_at,
                                                 content, archived, service_id, subject, created_by_id, version, process_type, hidden)
                                  VALUES ('{}', '{}', '{}', '{}', '{}', False, '{}', '{}', '{}', 1, '{}', False)
                               """
+
+    template_history_insert = """INSERT INTO templates_history (id, name, template_type, created_at,
+                                                                content, archived, service_id,
+                                                                subject, created_by_id, version, process_type, hidden)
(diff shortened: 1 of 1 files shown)


migrations/versions/0419_add_forced_pass_template.py
```diff
@@ -0,0 +1,65 @@
...
+revision = "0419_add_forced_pass_template"
+down_revision = "0327_add_password_expired"
+
+
+user_id = "6af522d0-2915-4e52-83a3-3690455a5fe6"
+service_id = "d6aa2c68-a2d9-4437-ab19-3ae8eb202553"
+
+
+def upgrade():
+    password = hashpw(str(uuid.uuid4()))
```
sastels, line 27: What is this used for?

jzbahrai: this was a mistake

jzbahrai: i added the template history table as well if you can review. 🙇🏽‍♀️

migrations/versions/0419_add_forced_pass_template.py
```diff
@@ -0,0 +1,83 @@
...
+        Pour réinitialiser votre mot de passe, veuillez cliquer sur le lien suivant :
+        [Réinitialisation de votre mot de passe](((url))?lang=fr)
+        Ce lien est unique. Ne le transmettez à personne. 
+        Si vous navez pas demandé ce courriel, veuillez [nous contacter](https://notification.canada.ca/contact?lang=fr).
+        """
+
+    op.execute(
+        template_history_insert.format(
+            "e9a65a6b-497b-42f2-8f43-1736e43e13b3",
+            "Notify email verification code",
```
jimleroyer, line 56: Should it be the same name as the one in the templates table?
```suggestion
            "Notify forced-password reset email",
```

jzbahrai: will change

jimleroyer (approved):
LGTM

PR #1524, opened 2022-04-19
fix(loginEvent): add loginEvent creation on password reset

andrewleith:
# Summary | Résumé

Create a `LoginEvent` when a user is forced to update their password, or requests a password reset, as this process automatically logs the user in.  The user will see this even on the Sign-in History page when they log in.

See cds-snc/notification-planning#615 for more details including test plan.


app/user/rest.py
```diff
@@ -616,6 +623,11 @@ def update_password(user_id):
         raise InvalidRequest(errors, status_code=400)
 
     update_user_password(user, pwd)
+
+    # save login event
+    if login_data is not None and login_data != {}:
```
jimleroyer, line 628: What about `if login_data` style? or you prefer the more explicit version?

jimleroyer (commented):
Looks good to me, I'd just recommend we add a test that checks if `save_login_event` was called so that we have one more regression in case that something changes in the future.

app/user/rest.py
andrewleith: Nah, I'm just a python newb :D - I'll fix it up!

andrewleith:
> Looks good to me, I'd just recommend we add a test that checks if `save_login_event` was called so that we have one more regression in case that something changes in the future.

@jimleroyer I added some tests.  I tried to patch the `save_login_event` method to see if it was called but I couldn't wrangle pytest into working, so I followed another pattern in some other tests.  Let me know if this is sufficient.

jzbahrai (approved):

PR #1527, opened 2022-04-20
Add redacted template

jzbahrai:
# Summary | Résumé

We need a row for every template in the `template_redacted` table. This adds it for the forced password email template

### DIFF
--- a/migrations/versions/0420_add_redacted_template.py
+++ b/migrations/versions/0420_add_redacted_template.py
@@ -1,0 +1,41 @@
+"""empty message
+
+Revision ID: 0420_add_redacted_template
+Revises: 0419_add_forced_pass_template
+Create Date: 2022-04-19 13:00:00
+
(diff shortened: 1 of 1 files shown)


migrations/versions/0420_add_redacted_template.py
```diff
@@ -0,0 +1,41 @@
...
+# revision identifiers, used by Alembic.
+from datetime import datetime
+
+from alembic import op
+
+revision = "0420_add_redacted_template"
+down_revision = "0419_add_forced_pass_template"
+
+user_id = "6af522d0-2915-4e52-83a3-3690455a5fe6"
+service_id = "d6aa2c68-a2d9-4437-ab19-3ae8eb202553"
```
sastels, line 20: don't need service_id, right?

sastels (approved):
works locally!

PR #1531, opened 2022-04-26
fix: multi-value param support for the Lambda API

patheard:
# Summary
Switch to the [`apig_wsgi` module](https://github.com/adamchainz/apig-wsgi), which provides support for multiple
GET parameters using the same name.  

This fixes the error we were seeing retrieving recent notifications for a 
service where the `status` URL parameter is passed more than once 
in the request.

# Related
* cds-snc/notification-planning#635

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -93,6 +93,3 @@ ignore_missing_imports = True
 
 [mypy-flask_sqlalchemy.*]
 ignore_missing_imports = True
-
-[mypy-awsgi.*]
-ignore_missing_imports = True
--- a/requirements-app.txt
+++ b/requirements-app.txt
@@ -1,7 +1,7 @@
 # Run `make freeze-requirements` to update requirements.txt
 # with package version changes made in requirements-app.txt
 
-aws-wsgi==0.2.7
+apig-wsgi==2.13.0
 boto==2.49.0
 cffi==1.14.5
 celery[sqs]==5.0.5
--- a/application.py
+++ b/application.py
@@ -3,9 +3,9 @@ from __future__ import print_function
 
 import os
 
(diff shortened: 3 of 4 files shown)


mypy.ini
```diff
@@ -93,6 +93,3 @@ ignore_missing_imports = True
 
 [mypy-flask_sqlalchemy.*]
 ignore_missing_imports = True
-
-[mypy-awsgi.*]
-ignore_missing_imports = True
```
jimleroyer, line 98: We might have to add the same for apig-wsgi eventually. Not blocking this review.

requirements-app.txt
```diff
@@ -1,7 +1,7 @@
 # Run `make freeze-requirements` to update requirements.txt
 # with package version changes made in requirements-app.txt
 
-aws-wsgi==0.2.7
+apig-wsgi==2.13.0
```
jimleroyer, line 4: This is a fairly big change to switch library. It seems more up to date though compared to `aws-wsgi`. Should we have plans for testing or you are confident we can deploy this to production?

jimleroyer (commented):
Left a few comments but that looks good.

requirements-app.txt
patheard: Confident, but I'd still like to give things a thorough test in Staging when this merges.  

The problem with the `aws-wsgi` module is that it was no longer supported and missing a lot of features that AWS has added for Lambda proxy events (like the multi-value params and support for `v2.0` of the event payload).

I found `apig-wsgi` yesterday while trying to solve problems with creating the Notify admin PR review env.

jimleroyer (approved):
:shipit:

PR #1534, opened 2022-05-03
Add logging for template ID

jzbahrai:
# Summary | Résumé

Ran this locally:
```
2022-05-03T18:19:53 api api INFO None "Trying to send notification for Template ID: 41e5b0bb-3d4f-42b0-ad6d-0be9c4ac8a77" [in /workspace/app/v2/notifications/post_notifications.py:202]
```

We are doing this so we can tell users which Templates are causing API key failures.

### DIFF
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -199,6 +199,8 @@ def post_notification(notification_type):
         notification_type,
     )
 
+    current_app.logger.info(f"Trying to send notification for Template ID: {template.id}")
+
     reply_to = get_reply_to_text(notification_type, form, template)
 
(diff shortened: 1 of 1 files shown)


andrewleith (approved):
LGTM :shipit:

sastels:
Just wondering: is this just temporary? Or something that in general will allow us to trace back errors.
Would be nice to add details when the error is thrown - looking at the code, it might involve a bit of refactoring though (sending the notification around, basically).

jimleroyer:
+1 on @sastels comment. Associating the notification ID would be useful to make it easy too. A comment to say to remove this in the future if that is temporary.

Good change though to target the "leaky" API key ID. We can improve that log in the future if that is useful.

PR #1545, opened 2022-05-11
feat: disconnect zendesk support

mohamed-cds:
CDS Platform is centralizing around Freshdesk and will no longer be pursuing Zendesk support use cases.

For more details see: https://gcdigital.slack.com/files/T2G2S06PM/F03EZH2GBC3


jimleroyer:
Do we plan to still use ZenDesk CRM capabilities now that we are turning away from the support functionality?

mohamed-cds:
> Do we plan to still use ZenDesk CRM capabilities now that we are turning away from the support functionality?

@jimleroyer No plans to move away from Zendesk Sell atm

PR #1548, opened 2022-05-11
Fix acknowledge notification code

jzbahrai:
# Summary | Résumé

After we send the notification, we need to acknowledge the notification is sent in Redis.
This PR addresses that workflow of acknowledging notifications that are already sent.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -1,8 +1,9 @@
 import json
 from collections import defaultdict, namedtuple
 from datetime import datetime
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Union
 from uuid import UUID
+from celery import uuid
 
 from flask import current_app
 from more_itertools import chunked
@@ -163,7 +164,7 @@ def job_complete(job: Job, resumed=False, start=None):
         )
 
 
-def choose_database_queue(template: Template, service: Service):
+def choose_database_queue(template: Any, service: Service):
     if Config.FF_PRIORITY_LANES:
         if service.research_mode:
             return QueueNames.RESEARCH_MODE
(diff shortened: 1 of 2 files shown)


app/celery/tasks.py
```diff
@@ -1000,3 +1011,37 @@ def get_recipient_csv(job: Job, template: Template) -> RecipientCSV:
         placeholders=template.placeholders,
         max_rows=get_csv_max_rows(job.service_id),
     )
+
+
+def _acknowledge_notification(notification_type, template, receipt):
```
jimleroyer, line 1016: Can we please add typing to this signature?

app/celery/tasks.py
```diff
@@ -1000,3 +1011,37 @@ def get_recipient_csv(job: Job, template: Template) -> RecipientCSV:
...
+    """
+    Acknowledge the notification has been saved to the DB and sent to the service.
+
+    Args:
+    notification_type: str
+        Type of notification being sent; either SMS_TYPE or EMAIL_TYPE
+    template: model.Template
+        Template used to send notification
+
+    Returns: None
```
jimleroyer, line 1026: Great to have documentation! 👍

app/celery/tasks.py
```diff
@@ -1000,3 +1011,37 @@ def get_recipient_csv(job: Job, template: Template) -> RecipientCSV:
...
+    elif notification_type == EMAIL_TYPE:
+        if Config.FF_PRIORITY_LANES:
+            if template.process_type == PRIORITY:
+                email_priority.acknowledge(receipt)
+            elif template.process_type == NORMAL:
+                email_normal.acknowledge(receipt)
+            elif template.process_type == BULK:
+                email_bulk.acknowledge(receipt)
+        else:
+            email_queue.acknowledge(receipt)
```
jimleroyer, line 1047: This feels like many conditions, it might be simplified with a dictionary of priority as key and functions as value. That would emulate proper function dispatch... which then someone might say just use an OO pattern, but I don't see which objects we could reuse around for that.

jimleroyer (approved):

app/celery/tasks.py
jzbahrai: so I added typing to this signature, but 
```"_acknowledge_notification" has incompatible type "Union[Template, TemplateHistory, Tuple[Union[Template, TemplateHistory], Dict[Any, Any]]]"; expected "Template"```

I think this is because of the way template can be retrieved (for cache or the db)
but typing doesn't seem to be aware of the diff :/ 
I added the type check as Any :/

PR #1554, opened 2022-05-25
Add try/except to find issues in signing

jzbahrai:
# Summary | Résumé

Add logging to check if we can't unsign the data correctly.


sastels (approved):
🚀

PR #1561, opened 2022-06-09
Added personalisation var limit

jimleroyer:
# Summary | Résumé

## Size limit on the overall template variables size

Added a size limit on the overall template variables size in order to avoid notifications stuck in SQS queues, which has a limit of 256k. This can be configured via the `PERSONALISATION_SIZE_LIMIT` environment variable. The default is set to `50kb`. The limit was determined by looking at our historic data of the past week and I found the max personalisation data we got was around 17kb.

This could be even too low considering that a batch of 10 notifications can be queued in SQS and if each of these have a size of 40kb, then only 7 are necessary to overflow the SQS limit.

## Limit on the number of attachments

Added a limit on the number of attachments a notification can contain. This can be configured via the `ATTACHMENT_NUM_LIMIT` environment variable. The default is set to `10`.

* Added new automated tests that test the conditional edges on number of attachments and variable overall size.

Let's check in the history of notifications of the past week...

* ...how many notifications contained more than 10 attachments.
* ...how many notifications got its personalisation data past the 50kb size.

      fonctionnalité existante?

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -214,6 +214,7 @@ class Config(object):
     SQLALCHEMY_POOL_RECYCLE = 300
     SQLALCHEMY_ECHO = env.bool("SQLALCHEMY_ECHO", False)
     PAGE_SIZE = 50
+    PERSONALISATION_SIZE_LIMIT = env.int("PERSONALISATION_SIZE_LIMIT", 1024 * 120)  # 120k bytes limit by default
     API_PAGE_SIZE = 250
     TEST_MESSAGE_FILENAME = "Test message"
     ONE_OFF_MESSAGE_FILENAME = "Report"
--- a/app/config.py
+++ b/app/config.py
@@ -218,8 +218,8 @@ class Config(object):
     SQLALCHEMY_ECHO = env.bool("SQLALCHEMY_ECHO", False)
     PAGE_SIZE = 50
     PERSONALISATION_SIZE_LIMIT = env.int(
-        "PERSONALISATION_SIZE_LIMIT", 1024 * 120
-    )  # 120k bytes limit by default for personalisation data per notification
+        "PERSONALISATION_SIZE_LIMIT", 1024 * 50
+    )  # 50k bytes limit by default for personalisation data per notification
     API_PAGE_SIZE = 250
     TEST_MESSAGE_FILENAME = "Test message"
     ONE_OFF_MESSAGE_FILENAME = "Report"
--- a/app/config.py
+++ b/app/config.py
@@ -195,6 +195,7 @@ class Config(object):
 
     NOTIFY_ENVIRONMENT = os.getenv("NOTIFY_ENVIRONMENT", "development")
     ADMIN_CLIENT_USER_NAME = "notify-admin"
+    ATTACHMENT_NUM_LIMIT = env.int("ATTACHMENT_NUM_LIMIT", 10)  # Limit of 10 attachments per notification.
     ATTACHMENT_SIZE_LIMIT = env.int(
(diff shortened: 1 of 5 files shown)


app/config.py
```diff
@@ -216,6 +217,9 @@ class Config(object):
     SQLALCHEMY_POOL_RECYCLE = 300
     SQLALCHEMY_ECHO = env.bool("SQLALCHEMY_ECHO", False)
     PAGE_SIZE = 50
+    PERSONALISATION_SIZE_LIMIT = env.int(
+        "PERSONALISATION_SIZE_LIMIT", 1024 * 120
```
andrewleith, line 221: You mention 50kb in the PR description but its set to 120kb here.

app/notifications/validators.py
```diff
@@ -274,17 +275,62 @@ def check_service_letter_contact_id(service_id, letter_contact_id, notification_
...
+    errors.extend(errors_vars)
+    errors.extend(errors_num_file)
+    errors.extend(errors_files)
+    return json_personalisation, errors
+
+
+def validate_personalisation(json_personalisation):
+    errors = []
+    values = [v for _, v in json_personalisation.items() if not isinstance(v, dict)]
+    concat_values = functools.reduce(lambda v1, v2: v1 + v2, values, "")
```
sastels, line 292: these values are almost certainly strings, I think... just wondering if we should make sure, like returning `f"{v1}{v2}"` or something.

app/notifications/validators.py
```diff
@@ -274,17 +275,62 @@ def check_service_letter_contact_id(service_id, letter_contact_id, notification_
...
+    json_personalisation, errors_vars = validate_personalisation(json_personalisation)
+    json_personalisation, errors_num_file = validate_personalisation_num_files(json_personalisation)
+    json_personalisation, errors_files = decode_personalisation_files(json_personalisation)
+    errors.extend(errors_vars)
+    errors.extend(errors_num_file)
+    errors.extend(errors_files)
+    return json_personalisation, errors
+
+
+def validate_personalisation(json_personalisation):
```
sastels, line 289: could make the function name more specific, ie `validate_personalisation_size()` perhaps, similar to how you did for `validate_personalisation_num_files()`

sastels:
> looks good to me! but should there be a manifest repo change as well

If we're good with just using the defaults for now then we don't need to add the env vars  ¯\_(ツ)_/¯

app/config.py
jimleroyer: Yeah it's true! I didn't push that change yet. I want to meet with the Forms team before setting on a definitive value (meeting set on Monday), as they make use of this one-var template pattern. I could get their feedback before moving on. 

I will update the PR to 120k though in the meantime.

app/notifications/validators.py
jimleroyer: Yeah these are strings, string interpolation could work (in the lambda), but I'd still need to reduce the values.

app/notifications/validators.py
jimleroyer: Oh yes, I meant to do that. Thanks!

jimleroyer:
> > looks good to me! but should there be a manifest repo change as well

> If we're good with just using the defaults for now then we don't need to add the env vars ¯\_(ツ)_/¯

Yeah that was my thinking for now. I can certainly introduce these though as it would be faster to react if we have an incident and want to change these quickly through manifests. So, not a blocker, but on my list let's say?

app/notifications/validators.py
sastels: 💯  Was thinking along the lines of
```
    concat_values = functools.reduce(lambda v1, v2: f"{v1}{v2}", values, "")
```

PR #1565, opened 2022-06-21
Upgrade Pytest to 5.3.2

jzbahrai:
# Summary | Résumé

Finish upgrading pytest to 5.3.2 (and others required)
Started the process in https://github.com/cds-snc/notification-api/pull/1564

Fixes largely copied from https://github.com/alphagov/notifications-api/pull/3228

Smoke test (in particular admin), specifically around the code this PR touched:
- `ServiceUser` and `ServicePermission`:
  - add a user to a service
  - change the user's permissions
  - remove the user from the service
- archiving  templates:
  - Make a template in a folder
  - See it in the `template_folder_map` table
  - Delete the template
  - Ensure that in the `templates` table it has `archived = True`
  - Ensure that it is no longer in the `template_folder_map` table
- query change in `delete_notifications_older_than_retention_by_type` and change to `insert_update_notification_history`:
  - ? these are harder to test. basically we want to ensure that the whole retention period still works. We should set it to 1 or 2 days for a particular service in staging and check that notifications have been moved to the history table after that time.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -20,7 +20,7 @@ from sqlalchemy import asc, desc, func
 from sqlalchemy.dialects.postgresql import insert
 from sqlalchemy.orm import joinedload
 from sqlalchemy.orm.exc import NoResultFound
-from sqlalchemy.sql import functions
+from sqlalchemy.sql import functions, text
 from sqlalchemy.sql.expression import case
 from werkzeug.datastructures import MultiDict
 
@@ -376,7 +376,7 @@ def _delete_for_query(subquery):
 
 
 def insert_update_notification_history(notification_type, date_to_delete_from, service_id):
-    notifications = db.session.query(*[x.name for x in NotificationHistory.__table__.c]).filter(
+    notifications = db.session.query(*[text(x.name) for x in NotificationHistory.__table__.c]).filter(
         Notification.notification_type == notification_type,
         Notification.service_id == service_id,
         Notification.created_at < date_to_delete_from,
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -20,7 +20,7 @@ from sqlalchemy import asc, desc, func
 from sqlalchemy.dialects.postgresql import insert
 from sqlalchemy.orm import joinedload
 from sqlalchemy.orm.exc import NoResultFound
-from sqlalchemy.sql import functions, text
+from sqlalchemy.sql import functions, literal_column
 from sqlalchemy.sql.expression import case
 from werkzeug.datastructures import MultiDict
 
@@ -319,8 +319,7 @@ def delete_notifications_older_than_retention_by_type(notification_type, qry_lim
         days=7
     )
     services_with_data_retention = [x.service_id for x in flexible_data_retention]
-    # service_ids_to_purge = Service.query.filter(Service.id.notin_(services_with_data_retention)).all()
-    service_ids_to_purge = Service.query.filter(Service.id.notin_(services_with_data_retention)).all()
+    service_ids_to_purge = db.session.query(Service.id).filter(Service.id.notin_(services_with_data_retention)).all()
 
     for service_id in service_ids_to_purge:
         if notification_type == LETTER_TYPE:
@@ -377,7 +376,7 @@ def _delete_for_query(subquery):
 
 
 def insert_update_notification_history(notification_type, date_to_delete_from, service_id):
-    notifications = db.session.query(*[text(x.name) for x in NotificationHistory.__table__.c]).filter(
+    notifications = db.session.query(*[literal_column(x.name) for x in NotificationHistory.__table__.c]).filter(
         Notification.notification_type == notification_type,
         Notification.service_id == service_id,
         Notification.created_at < date_to_delete_from,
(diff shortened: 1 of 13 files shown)


app/dao/notifications_dao.py
```diff
@@ -319,13 +319,13 @@ def delete_notifications_older_than_retention_by_type(notification_type, qry_lim
         days=7
     )
     services_with_data_retention = [x.service_id for x in flexible_data_retention]
-    service_ids_to_purge = db.session.query(Service.id).filter(Service.id.notin_(services_with_data_retention)).all()
+    service_ids_to_purge = Service.query.filter(Service.id.notin_(services_with_data_retention)).all()
```
jimleroyer, line 322: Hmm newer version of SQLAlchemy does not allow to query one field uniquely, instead of loading the whole object? Just wondering as the data payload will be high on the network, albeit not affecting performance assuming. Mostly curious on this change.

requirements.txt
```diff
@@ -28,7 +28,7 @@ psycopg2-binary==2.8.6
 PyJWT==2.1.0
 pytz==2021.1
 PyYAML==5.4.1
-SQLAlchemy==1.3.23
+SQLAlchemy==1.4.37
```
jimleroyer, line 31: Well I guess it's good to bump it up anyway, I am surprised to see it upgraded though, does SQLAlchemy require a certain version of pytest? 🤔

app/models.py
```diff
@@ -199,8 +199,6 @@ class ServiceUser(BaseModel):
 
     __table_args__ = (UniqueConstraint("user_id", "service_id", name="uix_user_to_service"),)
 
-    user = db.relationship("User")
```
jimleroyer, line 202: Uncessary now with the SQLAlchemy upgrade? or it really removes the relationship, which then I wonder why the class is still named `ServiceUser`? 🤔

app/dao/templates_dao.py
```diff
@@ -41,9 +41,6 @@ def dao_create_template(template):
 @transactional
 @version_class(VersionOptions(Template, history_class=TemplateHistory))
 def dao_update_template(template):
-    if template.archived:
-        template.folder = None
```
jimleroyer, line 45: None by default with recent SQLAlchemy upgrade?

app/template/rest.py
```diff
@@ -126,6 +126,8 @@ def update_template(service_id, template_id):
         raise InvalidRequest(errors, status_code=400)
 
     update_dict = template_schema.load(updated_template).data
+    if update_dict.archived:
+        update_dict.folder = None
```
jimleroyer, line 130: Is this related to https://github.com/cds-snc/notification-api/pull/1565/files#diff-d3ce135e4e0dcab49ab03e15c65040fcfa1fec9dc64722849014c1f46fdc83feL44 ?

app/user/rest.py
```diff
@@ -509,7 +509,7 @@ def set_permissions(user_id, service_id):
     # TODO fix security hole, how do we verify that the user
     # who is making this request has permission to make the request.
     service_user = dao_get_service_user(user_id, service_id)
-    user = service_user.user
+    user = get_user_by_id(user_id)
```
jimleroyer, line 512: Puzzling change here too, if we did remove the relationship to user from ServiceUser and then we make a separate db select to the database to get the user... why not dump dao_get_service_user? and use `dao_fetch_service_by_id`? 🤔

jimleroyer (commented):
Many puzzling changes, which I don't get. I know you copied GDS, I wonder why they did that. Were the comments on their PR explained it a bit?

requirements.txt
jzbahrai: yup, old version of sqlalchemy doesnt work with newer version of pytest.

tests/app/dao/test_templates_dao.py
```diff
@@ -99,35 +99,6 @@ def test_update_template(sample_service, sample_user):
     assert dao_get_all_templates_for_service(sample_service.id)[0].name == "new name"
 
 
-def test_update_template_in_a_folder_to_archived(sample_service, sample_user):
```
andrewleith, line 102: Is archiving a folder no longer a feature?

requirements_for_test.txt
```diff
@@ -3,12 +3,12 @@ flake8==3.8.4
 isort==5.6.4
 moto==1.3.14
 idna==2.8
-pytest==3.10.1  # pyup: <4
+pytest==5.3.2  # pyup: <4
```
andrewleith, line 6: Should we remove this pyup comment?

app/models.py
sastels: Note that the table still has the user_id in there. 
This change removes the warning
https://docs.sqlalchemy.org/en/14/errors.html#error-qzyx
We could explore the use of `back_populates` but since it's only used once in the code it's easier to just eliminate this line

app/template/rest.py
sastels: 👍

app/dao/notifications_dao.py
sastels: The main point of this change is to get back a list of services rather than service ids (which I just realized means the variable name should be renamed 😅 ) Otherwise the call to ` insert_update_notification_history()` below was causing an error `psycopg2.ProgrammingError: can't adapt type 'Row'`

jimleroyer: It seems only IDs are used though in the service objects, right? Are there other service fields used somewhere else?

jzbahrai: Jimmy to look at this.

jimleroyer (approved):
Let's :shipit:

sastels (approved):
👍

PR #1569, opened 2022-06-30
Fixed JSON serialization error with sqlalchemy.engine.row.Row

jimleroyer:
# Summary | Résumé

Added custom JSON encoder that adds support for `sqlalchemy.engine.row.Row`.

> A JSON encoder that adds edge case support for the Notify Python stack.
>
> Namely, these are currently supported:
>
> 1. Added support for the sqlalchemy.engine.row.Row data type. When we upgraded to version 4, a few JSON serialization started to fail as the library now returns a Row object on the session.query returns. This encoder adds support to convert it to a dict, which the json package supports by default.

* Added a test around the `get_delivered_notification_stats_by_month_data` REST endpoint which showed failure.
* **TODO**: Add a test for the new `NotifyJSONEncoder`

None.

---

> mentionné.

      fonctionnalité existante?


app/schemas.py
```diff
@@ -21,7 +21,7 @@
     validate_phone_number,
 )
 
-from app import db, ma, models
+from app import db, marshmallow, models
```
andrewleith, line 24: +1 on the explicit naming

app/__init__.py
```diff
@@ -78,10 +79,11 @@ def create_app(application, config=None):
 
     application.config["NOTIFY_APP_NAME"] = application.name
     init_app(application)
+    application.json_encoder = NotifyJSONEncoder
```
andrewleith, line 82: It wasn't immediately clear to me how this all works, but after review I think that this sets a custom json encoder in flask which is then called whenever we attempt to encode JSON using `flask.jsonify()`?

jimleroyer: Yes that's it. Flask offers its own abstraction that will make use of the regular Python `json` module, but we need to short-circuit it by specifying our own which simply overrides the existing ones, enriching it with the custom types we're interested in.

To add to the confusion, Notify does not always call on `jsonify` but makes use of marshmallow sometimes. It does not have a consistent JSON serialization strategy, depending if the object comes from the model or not.

PR #1570, opened 2022-06-30
Task/fix taskstests

jzbahrai:
# Summary | Résumé

Note: Plural save funcs - save_emails, save_smss (this is the code path we use post feature flag turn on).

Changed test_tasks.py and tasks.py. Main changes:

1. Removed `process_row` and changed all references to `process_rows`
2. Removed `save_sms` and `save_email`. Converted those code paths to `save_emails` and `save_smss`. 
3. Changed tests to accommodate the above two
4. Updated `incomplete_job` code to take `process_rows` instead of `process_row`
5. Changed the `handler` code to use the plural save tasks instead of the old deprecated ones
6. Marked all the `letter` code for deprecation
7.  Changed how we retrieve templates as there was an error in the code

# Testing

1. Updated tests, moved tests around into different classes
2. Moved all tests to use plural save func
3. Updated tests for incomplete jobs
4. Updated handler code

I tested handler code in two ways:  I added unit tests for it, and then I tested it by throwing an error in the save func. The second test showed me `max_retries` is never changed, it remains the same regardless of what "retry" number the task is on. Unfortunately I couldn't find a way to retrieve what "retry" number the task was on (there isn't one according to celery docs) but through logging, I did see that the tasks were tried a certain number of times as well as tasks that should complete, completed.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -814,7 +814,7 @@ def process_incomplete_job(job_id):
     template.process_type = db_template.process_type
 
     csv = get_recipient_csv(job, template)
-    rows = csv.get_rows()
+    rows = csv.get_rows() # This returns an iterator
     first_row = []
     for row in rows:
         if row.index > resume_from_row:
@@ -826,7 +826,7 @@ def process_incomplete_job(job_id):
                     metrics_logger, 1, notification_type=db_template.template_type, priority=db_template.process_type
                 )
                 first_row = []
-            exit
+            break
 
     job_complete(job, resumed=True)
 
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -526,6 +526,7 @@ def update_letter_notifications_to_error(self, notification_references):
     raise NotificationTechnicalFailureException(message)
 
 
+# TODO: Deprecated: Old handler Code
 def handle_save_error(task, notification, notification_id, exception):
     # Sometimes, SQS plays the same message twice. We should be able to catch an IntegrityError, but it seems
     # SQLAlchemy is throwing a FlushError. So we check if the notification id already exists then do not
@@ -581,15 +582,16 @@ def handle_batch_error_and_forward(
             # in the form of a dict
             if isinstance(template, tuple):
                 template = template[0]
-            retry_msg = "{task} notification for job {job} row number {row} and notification id {notif}".format(
+            retry_msg = "{task} notification for job {job} row number {row} and notification id {notif} and max_retries are {max_retry}".format(
                 task=task.__name__,
                 job=notification.get("job", None),
                 row=notification.get("row_number", None),
                 notif=notification_id,
+                max_retry=task.max_retries,
             )
-            current_app.logger.exception("Retry" + retry_msg)
+            current_app.logger.info("Retry" + retry_msg)
             try:
-                if task.max_retries == 5:
+                if task.max_retries == 5 and len(signed_and_verified) != 1:
                     save_fn.apply_async(
                         (service.id, [signed], None),
                         queue=choose_database_queue(template, service),
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -589,14 +589,16 @@ def handle_batch_error_and_forward(
                 notif=notification_id,
                 max_retry=task.max_retries,
             )
-            current_app.logger.info("Retry" + retry_msg)
+            current_app.logger.warning("Retry " + retry_msg)
             try:
                 if task.max_retries == 5 and len(signed_and_verified) != 1:
                     save_fn.apply_async(
                         (service.id, [signed], None),
                         queue=choose_database_queue(template, service),
                     )
+                    current_app.logger.warning("Made a new task to retry")
                 else:
+                    current_app.logger.warning("Retrying the current task")
                     task.retry(queue=QueueNames.RETRY, exc=exception)
             except task.MaxRetriesExceededError:
                 current_app.logger.error("Max retry failed" + retry_msg)
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -7,7 +7,6 @@ from uuid import UUID
 from flask import current_app
 from itsdangerous import BadSignature
 from more_itertools import chunked
-from notifications_utils.columns import Row
 from notifications_utils.recipients import RecipientCSV
 from notifications_utils.statsd_decorators import statsd
 from notifications_utils.template import SMSMessageTemplate, WithSubjectTemplate
@@ -93,7 +92,6 @@ from app.notifications.process_notifications import (
     send_notification_to_queue,
 )
 from app.notifications.validators import check_service_over_daily_message_limit
-from app.service.utils import service_allowed_to_send_to
 from app.utils import get_csv_max_rows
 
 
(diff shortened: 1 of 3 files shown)


app/celery/tasks.py
```diff
@@ -171,60 +169,14 @@ def choose_database_queue(template: Any, service: Service):
         return QueueNames.NORMAL_DATABASE
 
 
-def process_row(row: Row, template: Template, job: Job, service: Service):
```
jzbahrai, line 174: this code path isnt used anymore

app/dao/templates_dao.py
```diff
@@ -109,8 +109,7 @@ def dao_get_template_by_id(
                 return TemplateHistory.from_json(template_cache_decoded), template_cache_decoded
             else:
                 return Template.from_json(template_cache_decoded), template_cache_decoded
-
-    elif version is not None:
+    if version is not None:
```
jzbahrai, line 112: this should be an elif as if the template isn't returned from cache, it might still be in TemplateHistory

sastels: good catch!

sastels:
Smoke tests pass when running the branch locally.

app/celery/tasks.py
```diff
@@ -171,60 +169,14 @@ def choose_database_queue(template: Any, service: Service):
...
-                signed,
-            ),
-            task_kwargs,
-            queue=choose_database_queue(template, service),
-        )
-    else:
-        current_app.logger.debug("SMS {} failed as restricted service".format(notification_id))
-
-
 def process_rows(rows: List, template: Template, job: Job, service: Service):
```
sastels, line 172: could use `List[Row]` to be more specific.

app/celery/tasks.py
```diff
@@ -93,7 +92,6 @@
     send_notification_to_queue,
 )
 from app.notifications.validators import check_service_over_daily_message_limit
-from app.service.utils import service_allowed_to_send_to
```
sastels, line 96: Are we still checking this somewhere in the flow? Ie making sure that a trial service isn't sending to recipients it's not allowed to.

app/celery/tasks.py
```diff
@@ -740,11 +582,26 @@ def handle_batch_error_and_forward(
...
+            retry_msg = "{task} notification for job {job} row number {row} and notification id {notif} and max_retries are {max_retry}".format(
+                task=task.__name__,
+                job=notification.get("job", None),
+                row=notification.get("row_number", None),
+                notif=notification_id,
+                max_retry=task.max_retries,
             )
+            current_app.logger.warning("Retry " + retry_msg)
+            try:
+                if task.max_retries == 5 and len(signed_and_verified) != 1:
```
jimleroyer, line 594: Can you please add a comment on the reason why we want to catch the max retries at 5?

Did you test that `task.max_retries` decrease on every retry iteration?

app/celery/tasks.py
jzbahrai: yup we moved this https://github.com/cds-snc/notification-api/pull/1502/files

app/celery/tasks.py
jzbahrai: sorry forgot about the comment, will add

for the second - max_retries doesn't reduce on each retry iteration, it is stored elsewhere and isn't exposed to the task. I tested this manually and saw the max_retry stay the same but the number of times the retry was called was 5.

app/celery/tasks.py
jzbahrai: ah but in the case of the test its notification. going to leave this as is

app/celery/tasks.py
```diff
@@ -952,9 +812,19 @@ def process_incomplete_job(job_id):
...
-            process_row(row, template, job, job.service)
+            first_row.append(row)
+            for result in chunked(rows, Config.BATCH_INSERTION_CHUNK_SIZE):
+                result = first_row + result
+                process_rows(result, template, job, job.service)
+                put_batch_saving_bulk_created(
+                    metrics_logger, 1, notification_type=db_template.template_type, priority=db_template.process_type
+                )
+                first_row = []
+            exit
```
jimleroyer, line 827: exit?

app/celery/tasks.py
jzbahrai: but ah this is a good point cause now I can change the if statement. ok let me think about it

jzbahrai: added a comment and reran the tests.

tests/app/celery/test_tasks.py
```diff
@@ -849,26 +780,24 @@ def test_should_cancel_job_if_service_is_inactive(self, sample_service, sample_j
...
 
         job = jobs_dao.dao_get_job_by_id(sample_job.id)
         assert job.job_status == "cancelled"
         s3.get_job_from_s3.assert_not_called()
-        tasks.process_row.assert_not_called()
+        tasks.process_rows.assert_not_called()
 
 
-class TestProcessRow:
+class TestProcessRows:
```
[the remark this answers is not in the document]
jzbahrai: changed!

app/celery/tasks.py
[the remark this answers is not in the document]
jimleroyer: +1

tests/app/celery/test_tasks.py
```diff
@@ -970,79 +889,28 @@ def test_should_not_save_sms_if_restricted_service_and_invalid_number(self, noti
...
-                template=SMSMessageTemplate(template.__dict__),
-            ),
-            template,
-            job,
-            service,
-        )
-
-        assert not save_sms_mock.called
-
-    def test_should_not_save_sms_if_team_key_and_recipient_not_in_team(self, notify_db_session, mocker):
```
jimleroyer, line 991: Do we have another test to replace this one?

tests/app/celery/test_tasks.py
```diff
@@ -970,79 +889,28 @@ def test_should_not_save_sms_if_restricted_service_and_invalid_number(self, noti
...
+                )
+            ],
             template,
             job,
             service,
         )
 
         assert not save_sms_mock.called
 
-    def test_should_not_save_email_if_restricted_service_and_invalid_email_address(self, notify_db_session, mocker):
```
jimleroyer, line 1019: Do we have another test to replace this one?

tests/app/celery/test_tasks.py
```diff
@@ -1355,23 +1191,128 @@ def test_save_sms_uses_non_default_sms_sender_reply_to_text_if_provided(self, mo
...
-            sender_id=new_sender.id,
-        )
-
+        save_smss(service.id, [signer.sign(notification)], notification_id)
         persisted_notification = Notification.query.one()
         assert persisted_notification.reply_to_text == "new-sender"
 
 
-class TestSaveEmail:
+class TestHandler:
```
jimleroyer, line 1203: Can we please rename to TestSaveErrorHandling or something similar? it might not be clear which handler we are referring too otherwise.

tests/app/celery/test_tasks.py
```diff
@@ -1880,15 +1785,16 @@ def test_process_incomplete_job_sms_FF_PRIORITY_LANES_true(self, mocker, sample_
 
         assert completed_job.job_status == JOB_STATUS_FINISHED
 
-        assert save_sms.call_count == 8  # There are 10 in the file and we've added two already
+        assert save_smss.call_count == 1  # The save_smss call will be called once
+        assert len(save_smss.call_args[0][0][1]) == 8  # The unprocessed 8 notifications will be sent to save_smss
```
jimleroyer, line 1789: Good comments

jimleroyer (commented):
Lots of code cleanup 👍 

I am worried about two piece of logic though: the loop within a loop containing the `exit` (did we mean to put `break` instead?) and the retry logic.

app/celery/tasks.py
```diff
@@ -740,11 +582,28 @@ def handle_batch_error_and_forward(
...
+                job=notification.get("job", None),
+                row=notification.get("row_number", None),
+                notif=notification_id,
+                max_retry=task.max_retries,
             )
+            current_app.logger.warning("Retry " + retry_msg)
+            try:
+                # If >1 notification has failed, we want to make individual
+                # tasks to retry those notifications.
+                if len(signed_and_verified) != 1:
```
jimleroyer, line 596: Is that just me, I might be paranoid, I'd feel safer with

```suggestion
                if len(signed_and_verified) > 1:
```

😅

tests/app/celery/test_tasks.py
jzbahrai: same here: https://github.com/cds-snc/notification-api/pull/1502/files

app/celery/tasks.py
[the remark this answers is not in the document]
jzbahrai: ok was going to meet Steve irl but I will try explaining this here. If it doesn't make sense let us meet irl.

l#817. = rows -> this is a generator
We need to loop through it till we get to the resume_from_row,
On l#820 we do that.

We then what to loop through the remaining part of "rows" in a chunk.
We want to exit the outside loop (it could be a break as well I think in this case).

Let us walk through an example
rows = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
resume_from = 2
Config.BATCH_INSERTION_CHUNK_SIZE = 4
the first loop will yield from the generator until:
rows = [4, 5, 6, 7, 8, 9, 10]

but as we will have already yielded [3] and that should be included (as it is at row.index 2 and we want to resume from there) -> this is the first_row logic

Once that happens we want to then chunk `rows` using the batch_size of 4
This is why the for loop `exits` after `rows` has completely yielded everything. 

Let me know if we need to meet irl, happy to explain it.

app/celery/tasks.py
jzbahrai: I don't see why this would matter, but I can change it.

app/celery/tasks.py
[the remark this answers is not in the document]
sastels: 🎉 Thanks!

sastels: I think I'd prefer a `break` here instead. Makes me think we're exiting the program (which... I'm confused that we're not?)

app/celery/tasks.py
jzbahrai: 😬

app/celery/tasks.py
jzbahrai: I changed it in the last commit

tests/app/celery/test_tasks.py
jimleroyer: Hmm these tests in the PR seems to test the happy path, but that one removed is for a SMS that shouldn't be sent, right?

app/celery/tasks.py
```diff
@@ -740,11 +582,28 @@ def handle_batch_error_and_forward(
...
+                task=task.__name__,
+                job=notification.get("job", None),
+                row=notification.get("row_number", None),
+                notif=notification_id,
+                max_retry=task.max_retries,
             )
+            current_app.logger.warning("Retry " + retry_msg)
+            try:
+                # If >1 notification has failed, we want to make individual
+                # tasks to retry those notifications.
```
jimleroyer, line 595: 👍

tests/app/celery/test_tasks.py
jzbahrai: True! I can make a note to add this test in another PR (as this code path wasn't being used anyways with the feature flag turned on ). Does that work?

jimleroyer: Yeah, as long as we track it the task card. It's not a blocker to this PR.

jzbahrai: added https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/cds-snc/notification-planning/774

PR #1572, opened 2022-07-13
Remove letter code from tasks.py , related tests and imports

jzbahrai:
# Summary | Résumé

Cleaned up letter code.
Removed code from app/tasks.py, test_tasks.py and all unused imports.

Removed all periphery code that called functions from app/tasks.py


Makefile
```diff
@@ -53,6 +53,7 @@ format:
 	isort .
 	black --config pyproject.toml .
 	flake8 .
+	mypy .
```
andrewleith, line 56: Do we need to run `mypy` here?  If so, should we do the same in admin?  `mypy` is currently run as part of `make test` (in `run_scripts.sh`).

app/celery/tasks.py
```diff
@@ -34,24 +33,14 @@
     put_batch_saving_bulk_created,
     put_batch_saving_bulk_processed,
 )
-from app.celery import (  # noqa: F401
```
andrewleith, line 37: Any idea why these were imported if not used?

app/letters/rest.py
```diff
@@ -14,6 +12,4 @@
 def create_process_returned_letters_job():
```
andrewleith, line 12: Should this whole method be removed?

app/notifications/notifications_letter_callback.py
```diff
@@ -55,7 +50,5 @@ def process_letter_response():
 
         if filename.lower().endswith("rs.txt") or filename.lower().endswith("rsp.txt"):
             current_app.logger.info("DVLA callback: Calling task to update letter notifications")
-            update_letter_notifications_statuses.apply_async([filename], queue=QueueNames.NOTIFY)
```
andrewleith, line 58: Should this whole method be removed?

tests/app/celery/test_ftp_update_tasks.py
```diff
@@ -44,182 +16,6 @@ def notification_update():
...
-            update_letter_notifications_to_error([first.reference])
-    assert first.reference in str(e.value)
-
-    assert first.status == NOTIFICATION_TECHNICAL_FAILURE
-    assert first.sent_by is None
-    assert first.sent_at is None
-    assert first.updated_at == dt
-    assert second.status == NOTIFICATION_CREATED
-
-
```
andrewleith, line 222: So long, letter tests!

tests/app/letters/test_returned_letters.py
```diff
@@ -1,6 +1,7 @@
 import pytest
 
 
+@pytest.mark.skip(reason="Deprecated: LETTER CODE")
```
andrewleith, line 4: Just curious why you are marking some things for deprecation and removing others?  Just the process you are following?

andrewleith (commented):
Overall looks good, just a few questions.

tests/app/letters/test_returned_letters.py
jzbahrai: this is good question! I basically was only concentrating on app/tasks.py and test_tasks.py. These two were the instigator. Now if the code from them was called else where I was deleting that code as well. In the above case - the actual rest API hasn't been removed but as it was mocking part of the tasks celery code, we were causing a failure. When the rest API is removed, I was thinking we can remove these further tests.

app/notifications/notifications_letter_callback.py
jzbahrai: yes it should, but then the tests needs to be removed. So falling the principle I was using below - I didn't remove it.

app/letters/rest.py
jzbahrai: yes it should, but then the tests needs to be removed. So following the principle I was using below - I didn't remove it.

app/celery/tasks.py
jzbahrai: unsure :/

Makefile
jzbahrai: I would add it in admin as well. its part of scripts/run_tests but wasn't part of `make format` . I generally run `make format` before I push

tests/app/letters/test_returned_letters.py
andrewleith: Makes sense, thanks for the clarification.

andrewleith (approved):
LGTM

PR #1573, opened 2022-07-19
add tests for team key with recipient not allowed

sastels:
# Summary | Résumé

Finish up https://github.com/cds-snc/notification-planning/issues/29: Add a few tests that we deleted in #1570 around not sending using a team key if the recipient isn't on the team.

Essentially, parametrized the existing happy-path tests to also test cases where sending is not allowed.

### DIFF
--- a/tests/app/v2/notifications/test_post_notifications.py
+++ b/tests/app/v2/notifications/test_post_notifications.py
@@ -816,10 +816,10 @@ class TestRestrictedServices:
     @pytest.mark.parametrize(
         "notification_type,to_key,to,response_code",
         [
-            (SMS_TYPE, "phone_number", "+16132532235", 201),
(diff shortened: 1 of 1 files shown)


tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -813,53 +813,79 @@ def test_send_notification_uses_appropriate_queue_according_to_template_process_
 
 
 class TestRestrictedServices:
-    @pytest.mark.parametrize("restricted", [True])
```
jimleroyer, line 816: hmm that was a weird parameterization. 🤷‍♂️

jimleroyer (approved):
Nice reuse of existing tests. 👏

PR #1576, opened 2022-07-20
Fix send-notify-no-reply

jzbahrai:
# Summary | Résumé

We were incorrectly sending a list object to `send_notification_to_queue`. We fixed this and also cleaned up some logs we had added while debugging why this function wasn't working as expected.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -885,8 +885,7 @@ def send_notify_no_reply(self, data):
         ]
         current_app.logger.info(f"Data we are sending to persist_notifications is {data_to_send}")
         saved_notifications = persist_notifications(data_to_send)
-        if saved_notifications and len(saved_notifications) == 1:
-            send_notification_to_queue(saved_notifications[0], False, queue=QueueNames.NOTIFY)
+        send_notification_to_queue(saved_notifications[0], False, queue=QueueNames.NOTIFY)
     except Exception as e:
         try:
             current_app.logger.warning(f"The exception is {repr(e)}")
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -889,7 +889,6 @@ def send_notify_no_reply(self, data):
             send_notification_to_queue(saved_notifications[0], False, queue=QueueNames.NOTIFY)
     except Exception as e:
         try:
-            current_app.logger.warning("We are going to retry send_no_reply")
             current_app.logger.warning(f"The exception is {repr(e)}")
             self.retry(queue=QueueNames.RETRY)
         except self.MaxRetriesExceededError:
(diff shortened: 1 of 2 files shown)


tests/app/celery/test_tasks.py
```diff
@@ -2048,8 +2048,8 @@ def test_process_returned_letters_list_updates_history_if_notification_is_alread
 
 
 class TestSendNotifyNoReply:
-    def test_send_notify_no_reply(self, mocker, no_reply_template):
-        persist_mock = mocker.patch("app.celery.tasks.persist_notifications")
+    def test_send_notify_no_reply(self, mocker, sample_notification, no_reply_template):
+        persist_mock = mocker.patch("app.celery.tasks.persist_notifications", return_value=[sample_notification])
```
jimleroyer, line 2052: 👍

andrewleith (approved):
LGTM

app/celery/tasks.py
```diff
@@ -886,13 +884,13 @@ def send_notify_no_reply(self, data):
...
         ]
         current_app.logger.info(f"Data we are sending to persist_notifications is {data_to_send}")
-        saved_notification = persist_notifications(data_to_send)
-        send_notification_to_queue(saved_notification, False, queue=QueueNames.NOTIFY)
+        saved_notifications = persist_notifications(data_to_send)
+        if saved_notifications and len(saved_notifications) == 1:
+            send_notification_to_queue(saved_notifications[0], False, queue=QueueNames.NOTIFY)
     except Exception as e:
         try:
             current_app.logger.warning("We are going to retry send_no_reply")
```
jimleroyer, line 892: Can we remove that line now?

app/celery/tasks.py
```diff
@@ -886,13 +884,13 @@ def send_notify_no_reply(self, data):
             )
         ]
         current_app.logger.info(f"Data we are sending to persist_notifications is {data_to_send}")
-        saved_notification = persist_notifications(data_to_send)
-        send_notification_to_queue(saved_notification, False, queue=QueueNames.NOTIFY)
+        saved_notifications = persist_notifications(data_to_send)
+        if saved_notifications and len(saved_notifications) == 1:
+            send_notification_to_queue(saved_notifications[0], False, queue=QueueNames.NOTIFY)
```
jimleroyer, line 889: What happens if the length is not 1? Do we discard these? Can't we safely assume it will always be one?

jzbahrai: if the length is not 1, it means we didn't save. I don't know what we should do here? I think this check should always pass but I can remove it if that makes more sense.

jimleroyer: Yeah I'd remove, it would make sense. If it is not 1, it will crash and we will know. If it is not 1, we will never know and won't be able to bring a fix.

jzbahrai: done

PR #1578, opened 2022-07-29
Add api key prefix

andrewleith:
# Summary | Résumé

Implement an identifiable api key prefix so that we can perform proactive security scanning to find keys that were mistakenly committed to public repositories.

## Notes
- Prefix value is stored in `config.py` (value is currently `gcntfy-`)
- Prefix value is added to the secret before encrypting so that it can't be easily removed (vs. storing it in the key name column, for example)

## Identifying our keys in the wild
Here is a regex expression to identify our keys:
- `(?P<prefix>gcntfy-)(?P<keyname>.*)(?P<service_id>[-A-Za-z0-9]{36})-(?P<keyid>[-A-Za-z0-9]{36})`


app/authentication/auth.py
```diff
@@ -154,6 +154,12 @@ def _auth_with_api_key(api_key, service):
             service_id=service.id,
             api_key_id=api_key.id,
         )
+    
+    # TODO: uncomment this when the grace period for the token prefix is over
```
andrewleith, line 158: Adding the validation logic now even though we won't use it as the mechanism is fresh in my mind.  If you would prefer this not to be here let me know.

app/authentication/auth.py
```diff
@@ -136,11 +136,19 @@ def requires_auth():
         raise AuthError("Invalid token: signature, api token not found", 403, service_id=service.id)
 
 
-def _auth_by_api_key(auth_token):
+def _auth_by_api_key(auth_token):
```
andrewleith, line 139: This will be the eventual validation we enable.  I wrote it now while everything was fresh and then commented it out.  If anyone prefers this to be omitted, just let me know.

app/config.py
```diff
@@ -134,6 +134,9 @@ class Config(object):
     SECRET_KEY = os.getenv("SECRET_KEY")
     DANGEROUS_SALT = os.getenv("DANGEROUS_SALT")
 
+    # API key prefix
```
andrewleith, line 137: API is the source of truth for the prefix value.  This is a static value across all environments.

app/service/rest.py
```diff
@@ -345,7 +345,11 @@ def create_api_key(service_id=None):
     valid_api_key.service = fetched_service
     save_model_api_key(valid_api_key)
     unsigned_api_key = get_unsigned_secret(valid_api_key.id)
-    return jsonify(data=unsigned_api_key), 201
+    
+    # prefix the API key so they keys can be easily identified for security scanning
+    keydata = { "data": unsigned_api_key, "key_name": current_app.config["API_KEY_PREFIX"] + valid_api_key.name }
```
andrewleith, line 350: This is where API stamps the prefix on and passes it back to the user (via admin).  The data contract had to change here, so admin requires an update for this as well.

app/authentication/auth.py
```diff
@@ -137,10 +137,19 @@ def requires_auth():
 
 
 def _auth_by_api_key(auth_token):
+    # TODO: uncomment this when the grace period for the token prefix is over
```
andrewleith, line 140: This will be the eventual validation we enable. I wrote it now while everything was fresh and then commented it out. If anyone prefers this to be omitted, just let me know.

app/service/rest.py
```diff
@@ -345,7 +345,11 @@ def create_api_key(service_id=None):
     valid_api_key.service = fetched_service
     save_model_api_key(valid_api_key)
     unsigned_api_key = get_unsigned_secret(valid_api_key.id)
-    return jsonify(data=unsigned_api_key), 201
+
```
andrewleith, line 348: This is where API stamps the prefix on and passes it back to the user (via admin). The data contract had to change here, so admin requires an update for this as well.

app/authentication/auth.py
```diff
@@ -137,10 +137,19 @@ def requires_auth():
...
+
     try:
         # take last 36 chars of string so that it works even if the full key is provided.
         auth_token = auth_token[-36:]
         api_key = get_api_key_by_secret(auth_token)
+
+        # TODO: uncomment this when the grace period for the token prefix is over
+        # check for token prefix
+        # if current_app.config["API_KEY_PREFIX"] not in orig_token:
+        #     raise AuthError("Invalid token: you must re-generate your API key to continue using GC Notify", 403, service_id=api_key.service.id, api_key_id=api_key.id)
```
sastels, line 151: They will also get this error if they just send the secret part as we currently require. We should add something here to point this out.

andrewleith: When you generate the key, admin gives you a single string that includes key name, svc id and secret id.  Are there docs somewhere that suggest to users not providing back exactly what admin spits out? 🤔

sastels: yeah, right now we explicitly tell users to only put the last uuid in the header 😞 
https://documentation.notification.canada.ca/en/start.html#headers

sastels (approved):
This all looks awesome and works locally. Just that comment around the (admittedly commented out) error. Maybe just put some sort of commented out note so we don't forget when we switch off the grace period.

app/authentication/auth.py
andrewleith: Very interesting.  While this does work when calling the API directly, it fails using any of the client libraries.  We should include this info in our communications when we are asking clients to regenerate their keys.  I'll make a note over it in that other card.

PR #1581, opened 2022-08-11
Re-create key_type prop when signing notifications.

andrewleith:
# Summary | Résumé

This PR fixes the issue where notify was ignoring the key_type for bulk sends using a testing API key.  Now, when the notification is signed after being posted, the key_type is correctly carried over.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -159,6 +159,7 @@ def process_rows(rows: List, template: Template, job: Job, service: Service):
         signed_row = signer.sign(
             {
                 "api_key": job.api_key_id and str(job.api_key_id),
+                "key_type": job.api_key.key_type if job.api_key else KEY_TYPE_NORMAL,
                 "template": str(template.id),
                 "template_version": job.template_version,
                 "job": str(job.id),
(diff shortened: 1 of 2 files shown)


app/celery/tasks.py
```diff
@@ -159,6 +159,7 @@ def process_rows(rows: List, template: Template, job: Job, service: Service):
         signed_row = signer.sign(
             {
                 "api_key": job.api_key_id and str(job.api_key_id),
+                "key_type": job.api_key.key_type if job.api_key else KEY_TYPE_NORMAL,
```
andrewleith, line 162: I'm not sure if it's possible for `job.api_key` not to exist but I added a check here so that if it didn't it would default to normal as it does elsewhere in the system.  If this should default more strictly (i.e. to a TEST type), please let me know.

sastels (approved):
Running locally with a test key: 
- the notifications are not sent
- a job and notifications are added to the database with correct key type set, and notification_status set to delivered
- I get a 201 back from my POST

This is identical to the current behaviour for posting to /email with a test key, so I think this is exactly the right way to fix this! 🎉

PR #1592, opened 2022-09-12
Feat/include sms parts in daily stats part1

andrewleith:
# Summary | Résumé

This PR adds `billable_units` to the `ft_notification_status` table so we can track statistics on SMS parts usage and enable daily and yearly limits for them.

### DIFF
--- a/app/dao/fact_notification_status_dao.py
+++ b/app/dao/fact_notification_status_dao.py
@@ -85,6 +85,7 @@ def query_for_fact_status_data(table, start_date, end_date, notification_type, s
             table.key_type,
             table.status,
             func.count().label("notification_count"),
+            func.sum(table.billable_units).label("billable_units"),
         )
         .filter(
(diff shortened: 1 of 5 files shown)


jzbahrai (approved):
this looks good to me! Has this been tested locally as well to see that it works correctly?

andrewleith:
> this looks good to me! Has this been tested locally as well to see that it works correctly?

Yes I've tested it locally and everything seems to work fine.  There should be are no visible changes as a result of this in ADMIN, as the new field would be ignored.

jzbahrai (approved):

smcmurtry (approved):
I've now tested this locally sending a multipart sms message, and after running the celery job to update the fact table, I'm seeing the `billible_units` column there being set correctly. Looks good!

Screen Shot 2022-09-21 at 11 58 43 AM

PR #1599, opened 2022-09-22
Add `billable_units` to DAO queries

andrewleith:
# Summary | Résumé

This PR adds `billable_units` to the 2 queries needed for https://github.com/cds-snc/notification-admin/pull/1334 when the `FF_SPIKE_SMS_DAILY_LIMIT` is turned on:
- fetch_notification_status_for_service_for_today_and_7_previous_days
- _stats_for_service_query

## Notes
At first I was trying to add `billable_units` data to the current payload which entailed changing the data contracts a bit between API and ADMIN and created a lot of side effects I was having trouble accounting for.

Instead, I opted to replace the `count` field with the data from `billable_units` only when the type was `sms` so that no data contracts would need to change.  Now, when admin would get the information for total sms sent, it will be total **sms parts** sent instead (at least for the 2 queries updated here!)

##### `_stats_for_service_query`
- [ ] Ensure the sum of `billable_units` are returned in the `count` field from the method `_stats_for_service_query` for notifications of type `sms`
- [ ] Ensure the count of `Notification_id` are returned in the `count` field from the method `_stats_for_service_query` for notifications of type `email`
##### `fetch_notification_status_for_service_for_today_and_7_previous_days`
- [ ] Ensure `billable_units` are returned in the `count` field from the method `fetch_notification_status_for_service_for_today_and_7_previous_days` for notifications of type `sms`
- [ ] Ensure `notification_count` are returned in the `count` field from the method `fetch_notification_status_for_service_for_today_and_7_previous_days` for notifications of type `email`


app/dao/fact_notification_status_dao.py
```diff
@@ -246,7 +246,12 @@ def fetch_notification_status_for_service_for_today_and_7_previous_days(service_
         FactNotificationStatus.notification_type.label("notification_type"),
         FactNotificationStatus.notification_status.label("status"),
         *([FactNotificationStatus.template_id.label("template_id")] if by_template else []),
-        FactNotificationStatus.notification_count.label("count"),
+        *([case(
+                [
+                    (FactNotificationStatus.notification_type == 'email', FactNotificationStatus.notification_count),
+                ],
+                else_=FactNotificationStatus.billable_units,
+        ).label("count")] if current_app.config["FF_SPIKE_SMS_DAILY_LIMIT"] else [FactNotificationStatus.notification_count.label("count")]),
```
andrewleith, line 254: This is a little hard to read at first, but is roughly this in SQL:
```
SELECT notification_type, notification_status, template_id
  CASE 
    WHEN notification_type = 'email' THEN notification_count
    ELSE billable_units
  END AS count,
FROM ft_notification_status
```

app/dao/fact_notification_status_dao.py
```diff
@@ -246,7 +246,18 @@ def fetch_notification_status_for_service_for_today_and_7_previous_days(service_
...
+                case(
+                    [
+                        (FactNotificationStatus.notification_type == "email", FactNotificationStatus.notification_count),
+                    ],
+                    else_=FactNotificationStatus.billable_units,
+                ).label("count")
+            ]
+            if current_app.config["FF_SPIKE_SMS_DAILY_LIMIT"]
+            else [FactNotificationStatus.notification_count.label("count")]
+        ),
```
andrewleith, line 260: This creates a SQL case statement to choose either `notification_count` or `billable_units`, depending on `notification_type`, roughly this in SQL:
```
SELECT notification_type, notification_status, template_id
  CASE 
    WHEN notification_type = 'email' THEN notification_count
    ELSE billable_units
  END AS count,
FROM ft_notification_status
```

sastels (changes requested):
Looks good, just need to make sure that for now we're testing both FF on and FF off cases.

app/dao/fact_notification_status_dao.py
andrewleith: Yes! It is definitely strange looking syntax, but that is the api that [SqlAlchemy provides for a case statement](https://docs.sqlalchemy.org/en/14/core/sqlelement.html#sqlalchemy.sql.expression.case)!

tests/app/dao/test_fact_notification_status_dao.py
```diff
@@ -328,17 +329,30 @@ def test_fetch_notification_status_by_template_for_service_for_today_and_7_previ
...
-        ("email Template Name", False, mock.ANY, "email", "delivered", 3),
-        ("letter Template Name", False, mock.ANY, "letter", "delivered", 5),
-        ("sms Template 1", False, mock.ANY, "sms", "created", 1),
-        ("sms Template Name", False, mock.ANY, "sms", "created", 1),
-        ("sms Template 1", False, mock.ANY, "sms", "delivered", 1),
-        ("sms Template 2", False, mock.ANY, "sms", "delivered", 1),
-        ("sms Template Name", False, mock.ANY, "sms", "delivered", 10),
-        ("sms Template Name", False, mock.ANY, "sms", "delivered", 11),
-    ] == sorted(results, key=lambda x: (x.notification_type, x.status, x.template_name, x.count))
+    if current_app.config["FF_SPIKE_SMS_DAILY_LIMIT"]:
```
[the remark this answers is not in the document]
andrewleith: Interesting point. I was thinking the tests will run as they did before when the FF is off (i.e. in prod) and test against the new functionality locally and in staging once the FF is on.  

But what you are proposing sounds cleaner, let me try that!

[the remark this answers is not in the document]
andrewleith: Done!

app/dao/fact_notification_status_dao.py
```diff
@@ -258,7 +269,18 @@ def fetch_notification_status_for_service_for_today_and_7_previous_days(service_
...
+                [
+                    case(
+                        [
+                            (Notification.notification_type == "email", func.count()),
+                        ],
+                        else_=func.sum(Notification.billable_units),
+                    ).label("count")
+                ]
+                if current_app.config["FF_SMS_PARTS_UI"]
+                else [func.count(Notification.id).label("count")]
```
sastels, line 282: I think `func.count(Notification.id).label("count")` is the same as `func.count().label("count")` - any particular reason for the change in the FF off case?

andrewleith: Hmm I think I just explicitly typed this out, probably confused myself with all of this horrid syntax :)

It SHOULD be the same, but I'll set it back to how it was exactly in the else case.  Good eye!
```suggestion
                else [func.count().label("count")]
```

sastels (approved):
Looks good!

smcmurtry (approved):
I did a brief test with this PR + my dashboard ui PR and the "SMS remaining" numbers are as I would expect both with the feature flag on and off. Looks good!

PR #1606, opened 2022-09-27
Add check for sms daily limit

jzbahrai:
# Summary | Résumé

Adds a check against `sms_daily_limit` when using the `/bulk` endpoint.

This PR does at least add a check against the service's `sms_daily_limit` (though, as mentioned, just comparing the number of sms notifications, not the number of message parts).

- you will need a `pip install -r requirements.txt`
- set your `sms_daily_limit` to 3
- try to post to `/bulk` with 4 or more sms fragments (in any number of messages).
- you should get an error message back and no texts sent.
```
{
    "errors":[
        {
            "error": "BadRequestError",
            "message": "You only have 3 remaining sms message parts before you reach your daily limit. You've tried to send 4 message parts."
        }
    ],
    "status_code": 400
}
```

None.


requirements-app.txt
```diff
@@ -50,7 +50,8 @@ MarkupSafe==2.0.1
 # REVIEW: v2 is using sha512 instead of sha1 by default (in v1)
 itsdangerous==2.0.1
 
-git+https://github.com/cds-snc/notifier-utils.git@48.3.0#egg=notifications-utils
+# git+https://github.com/cds-snc/notifier-utils.git@48.3.0#egg=notifications-utils
+git+https://github.com/cds-snc/notifier-utils.git@test-sms-in-RecipientCSV#egg=notifications-utils
```
sastels, line 54: TODO: fix after utils PR merges

requirements.txt
```diff
@@ -52,7 +52,8 @@ MarkupSafe==2.0.1
 # REVIEW: v2 is using sha512 instead of sha1 by default (in v1)
 itsdangerous==2.0.1
 
-git+https://github.com/cds-snc/notifier-utils.git@48.3.0#egg=notifications-utils
+# git+https://github.com/cds-snc/notifier-utils.git@48.3.0#egg=notifications-utils
+git+https://github.com/cds-snc/notifier-utils.git@test-sms-in-RecipientCSV#egg=notifications-utils
```
sastels, line 56: TODO: fix after utils PR merges

andrewleith:
There are a lot of balls in the air right now which may explain this, but I get a different limit in the admin vs the API:
**api**
image

**admin**
image

sastels:
> There are a lot of balls in the air right now which may explain this, but I get a different limit in the admin vs the API: **api** image
> 
> **admin** image

This is quite possible - we're computing in 2 different ways that should be the same but there could be (well - are) differences - api is mainly using redis to maintain a counter (so that it can check for each email) while the frontend is just doing a fresh query each time. This should be its own "find and fix the discrepancy" card. (We could also switch admin to also using the redis counter instead of the query if/when we have faith in the redis counter).

andrewleith:
> > There are a lot of balls in the air right now which may explain this, but I get a different limit in the admin vs the API: **api** image
> > **admin** image
> 
> This is quite possible - we're computing in 2 different ways that should be the same but there could be (well - are) differences - api is mainly using redis to maintain a counter (so that it can check for each email) while the frontend is just doing a fresh query each time. This should be its own "find and fix the discrepancy" card. (We could also switch admin to also using the redis counter instead of the query if/when we have faith in the redis counter).

I vote for changing admin to use the redis counter, since that is what is used on the backend for the validation, it should be what the user sees as well.  Getting it into the UI will help us find any bugs easier also!

PR #1607, opened 2022-09-27
fix/check daily_sms_limit wherever we check daily_message_limit

sastels:
# Summary | Résumé

There were several places that we were calling `check_service_over_daily_message_limit` that we weren't calling `check_service_over_daily_sms_limit`. This PR fixes that.

Basically, whenever you send an sms for the first time in a day, you should see the redis counter `sms-<service_id>-date-count` created.

(example: `sms-3ae2fb6a-bec7-47d6-8f8b-36e2519f7c6d-2022-09-27-count`)

None.

### DIFF
--- a/app/notifications/rest.py
+++ b/app/notifications/rest.py
@@ -97,7 +97,7 @@ def send_notification(notification_type):
     if errors:
         raise InvalidRequest(errors, status_code=400)
 
-    check_rate_limiting(authenticated_service, api_user)
+    check_rate_limiting(authenticated_service, api_user, notification_type)
 
     template = templates_dao.dao_get_template_by_id_and_service_id(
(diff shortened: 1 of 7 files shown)


app/notifications/rest.py
```diff
@@ -97,7 +97,7 @@ def send_notification(notification_type):
     if errors:
         raise InvalidRequest(errors, status_code=400)
 
-    check_rate_limiting(authenticated_service, api_user)
+    check_rate_limiting(authenticated_service, api_user, notification_type)
```
andrewleith, line 100: 👍 this makes sense to me, i like this pattern of passing the notification type since we will now have 2 different checks

andrewleith (approved):
LGTM

PR #1612, opened 2022-09-29
add special errors to return for over sms limits

sastels:
# Summary | Résumé

These error messages will be returned to admin during manual sends which will then be able to display an appropriate error message to the user.

Note that if the FF is off these new errors won't be used.

Run together with https://github.com/cds-snc/notification-admin/pull/1343

None.


andrewleith (approved):
LGTM

PR #1613, opened 2022-09-29
Fix issues regarding the celery script

jzbahrai:
# Summary | Résumé

There were two bugs in the celery exit script.
1. The PIDs on the kube setup show up in the first column (locally it shows up in column 2)
2. I was only printing (🤦🏽 ) the kill command, I needed to execute it as well.

### DIFF
--- a/scripts/run_celery_exit.sh
+++ b/scripts/run_celery_exit.sh
@@ -10,7 +10,7 @@ function get_celery_pids {
   # and keep only these PIDs
 
   set +o pipefail # so grep returning no matches does not premature fail pipe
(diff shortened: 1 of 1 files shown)


scripts/run_celery_exit.sh
```diff
@@ -23,6 +23,7 @@ function send_signal_to_celery_processes {
   for value in ${APP_PIDS}
   do
     echo kill -s ${1} $value
+    kill -s ${1} $value
```
sastels, line 26: 😂

sastels (approved):
:shipit:

PR #1619, opened 2022-10-06
Fix/non default replyto

andrewleith:
# Summary | Résumé

This PR makes it possible to select a non-default reply to when multiple have been created.  This applies to bulk and one-off API sending.

---

## Test instructions | Instructions pour tester la modification

- Setup a service with 2 reply-to addresses.
- Test API one-off:
   - Send a one-off email notification and include the `email_reply_to_id` of the non-default reply to address
   - Ensure that reply-to address is set in the email you receive
- Test API bulk:
   - Send a bulk email notification and include the `reply_to_id` of the non-default reply to address
   - Ensure that reply-to address is set in the email you receive

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -357,22 +357,22 @@ def save_emails(self, service_id: Optional[str], signed_notifications: List[Any]
             else:
                 reply_to_text = template.get_reply_to_text()  # type: ignore
 
-        notification["reply_to_text"] = reply_to_text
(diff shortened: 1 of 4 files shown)


sastels (approved):
- Code looks good
- tested locally with /email
- tested locally with /bulk

Sad that we sometimes call it "email_reply_to_id", sometimes "reply_to_id" and in the code it's often "reply_to_text" 😞

PR #1620, opened 2022-10-06
perf test README redirect

sastels:
# Summary | Résumé

Add a README for tests-perf pointing user to the subdirectories, where we have the real documentation.


andrewleith (approved):
:shipit:

PR #1622, opened 2022-10-06
Add sleep function for testing

jzbahrai:
# Summary | Résumé

This is for testing celery and to see what happens when we kill celery when a process is still running.


sastels (approved):
💤

PR #1624, opened 2022-10-06
Added some type definitions to `models.py` and `validators.py`

smcmurtry:
# Summary | Résumé

I wanted to add some type annotations to `validators.py` to help make it easier to fix the "over sms daily limit" logic and error messages. This PR should not have an effect on app behaviour - the types are purely annotations to make development easier.

I referred to the mypy docs here: https://mypy.readthedocs.io/en/stable/literal_types.html

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -930,7 +930,7 @@ class ApiKey(BaseModel, Versioned):
         if secret:
             self._secret = signer.sign(str(secret))
 
-
+ApiKeyType = Literal["normal", "team", "test"]
 KEY_TYPE_NORMAL = "normal"
 KEY_TYPE_TEAM = "team"
 KEY_TYPE_TEST = "test"
@@ -1269,6 +1269,7 @@ SMS_PROVIDERS = [SNS_PROVIDER]
 EMAIL_PROVIDERS = [SES_PROVIDER]
 PROVIDERS = SMS_PROVIDERS + EMAIL_PROVIDERS
 
+NotificationType = Literal["email", "sms", "letter"]
 NOTIFICATION_TYPE = [EMAIL_TYPE, SMS_TYPE, LETTER_TYPE]
 notification_types = db.Enum(*NOTIFICATION_TYPE, name="notification_type")
 
--- a/app/models.py
+++ b/app/models.py
@@ -1,7 +1,7 @@
 import datetime
 import itertools
 import uuid
-from typing import Any, Iterable, Optional
+from typing import Any, Iterable, Literal, Optional
 
 from flask import current_app, url_for
 from flask_sqlalchemy.model import DefaultMeta
@@ -33,6 +33,8 @@ from app import DATETIME_FORMAT, db, signer
 from app.config import QueueNames
 from app.encryption import check_hash, hashpw
 from app.history_meta import Versioned
+
+TemplateType = Literal["sms", "email", "letter"]
 
 SMS_TYPE = "sms"
 EMAIL_TYPE = "email"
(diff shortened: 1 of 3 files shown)


app/models.py
```diff
@@ -34,6 +34,8 @@
 from app.encryption import check_hash, hashpw
 from app.history_meta import Versioned
 
+TemplateType = Literal["sms", "email", "letter"]
```
sastels, line 37: We already have
https://github.com/cds-snc/notification-api/pull/1624/files#diff-90c680eba43456b516da8b4c2573d467ae17d1b0ed4373549f2a593ced3616d5R43 

Probably just need one of these?

app/models.py
```diff
@@ -929,6 +931,7 @@ def secret(self, secret):
             self._secret = signer.sign(str(secret))
 
 
+ApiKeyType = Literal["normal", "team", "test"]
 KEY_TYPE_NORMAL = "normal"
 KEY_TYPE_TEAM = "team"
 KEY_TYPE_TEST = "test"
```
sastels, line 937: Should use these constants in the `ApiKeyType`

app/models.py
```diff
@@ -1267,6 +1270,7 @@ def get_link(self):
 EMAIL_PROVIDERS = [SES_PROVIDER]
 PROVIDERS = SMS_PROVIDERS + EMAIL_PROVIDERS
 
+NotificationType = Literal["email", "sms", "letter"]
 NOTIFICATION_TYPE = [EMAIL_TYPE, SMS_TYPE, LETTER_TYPE]
```
sastels, line 1274: This seems duplicating too - maybe I'm missing something...

sastels (changes requested):
A few questions / suggestions, but adding types is always good for sure!

app/models.py
smcmurtry: Looks like that just links to this PR

app/models.py
smcmurtry: So I tried:
```ApiKeyType  = Literal[KEY_TYPE_NORMAL, KEY_TYPE_TEAM, KEY_TYPE_TEST]```

But that throws an error since a `Literal` needs to be composed of other `Literal`s such as strings, and not pointers. So I thought maybe I could re-assign `KEY_TYPE_NORMAL = Literal["normal"]` and then the statement above would work. But then `KEY_TYPE_NORMAL` just represents a type rather than a value, and can no longer be used like this: `if some_var == KEY_TYPE_NORMAL :...`, since that will throw an error.

So an alternative that would work would be to set:
```
KeyTypeNormal = Literal["normal"]
KEY_TYPE_NORMAL: KeyTypeNormal = "normal"
...
ApiKeyType = Literal[KeyTypeNormal, ...]
```

But that doesn't decrease the amount of repetition needed to create the type. 

So I couldn't get around the duplication, but let me know if you can figure one out.

app/models.py
smcmurtry: See my comment here: https://github.com/cds-snc/notification-api/pull/1624/files#r990332042

app/models.py
smcmurtry: Okay so it turns out `ApiKeyType = Literal[KEY_TYPE_NORMAL, KEY_TYPE_TEAM, KEY_TYPE_TEST]  # type: ignore ` works! The type created is correct, and we can just ignore mypy's complaints about this.

sastels (approved):
LGTM! Will definitely make development easier 😅

PR #1625, opened 2022-10-07
Refactor code around sms fragments in redis

sastels:
# Summary | Résumé

Refactor the code around the sms counter to make
- the redis counter always equal the number of fragments sent
- make the code much easier to read / use

My devcontainer is being flakey 😞  but it looks like this might have fixed
https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/cds-snc/notification-planning/912

Should verify this in staging though.

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -130,7 +130,6 @@ from app.service.service_senders_schema import (
     add_service_sms_sender_request,
 )
 from app.service.utils import get_safelist_objects
-from app.sms_fragment_utils import delete_daily_sms_fragment_count
 from app.user.users_schema import post_set_permissions_schema
 from app.utils import pagination_links
 
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -297,7 +297,6 @@ def update_service(service_id):
         if not fetched_service.restricted:
             _warn_service_users_about_message_limit_changed(service_id, current_data)
     if sms_limit_changed:
-        delete_daily_sms_fragment_count(service_id)
         redis_store.delete(near_sms_daily_limit_cache_key(service_id))
         redis_store.delete(over_sms_daily_limit_cache_key(service_id))
         if not fetched_service.restricted:
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -8,7 +8,6 @@ from notifications_utils.clients.redis import (
     near_sms_daily_limit_cache_key,
     over_daily_limit_cache_key,
     over_sms_daily_limit_cache_key,
-    sms_daily_count_cache_key,
 )
 from notifications_utils.letter_timings import letter_can_be_cancelled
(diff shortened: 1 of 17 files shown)


app/sms_fragment_utils.py
```diff
@@ -0,0 +1,33 @@
...
+    if current_app.config["REDIS_ENABLED"]:
+        fetch_daily_sms_fragment_count(service_id)  # to make sure it's set in redis
+        cache_key = sms_daily_count_cache_key(service_id)
+        redis_store.incrby(cache_key, increment_by)
+
+
+def delete_daily_sms_fragment_count(service_id: UUID):
+    if current_app.config["REDIS_ENABLED"]:
+        cache_key = sms_daily_count_cache_key(service_id)
+        redis_store.delete(cache_key)
```
andrewleith, line 33: Nice! So clean! 🧹

app/service/rest.py
```diff
@@ -297,7 +297,7 @@ def update_service(service_id):
         if not fetched_service.restricted:
             _warn_service_users_about_message_limit_changed(service_id, current_data)
     if sms_limit_changed:
-        redis_store.delete(sms_daily_count_cache_key(service_id))
+        delete_daily_sms_fragment_count(service_id)
```
andrewleith, line 300: Why does this get deleted when the limit changes?  Doesn't this number represent the number of fragments sent today?

app/service/rest.py
```diff
@@ -297,7 +297,7 @@ def update_service(service_id):
...
         redis_store.delete(near_sms_daily_limit_cache_key(service_id))
```
andrewleith, line 301: `near_sms_daily_limit_cache_key` and `over_sms_daily_limit_cache_key` don't appear to be used anywhere in the API - Are they for future enhancements?

app/service/rest.py
sastels: yeah.... I had copied the code around the general daily limit, but that cache key is called "daily_limit_cache_key" and it looks like Antoine thought it was the cache key for the daily limit (silly Antoine) and so he added code here to delete it https://github.com/cds-snc/notification-api/pull/1258

But that's just a mistake caused by our bad naming

app/service/rest.py
sastels: We use them to check if we've already sent the service owners the warning / alert emails
https://github.com/cds-snc/notification-api/blob/c8393182334506d1ec0d0648319bed6569021ea4/app/notifications/validators.py#L180-L181
https://github.com/cds-snc/notification-api/blob/c8393182334506d1ec0d0648319bed6569021ea4/app/notifications/validators.py#L202-L203

andrewleith (approved):
LGTM!

PR #1626, opened 2022-10-07
Added more type stubs/definitions for SQLAlchemy

smcmurtry:
# Summary | Résumé

I found another package with more type definitions for SQLAlchemy. This allows us to remove the "mypy ignore" setting for that import.

### DIFF
--- a/requirements_for_test.txt
+++ b/requirements_for_test.txt
@@ -19,6 +19,6 @@ jinja2-cli[yaml]==0.8.2
 black==21.5b2
 locust==2.10.1
 mypy==0.812
-sqlalchemy-stubs==0.4
+sqlalchemy2-stubs==0.0.2a27
 networkx==2.8.6 # not directly required, pinned by Snyk to avoid a vulnerability
 pytest-mock-resources[redis]==2.4.0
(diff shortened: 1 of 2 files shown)


requirements_for_test.txt
```diff
@@ -20,5 +20,6 @@ black==21.5b2
 locust==2.10.1
 mypy==0.812
 sqlalchemy-stubs==0.4
```
andrewleith, line 22: Should we still have the old stubs package here?  From the sqlalchemy2-stubs README:

> The stubs replace the use of the "sqlalchemy-stubs" package published by Dropbox.

It also mentions a mypy extension to go with this - do we need that as well?

smcmurtry: > It also mentions a mypy extension to go with this - do we need that as well?
We've got it already here: https://github.com/cds-snc/notification-api/blob/main/requirements_for_test.txt#L21
And we are running it in our CI here: https://github.com/cds-snc/notification-api/blob/main/scripts/run_tests.sh#L35

smcmurtry: > Should we still have the old stubs package here?

Interesting that they say they replace the old stubs. When I removed the old ones I was getting errors. But I'll try again and see what happens.

jzbahrai (approved):

PR #1627, opened 2022-10-07
Remove cloudfoundry related code.

CalvinRodo:
# Summary | Résumé

There was some old Code and tests that were related to running the notify api on cloudfoundry, it was pretty nicely segregated from the rest of the code so figured it was an easy PR to just make things a little tidier.

I took a look at what's running in staging and prod and also poked at the notification-manifests repo and can't find any reference to the `VCAP_SERVICES` environment variable anywhere so this should have no adverse impact to notify.

VCAP_SERVICES is used by CloudFoundry to inject information to running containers if you were curious what it did: https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#VCAP-SERVICES

Doesn't look like this requires any sort of testing as it's just removing unused code.


jimleroyer (approved):
LGTM, thanks for removing the cruft.

CalvinRodo:
@jimleroyer lemme know if you want me to merge this to main or if someone from notify should

PR #1644, opened 2022-10-17
Fix/hop over sms limit

sastels:
# Summary | Résumé

https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/cds-snc/notification-planning/931

Check if there's enough fragments left in the `/sms` send

- Set your SMS limit to 1
- Send a 2 part `/sms`
- should get an error message

None.


smcmurtry (approved):
I was able to reproduce the bug on staging. I tried the same "hop" test running your branch locally and the bug was fixed! Looks good.

PR #1645, opened 2022-10-19
Fix for bug where user can schedule and send more SMS than daily limit allows

smcmurtry:
# Summary | Résumé

Working on https://github.com/cds-snc/notification-planning/issues/948

This PR is 99% type annotations and renaming variables that were being over-written in a confusing way. I might have gone a bit overboard. But there were so many objects with the same names getting defined and redefined that I was not able to change the code without adding some type safety. 

After adding the type definitions I believe I found a few bugs in the code as well which I indicated with comments.

~Draft for now as there are probably some tests to fix.~ Fixed. ~I may also try to add the limit check to the cleanup task that catches notifications more than 2 hours old.~ I'll do that in a separate PR since we may not want to include it in V1.

      fonctionnalité existante?


app/celery/tasks.py
```diff
@@ -287,58 +297,57 @@ def save_smss(self, service_id: Optional[str], signed_notifications: List[Any],
 
     check_service_over_daily_message_limit(KEY_TYPE_NORMAL, service)
 
-    research_mode = service.research_mode  # type: ignore
-
     current_app.logger.info(f"Sending following sms notifications to AWS: {notification_id_queue.keys()}")
-    for notification in saved_notifications:
-        queue = notification_id_queue.get(notification.id) or template.queue_to_use()  # type: ignore
+    for notification_obj in saved_notifications:
+        check_if_request_would_put_service_over_daily_sms_limit(KEY_TYPE_NORMAL, service, notification_obj.billable_units)  # type: ignore
```
smcmurtry, line 302: Here is the only "real" change I made to the logic, to add the additional limit check

app/celery/tasks.py
```diff
@@ -213,24 +216,25 @@ def save_smss(self, service_id: Optional[str], signed_notifications: List[Any],
...
         service = dao_fetch_service_by_id(service_id, use_cache=True)
 
         template = dao_get_template_by_id(
-            notification.get("template"), version=notification.get("template_version"), use_cache=True
+            _notification.get("template"), version=_notification.get("template_version"), use_cache=True
         )
-        sender_id = notification.get("sender_id")
-        notification_id = notification.get("id", create_uuid())
-        notification["notification_id"] = notification_id
+        # bug: _notification does not have "sender_id" key
```
jzbahrai, line 234: can you put it as a TODO - I think we should remove this in the next PR - but merge this first without removing it

app/celery/tasks.py
```diff
@@ -287,58 +297,57 @@ def save_smss(self, service_id: Optional[str], signed_notifications: List[Any],
...
+                notification_obj.created_at,
+                notification_obj.job,
             )
         )
 
 
 @notify_celery.task(bind=True, name="save-emails", max_retries=5, default_retry_delay=300)
 @statsd(namespace="tasks")
-def save_emails(self, service_id: Optional[str], signed_notifications: List[Any], receipt: Optional[UUID]):
+def save_emails(self, _service_id: Optional[str], signed_notifications: List[Any], receipt: Optional[UUID]):
```
jzbahrai, line 322: why did we rename this variable?

app/celery/tasks.py
```diff
@@ -287,58 +297,57 @@ def save_smss(self, service_id: Optional[str], signed_notifications: List[Any],
...
-        notification_id = notification.get("id", create_uuid())
-        notification["notification_id"] = notification_id
+        # bug: _notification does not have key "sender_id"
+        sender_id = _notification.get("sender_id")  # type: ignore
+        notification_id = _notification.get("id", create_uuid())
         reply_to_text = ""  # type: ignore
-
         if (
-            "reply_to_text" in notification and notification["reply_to_text"]
+            "reply_to_text" in _notification and _notification["reply_to_text"]
```
jzbahrai, line 348: are you renaming to indicate private variables? by default everything is private within the scope of the func?

app/celery/tasks.py
```diff
@@ -377,6 +391,9 @@ def save_emails(self, service_id: Optional[str], signed_notifications: List[Any]
             f"Saved following notifications into db: {notification_id_queue.keys()} associated with receipt {receipt}"
         )
         if receipt:
+            # bug: template is whatever it was set to last in the for loop above
+            # at this point in the code we have a list of notifications (saved_notifications)
+            # which could use multiple templates
```
jzbahrai, line 396: this seems legit, lets discuss offline

app/celery/tasks.py
smcmurtry: Yes good idea! I'll do that.

app/celery/tasks.py
smcmurtry: Renamed because below on line 338, `service_id` is set to a different value. It goes from being `Optional[str]` to `str`, i.e. it can't be `None` anymore, since it is used to look up the service from the db. The type checker mypy fails the check if a variable changes type. 

But I probably could have chosen a better name here!

smcmurtry: I think mypy enforces the rule that variables shouldn't change types because it makes the code easier to understand

app/celery/tasks.py
smcmurtry: So before, we just had one `notification` variable that started out containing the signed data, and then the code added additional keys one by one. I wanted to define what keys are present in the signed data (that's the `NotificationDictToSign` type), and what additional keys need to be there before the notification is created in the db (that's the `VerifiedNotification` type).  

So that meant having two variables instead of one:  `_notification` for the signed data and `notification` for the signed data plus the additional keys we need to add in.

But I could probably have chosen different names that are a bit more descriptive.

app/celery/tasks.py
```diff
@@ -245,21 +249,27 @@ def save_smss(self, service_id: Optional[str], signed_notifications: List[Any],
...
+            "api_key_id": _notification.get("api_key", None),
+            "created_at": datetime.utcnow(),
+            "job_id": _notification.get("job", None),
+            "job_row_number": _notification.get("row_number", None),
+        }
+
         verified_notifications.append(notification)
-        notification_id_queue[notification_id] = notification.get("queue")
+        # bug: notification does not have "queue" key
+        notification_id_queue[notification_id] = notification.get("queue")  # type: ignore
```
jzbahrai, line 272: I thought I put a comment here 🤦🏽 this is coming from the signed notification - the notification table has a "queue_name" -> we are sending it through the process_rows function https://github.com/cds-snc/notification-api/blob/main/app/celery/tasks.py#L169

smcmurtry: Okay thanks for pointing that out! I hadn't noticed that function. I'll add `queue` to the `NotificationDictToSign` type

app/celery/tasks.py
sastels, line 302: We should get the service for this particular `notification_obj`

smcmurtry: Yes you're right! I'll fix that.

sastels:
Running locally with limit 4 fragments, sending 2 small sms via `/bulk`, the notifications get created but not sent. 2 emails sent through `/bulk` went through fine.
image

smcmurtry:
I was able to get in here with the debugger and it turns out `billable_units` and `service` are missing from the `notification_obj`. The object is created over here and it looks like it is not actually of type `Notification` since it hasn't been saved in the db yet: https://github.com/cds-snc/notification-api/blob/main/app/notifications/process_notifications.py#L335

I will just use `0` for billable units to prevent sending if we're already over, similar to how the combined daily limit works.

smcmurtry:
Okay I've tested this some more and I believe it should be working similarly to the combined limit check now.

sastels (approved):
Love the added `typing`!

Tested locally, previous "can't send" behaviour mentioned above does not happen, and extra scheduled jobs get stuck in "sending" limbo just as they do for jobs over the general daily limit.

:shipit:

PR #1653, opened 2022-10-25
Update SMS daily limit error message content

smcmurtry:
# Summary | Résumé

Update SMS daily limit error message content for clarity

### DIFF
--- a/app/v2/errors.py
+++ b/app/v2/errors.py
@@ -32,7 +32,7 @@ class TooManyRequestsError(InvalidRequest):
 
 class TooManySMSRequestsError(InvalidRequest):
     status_code = 429
-    message_template = "Exceeded SMS daily sending limit of {}"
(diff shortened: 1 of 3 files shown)


app/v2/errors.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, sending_limit):
 
 class TooManySMSRequestsError(InvalidRequest):
     status_code = 429
-    message_template = "Exceeded sms send limits ({}) for today"
+    message_template = "Exceeded SMS daily sending limit of {}"
```
sastels, line 35: I think if we leave the number in (which is fine because you've removed the confusion by explicitly said what it is) then we should say its for fragments, maybe something like "Exceeded SMS fragments daily sending limit of {}"?

sastels (changes requested):
Just another small request to remove a differnt potential for confusion 😂

PR #1654, opened 2022-10-26
Update the /bulk endpoint sms limit checking to use the shared logic

smcmurtry:
# Summary | Résumé

Update the /bulk endpoint sms limit checking to use the shared logic.

This is a fix for a bug @andrewleith found where post requests to /bulk were not failing even though he had submitted more sms fragments for sending than his limit allowed. These SMS were stuck in sending.


sastels (approved):
LGTM! using /bulk, sends when it should, returns error when it should, sends warning emails.
Get the error from `check_service_over_daily_sms_limit()` which is good.
```
{
    "errors":[
        {
            "error": "LiveServiceTooManySMSRequestsError",
            "message": "Exceeded SMS daily sending limit of 6 fragments"
            }
    ],
    "status_code": 429
}
```

PR #1655, opened 2022-10-26
Fix "test runner" tab in VSCode

smcmurtry:
# Summary | Résumé

Updating these values in the pytest config fixes local test failures I was seeing and allows me to run the tests using VSCodes "test runner" tab. 

Draft for now since I am not sure if CI will like this change or not.


smcmurtry:
This change didn't work in CI so I will close for now.

PR #1656, opened 2022-10-27
fix(docker): pin redis to 6.2 to match staging/prod/devcontainer

andrewleith:
# Summary | Résumé

This PR is to keep the docker image on the same version of redis as everything else.


patheard (approved):
Confirmed the image can pull and the included redis version is 6.2.7 👍🏻

PR #1660, opened 2022-10-31
Add the warning email trigger to the csv send via admin codepath

smcmurtry:
# Summary | Résumé

This to fix a bug Jimmy found where the SMS Limit related warning emails were not being triggered for a send via the Admin site.

The check we were performing (`check_if_request_would_put_service_over_daily_sms_limit`) did not include the logic to send the warning emails. I switched it to the other check function that is responsible for sending these emails:  `check_service_over_daily_sms_limit`.

See bug card.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -70,11 +70,13 @@ from app.models import (
     Template,
 )
 from app.notifications.process_notifications import (
-    check_if_request_would_put_service_over_daily_sms_limit,
     persist_notifications,
     send_notification_to_queue,
 )
-from app.notifications.validators import check_service_over_daily_message_limit
+from app.notifications.validators import (
+    check_service_over_daily_message_limit,
+    check_service_over_daily_sms_limit,
+)
 from app.types import VerifiedNotification
 from app.utils import get_csv_max_rows
 from app.v2.errors import (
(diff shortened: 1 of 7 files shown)


app/celery/tasks.py
```diff
@@ -298,12 +303,11 @@ def save_smss(self, service_id: Optional[str], signed_notifications: List[Signed
 
     # we should move this check inside the for loop below
     check_service_over_daily_message_limit(KEY_TYPE_NORMAL, service)
-
     current_app.logger.info(f"Sending following sms notifications to AWS: {notification_id_queue.keys()}")
     for notification_obj in saved_notifications:
         try:
-            check_if_request_would_put_service_over_daily_sms_limit(KEY_TYPE_NORMAL, service, 0)  # type: ignore
-        except LiveServiceTooManySMSRequestsError:
+            check_service_over_daily_sms_limit(KEY_TYPE_NORMAL, service)
```
jimleroyer, line 309: Great simple fix. With the goal in mind to avoid potential confusion like this one, could we rename the function to indicate a send might happen so this is clearer? Something like `check_service_over_daily_sms_limit_and_send`? That would be the quick and easy confusion fix.

It looks to me though the long-term ideal fix would be to decouple the checks from the sending though, to have one concern per function (1 function check and 1 function send). The sending function could then be called in the except clause.

app/celery/tasks.py
```diff
@@ -298,12 +303,11 @@ def save_smss(self, service_id: Optional[str], signed_notifications: List[Signed
...
+        except (LiveServiceTooManySMSRequestsError, TrialServiceTooManyRequestsError):
```
jimleroyer, line 310: This was another potential bug where trial services were not covered by this?

smcmurtry: Yes - without the extra error getting caught here, a trial service could have sent above their limit in some cases.

app/celery/tasks.py
smcmurtry: > It looks to me though the long-term ideal fix would be to decouple the checks from the sending
Yes I agree - that would make it much clearer

> could we rename the function to indicate a send might happen so this is clearer
Good idea, I'll make that change now

smcmurtry: I renamed to `check_service_over_daily_sms_limit_and_warn` since that is the language used in some of the other functions like `warn_about_daily_sms_limit`.

PR #1661, opened 2022-10-31
Experiment: Remove Redis from the SMS daily limit logic

smcmurtry:
# Summary | Résumé

Probably a controversial experiment to remove redis from the SMS daily limit logic 👿 

I thought I would try this to see if it fixes the bug where SMS CSV sends are getting stuck in sending before the limit is exceeded. 

This does seem to fix the bug.


smcmurtry:
This is probably not the right direction to go in.

PR #1662, opened 2022-11-01
Move all SMS Limits logic to the "front door" of the API

smcmurtry:
# Summary | Résumé

This PR:
- moves all the SMS Limit logic out of celery and process_notifications.py, moves this to the "front door" of the API to where the post requests come in
- splits the "sms daily limit check" and the "send sms limit emails" logic into 2 functions
- makes the `key_type` arguments optional - often we are just passing in a default value
- adds a `sms_requested` argument to the check function, so that we chan check if the requested send would put the service over its limit
- adds the "sms daily limit check" logic to the "CSV send through admin" codepath, in app/job/rest.py. This will prevent the job from being created if the service is at or over their limit
- unit testing will be added in a separate PR by @andrewleith 

See https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/cds-snc/notification-planning/947

      fonctionnalité existante?

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -79,10 +79,6 @@ from app.notifications.validators import (
 )
 from app.types import VerifiedNotification
 from app.utils import get_csv_max_rows
-from app.v2.errors import (
-    LiveServiceTooManySMSRequestsError,
-    TrialServiceTooManyRequestsError,
-)
 
 
 @notify_celery.task(name="process-job")
@@ -305,7 +301,7 @@ def save_smss(self, service_id: Optional[str], signed_notifications: List[Signed
     check_service_over_daily_message_limit(KEY_TYPE_NORMAL, service)
     current_app.logger.info(f"Sending following sms notifications to AWS: {notification_id_queue.keys()}")
     for notification_obj in saved_notifications:
-        send_warning_sms_limit_emails_if_needed( service)
+        send_warning_sms_limit_emails_if_needed(KEY_TYPE_NORMAL, notification_obj.service)  # type: ignore
 
         queue = notification_id_queue.get(notification_obj.id) or template.queue_to_use()  # type: ignore
         send_notification_to_queue(
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -73,10 +73,7 @@ from app.notifications.process_notifications import (
     persist_notifications,
     send_notification_to_queue,
 )
-from app.notifications.validators import (
-    check_service_over_daily_message_limit,
-    send_warning_sms_limit_emails_if_needed,
-)
+from app.notifications.validators import check_service_over_daily_message_limit
 from app.types import VerifiedNotification
 from app.utils import get_csv_max_rows
 
@@ -301,8 +298,6 @@ def save_smss(self, service_id: Optional[str], signed_notifications: List[Signed
     check_service_over_daily_message_limit(KEY_TYPE_NORMAL, service)
     current_app.logger.info(f"Sending following sms notifications to AWS: {notification_id_queue.keys()}")
     for notification_obj in saved_notifications:
-        send_warning_sms_limit_emails_if_needed(notification_obj.service)
-
         queue = notification_id_queue.get(notification_obj.id) or template.queue_to_use()  # type: ignore
         send_notification_to_queue(
             notification_obj,
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -75,7 +75,7 @@ from app.notifications.process_notifications import (
 )
 from app.notifications.validators import (
     check_service_over_daily_message_limit,
-    check_service_over_daily_sms_limit_and_warn,
+    send_warning_sms_limit_emails_if_needed,
 )
 from app.types import VerifiedNotification
 from app.utils import get_csv_max_rows
@@ -305,11 +305,7 @@ def save_smss(self, service_id: Optional[str], signed_notifications: List[Signed
     check_service_over_daily_message_limit(KEY_TYPE_NORMAL, service)
     current_app.logger.info(f"Sending following sms notifications to AWS: {notification_id_queue.keys()}")
     for notification_obj in saved_notifications:
-        try:
-            check_service_over_daily_sms_limit_and_warn(KEY_TYPE_NORMAL, service)
-        except (LiveServiceTooManySMSRequestsError, TrialServiceTooManyRequestsError):
-            # if notification would put service over limit, don't add it to the queue
-            continue
+        send_warning_sms_limit_emails_if_needed( service)
 
         queue = notification_id_queue.get(notification_obj.id) or template.queue_to_use()  # type: ignore
         send_notification_to_queue(
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -301,7 +301,7 @@ def save_smss(self, service_id: Optional[str], signed_notifications: List[Signed
     check_service_over_daily_message_limit(KEY_TYPE_NORMAL, service)
     current_app.logger.info(f"Sending following sms notifications to AWS: {notification_id_queue.keys()}")
     for notification_obj in saved_notifications:
-        send_warning_sms_limit_emails_if_needed(KEY_TYPE_NORMAL, notification_obj.service)  # type: ignore
+        send_warning_sms_limit_emails_if_needed(notification_obj.service)
 
         queue = notification_id_queue.get(notification_obj.id) or template.queue_to_use()  # type: ignore
         send_notification_to_queue(
--- a/app/job/rest.py
+++ b/app/job/rest.py
@@ -23,8 +23,10 @@ from app.models import (
     JOB_STATUS_CANCELLED,
     JOB_STATUS_PENDING,
     JOB_STATUS_SCHEDULED,
+    KEY_TYPE_NORMAL,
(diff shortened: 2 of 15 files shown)


app/notifications/validators.py
```diff
@@ -99,17 +100,31 @@ def check_service_over_daily_message_limit(key_type: ApiKeyType, service: Servic
...
+    _send_warning_sms_limit_emails_if_needed(service)
 
 
 def check_rate_limiting(service: Service, api_key: ApiKey, template_type: TemplateType):
     check_service_over_api_rate_limit(service, api_key)
     check_service_over_daily_message_limit(api_key.key_type, service)
-    if template_type == "sms":
-        check_service_over_daily_sms_limit_and_warn(api_key.key_type, service)
+    if template_type == SMS_TYPE:
+        check_sms_daily_limit(api_key.key_type, service)
```
andrewleith, line 127: This is causing the current test failures. The params are reversed, should be:
```suggestion
        check_sms_daily_limit(service, api_key.key_type)
```

app/celery/tasks.py
```diff
@@ -305,11 +301,7 @@ def save_smss(self, service_id: Optional[str], signed_notifications: List[Signed
     check_service_over_daily_message_limit(KEY_TYPE_NORMAL, service)
     current_app.logger.info(f"Sending following sms notifications to AWS: {notification_id_queue.keys()}")
     for notification_obj in saved_notifications:
-        try:
-            check_service_over_daily_sms_limit_and_warn(KEY_TYPE_NORMAL, service)
-        except (LiveServiceTooManySMSRequestsError, TrialServiceTooManyRequestsError):
-            # if notification would put service over limit, don't add it to the queue
-            continue
+        send_warning_sms_limit_emails_if_needed(notification_obj.service)
```
andrewleith, line 304: Really this shouldn't be needed when we start incrementing in API (i.e. not in celery), as the email will be triggered when the POST request is made.

andrewleith (changes requested):
This is looking good.  The next crucial piece would be to get the incrementing out of celery, since it happening asynchronously is the cause of some notifications getting stuck in limbo.  

Do you want me to add my tests directly to this branch and we can put it all together? Or would you prefer a separate PR after this one?

app/celery/tasks.py
smcmurtry: Sounds good, I will move this upstream

app/notifications/validators.py
smcmurtry: I gave the `key_type` argument a default value and made it optional, so that's why the order switched here. Arguably that is not worth the trouble.

smcmurtry:
>Do you want me to add my tests directly to this branch and we can put it all together? Or would you prefer a separate PR after this one?

@andrewleith I would prefer a separate PR if you don't mind. Thanks!

app/notifications/validators.py
andrewleith: Right but the args are in the wrong order so this crashes.. unless I'm missing something.

app/notifications/validators.py
```diff
@@ -99,17 +105,88 @@ def check_service_over_daily_message_limit(key_type: ApiKeyType, service: Servic
...
+
+
+def send_warning_sms_limit_emails_if_needed(service: Service, key_type: ApiKeyType = KEY_TYPE_NORMAL):
+    if not current_app.config["FF_SPIKE_SMS_DAILY_LIMIT"]:
+        return
+    if key_type == KEY_TYPE_TEST:
+        return
+    if not current_app.config["REDIS_ENABLED"]:
+        return
+    messages_sent = fetch_daily_sms_fragment_count(service.id)
```
andrewleith, line 138: I think we should add one to represent the current notification being sent.  That way, the user will get the warning when they send their 8th email when their limit is 10 rather than their 9th.  Thoughts?

```suggestion
    messages_sent = fetch_daily_sms_fragment_count(service.id) + 1
```

andrewleith: Nevermind, you already handle this by considering the `requested_sms`!

app/notifications/validators.py
```diff
@@ -99,17 +103,72 @@ def check_service_over_daily_message_limit(key_type: ApiKeyType, service: Servic
...
+def check_sms_daily_limit(service: Service, requested_sms=0):
+    messages_sent = fetch_todays_requested_sms_count(service.id)
+    over_sms_daily_limit = (messages_sent + requested_sms) >= service.sms_daily_limit
+
+    # Send a warning when reaching the daily message limit
+    if not over_sms_daily_limit:
+        return
+
+    current_app.logger.info(
+        f"service {service.id} has been rate limited for daily sms use sent {int(messages_sent)} limit {service.sms_daily_limit}"
```
andrewleith, line 115: We may want to adjust this text slightly, as this isn't really rate limiting and might confuse us in prod support.

```suggestion
        f"service {service.id} is exceeding their daily sms limit [total sent today: {int(messages_sent)} limit: {service.sms_daily_limit}, attempted send: {requested_sms}"
```

app/notifications/validators.py
```diff
@@ -99,17 +103,72 @@ def check_service_over_daily_message_limit(key_type: ApiKeyType, service: Servic
...
+        return
+
+    check_sms_daily_limit(service, requested_sms)
+    increment_todays_requested_sms_count(service.id, requested_sms)
+    send_warning_sms_limit_emails_if_needed(service)
+
+
+def check_rate_limiting(service: Service, api_key: ApiKey):
     check_service_over_api_rate_limit(service, api_key)
     check_service_over_daily_message_limit(api_key.key_type, service)
```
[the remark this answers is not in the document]
andrewleith: My bad, this is the overall message limit check, which is still needed.  

I think we should get it out of the `check_rate_limiting` function though as that is a bit confusing since the overall daily limit is different than rate limiting. 🤔

app/notifications/validators.py
smcmurtry: Good call!

app/notifications/validators.py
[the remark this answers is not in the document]
smcmurtry: Yeah I agree that "rate limiting" is not a good name for this function. Let's fix that in a separate PR though, since it's not directly related to the SMS limits and this PR is already getting big 😅

app/notifications/validators.py
```diff
@@ -99,17 +103,72 @@ def check_service_over_daily_message_limit(key_type: ApiKeyType, service: Servic
     counter_name="rate_limit.live_service_daily_sms",
     exception=LiveServiceTooManySMSRequestsError,
 )
-def check_service_over_daily_sms_limit_and_warn(key_type: ApiKeyType, service: Service):
-    if current_app.config["FF_SPIKE_SMS_DAILY_LIMIT"] and key_type != KEY_TYPE_TEST and current_app.config["REDIS_ENABLED"]:
-        fragments_sent = fetch_daily_sms_fragment_count(service.id)
-        warn_about_daily_sms_limit(service, fragments_sent)
+def check_sms_daily_limit(service: Service, requested_sms=0):
+    messages_sent = fetch_todays_requested_sms_count(service.id)
+    over_sms_daily_limit = (messages_sent + requested_sms) >= service.sms_daily_limit
```
andrewleith, line 108: This defines being over the limit when you are actually at the limit.  It prevents a user from sending their last notification and stops them from getting the email that let's them know they've reached their limit

```suggestion
    over_sms_daily_limit = (messages_sent + requested_sms) > service.sms_daily_limit
```

app/notifications/validators.py
```diff
@@ -99,17 +103,72 @@ def check_service_over_daily_message_limit(key_type: ApiKeyType, service: Servic
...
+    if service.restricted:
+        raise TrialServiceTooManySMSRequestsError(service.sms_daily_limit)
+    else:
+        raise LiveServiceTooManySMSRequestsError(service.sms_daily_limit)
+
+
+def send_warning_sms_limit_emails_if_needed(service: Service):
+    todays_requested_sms = fetch_todays_requested_sms_count(service.id)
+    nearing_sms_daily_limit = todays_requested_sms >= NEAR_DAILY_LIMIT_PERCENTAGE * service.sms_daily_limit
+    over_sms_daily_limit = todays_requested_sms >= service.sms_daily_limit
```
andrewleith, line 126: Technically we should call this `over_or_at_sms_daily_limit`

app/notifications/validators.py
smcmurtry: Hmm good point. I think what we actually want is to block them from sending more than the limit but warn them they are at the limit when they are at the limit. I'll adjust.

app/job/rest.py
```diff
@@ -132,6 +137,15 @@ def create_job(service_id):
     data["template"] = data.pop("template_id")
     template = dao_get_template_by_id(data["template"])
 
+    if template.template_type == SMS_TYPE:
+        job = get_job_from_s3(service_id, data["id"])
+        recipient_csv = RecipientCSV(
+            job,
+            template_type=template.template_type,
+            placeholders=template._as_utils_template().placeholders,
```
andrewleith, line 145: Need to pass in template for the fragment count to be calculated properly here.

```suggestion
            placeholders=template._as_utils_template().placeholders,
            template=Template(template.__dict__),
```

andrewleith (approved):
This is good to go from my point of view.  Good work team! 🥂

PR #1667, opened 2022-11-08
Issue909

ldraney:
# Summary | Résumé

No big changes to the actions, I simply appended "-node16" to the relevant aws actions, and created a file to test actions.  Since almost all of the relevant actions are being called with dev_deploy.yaml anyway, I simply ran that test after my changes and saw that the only change was no warnings on the output.


sastels:
@ldraney  I think you meant to make a PR into https://github.com/department-of-veterans-affairs/notification-api ?

ldraney:
Yes, I did.  Thank you!  I fixed what caused this.

PR #1678, opened 2022-11-17
Fix for reply to bug

smcmurtry:
# Summary | Résumé

This is the API side of the fix for the reply-to bug and UX issues described on the attached card.

The API no longer returns an error when trying to archive the default reply-to address, as long as it is the only one. This allows the user to return to the initial state, before they entered a reply-to address. This makes adding a reply-to address a "reversible" action and solves that UX problem where they are not warned their action was irreversible.

Trying to delete the default reply-to address will still return an error, if other non-archived reply-tos exist. This is to ensure we don't get into a buggy state. If the user clicks to archive a default reply-to in this case, the Admin should warn them and then assign default to one of their other addresses before the deletion occurs.

Test that this still works as expected with the current Admin site.

This should be released before the admin PR: https://github.com/cds-snc/notification-admin/pull/1398

      fonctionnalité existante?

### DIFF
--- a/app/dao/service_email_reply_to_dao.py
+++ b/app/dao/service_email_reply_to_dao.py
@@ -70,7 +70,7 @@ def archive_reply_to_email_address(service_id, reply_to_id):
     if reply_to_archive.is_default:
         non_archived_reply_tos = dao_get_reply_to_by_service_id(service_id)
         if len(non_archived_reply_tos) > 1:
-            # this error should not be displayed in Admin, only if a user manually sends a post request 
+            # this error should not be displayed in Admin, only if a user manually sends a post request
             raise ArchiveValidationError("You cannot delete a default email reply to address if other reply to addresses exist")
         reply_to_archive.is_default = False
 
--- a/app/dao/service_email_reply_to_dao.py
+++ b/app/dao/service_email_reply_to_dao.py
@@ -68,7 +68,11 @@ def archive_reply_to_email_address(service_id, reply_to_id):
     reply_to_archive = ServiceEmailReplyTo.query.filter_by(id=reply_to_id, service_id=service_id).one()
 
     if reply_to_archive.is_default:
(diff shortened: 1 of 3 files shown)


Pensai:
From what I have been able to test it is looking great so far. I will test further once the [other PR](https://github.com/cds-snc/notification-admin/pull/1398) is good to go.

app/dao/service_email_reply_to_dao.py
```diff
@@ -68,7 +68,11 @@ def archive_reply_to_email_address(service_id, reply_to_id):
     reply_to_archive = ServiceEmailReplyTo.query.filter_by(id=reply_to_id, service_id=service_id).one()
 
     if reply_to_archive.is_default:
-        raise ArchiveValidationError("You cannot delete a default email reply to address")
+        non_archived_reply_tos = dao_get_reply_to_by_service_id(service_id)
+        if len(non_archived_reply_tos) > 1:
+            # this error should not be displayed in Admin, only if a user manually sends a post request
+            raise ArchiveValidationError("You cannot delete a default email reply to address if other reply to addresses exist")
```
jimleroyer, line 74: Hmm based on the [last comment on the issue explaining the solution](https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/cds-snc/notification-planning/684), I thought we would allow deletion of the default reply to address if others exists? I must misunderstand the code here, it seems we prevent it if other addresses exist, no?

smcmurtry: On Admin, I am switching the default to be one of the other reply-to addresses before deleting the old default: https://github.com/cds-snc/notification-admin/pull/1398/files#diff-a87e3297d1ea8866e40da7b2861cd7d5d8a604e7847705ebb8d5d455ff6a6877R741-R750

jimleroyer (approved):
LGTM!

PR #1708, opened 2022-12-13
Python 3.10 v2

sastels:
# Summary | Résumé

#1679 but with a working docker image! 🎉  😅 

Basically - in #1679 we were still using `eventlet` - here we switch to `gevent` as was done for admin in https://github.com/cds-snc/notification-admin/pull/1400,

# Test instructions | Instructions pour tester la modification

rebuild environment (this uses Python 3.10) and test with admin.

To test the docker image:
```
docker build -f ci/Dockerfile -t api-3.10 .
docker run -p 6013:6011 api-3.10
```
You should see normal api startup logging. Then navigate to `localhost:6013` to see the status page.
(note: bound to port 6013 here since you might have api's devcontainer running and bound to the normal 6011)

# Release Instructions | Instructions pour le déploiement

We're switching gunicorn workers from eventlet to gevent . Shouldn't make much difference, but would be good to compare perf tests of k8s pods on staging before release. (shouldn't affect lambda)

# Reviewer checklist | Liste de vérification du réviseur

This is a suggested checklist of questions reviewers might ask during their
review | Voici une suggestion de liste de vérification comprenant des questions
que les réviseurs pourraient poser pendant leur examen :

- [ ] Is the code maintainable? | Est-ce que le code peut être maintenu?
- [ ] Have you tested it? | L’avez-vous testé?
- [ ] Are there automated tests? | Y a-t-il des tests automatisés?
- [ ] Does this cause automated test coverage to drop? | Est-ce que ça entraîne
      une baisse de la quantité de code couvert par les tests automatisés?
- [ ] Does this break existing functionality? | Est-ce que ça brise une
      fonctionnalité existante?
- [ ] Does this change the privacy policy? | Est-ce que ça entraîne une
      modification de la politique de confidentialité?
- [ ] Does this introduce any security concerns? | Est-ce que ça introduit des
      préoccupations liées à la sécurité?
- [ ] Does this significantly alter performance? | Est-ce que ça modifie de
      façon importante la performance?
- [ ] What is the risk level of using added dependencies? | Quel est le degré de
      risque d’utiliser des dépendances ajoutées?
- [ ] Should any documentation be updated as a result of this? (i.e. README
      setup, etc.) | Faudra-t-il mettre à jour la documentation à la suite de ce
      changement (fichier README, etc.)?

PR #1679, opened 2022-11-17
Upgrade Python to 3.10

sastels:
# Summary | Résumé

https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/cds-snc/notification-planning/959

Upgrade Python from 3.9 to 3.10

Notes:
- using a branch of utils for now
- need to switch from the library `orderedset` to `ordered_set` since the former does not support 3.10 and is possibly abandonware.

# Test instructions | Instructions pour tester la modification

See if admin still works!

PR #1711, opened 2022-12-14
Python 3.10 v3

sastels:
# Summary | Résumé

Upgrade api to Python 3.10

Following #1679 and #1708 which both had to be reverted 😓 

in #1679 the k8s pod failed to start. This was due to our version of eventlet not supporting Python 3.10. Later versions of eventlet do support 3.10 but they break gunicorn 😞 . So in #1708 we switched from eventlet to gevent and fixed the k8s pod.

However, there was another issue:  the api lambda in #1708 (and #1679) would crash on deploy. This was because the lambda function requires a New Relic layer to allow the connection to New Relic. The layer we were using did not support 3.10, and New Relic hasn't released a 3.10 version of the layer yet.

We forked the[ layer's gitrepo](https://github.com/newrelic/newrelic-lambda-layers) and [made some changes](https://github.com/sastels/newrelic-lambda-layers/tree/feat/python-3.10) that allows us to build a Python 3.10 version of this layer. We use this layer (`newrelic-layer.zip`) in the lambda's docker image of this PR in a similar manner to how the 3.9 layer is used currently in main.

- The lambda function for this PR (well, sha 8205ea2)  has been deployed on staging [here](https://ca-central-1.console.aws.amazon.com/lambda/home?region=ca-central-1#/functions/api-lambda/versions/173?tab=testing). 
- It is not connected to the api gateway, but can be tested using AWS's test button. 
- This allows you to send a simulated request to the lambda handler (you can choose a api gateway template and a path, either one that exists such as `_status` or one that doesn't.
- you can view the response from the lambda, and go to [New Relic](https://one.newrelic.com/nr1-core/lambda/invocations/MjY5MTk3NHxJTkZSQXxOQXwtNzgwNDUyNTc5NzAyODI1NTcyNw?duration=1800000&state=5c990bbe-7ed6-5b5d-9d5f-66ef3e60499c) to see the lambda requests recorded.

None.


.github/workflows/lambda_review.yml
```diff
@@ -0,0 +1,48 @@
+
+name: Build and push lambda image to staging for this PR (temporary)
```
sastels, line 2: 🔥 this before merging. Or maybe just set it not to run, it's kinda handy to have around.

application.py
```diff
@@ -4,26 +4,15 @@
 import os
 
 import newrelic.agent  # See https://bit.ly/2xBVKBH
-import sentry_sdk
```
sastels, line 7: Sentry doesn't work with 3.10 and we're not using it in prod anyway. Discussed IRL and decided to remove it from the app.

ci/Dockerfile.lambda
```diff
@@ -17,6 +17,7 @@ WORKDIR ${TASK_ROOT}
 COPY requirements.txt ${TASK_ROOT}
 RUN set -ex && pip3 install -r requirements.txt
 RUN pip3 install awslambdaric newrelic-lambda
+RUN pip install --no-cache-dir newrelic
```
sastels, line 20: hrm. not sure why I added this. Was it needed or was I just poking things randomly to see if I could get New Relic working? Before merge should see if things work without this line.

tests/app/celery/test_letters_pdf_tasks.py
```diff
@@ -491,7 +491,7 @@ def test_process_letter_task_check_virus_scan_passed(
     source_bucket_name = current_app.config["LETTERS_SCAN_BUCKET_NAME"]
     target_bucket_name = current_app.config[bucket_config_name]
 
-    conn = boto3.resource("s3", region_name="ca-central-1")
```
sastels, line 494: tests now fail if we specify the region here.

tests/app/dao/test_fact_notification_status_dao.py
```diff
@@ -1,7 +1,7 @@
 from datetime import date, datetime, timedelta
+from unittest import mock
 from uuid import UUID
 
-import mock
```
sastels, line 4: this is odd - why did we need to change where we import mock from? investigate whether this is needed.

tests/app/user/test_rest.py
```diff
@@ -1,8 +1,8 @@
 import base64
 import json
+from unittest import mock
 from uuid import UUID
 
-import mock
```
sastels, line 5: Again with the different mock?

tests/app/dao/test_fact_notification_status_dao.py
andrewleith: Looks like unittest.mock is now part of the standard python library:
> mock is now part of the Python standard library, available as [unittest.mock](https://docs.python.org/dev/library/unittest.mock.html) in Python 3.3 onwards.

[Source](https://pypi.org/project/mock/), [Docs](https://docs.python.org/dev/library/unittest.mock.html)

sastels:
hooked PR lambda to api gateway.
- smoke tests passed
- admin working
- new relic getting data
  - invocations 
  - "no row found" errors caused by too fast smoke test polling (expected)

tests/app/dao/test_fact_notification_status_dao.py
sastels: yeah, but odd that it works with 3.9 🙃 
But now if you just have `import mock` we get
```
E   ModuleNotFoundError: No module named 'mock'
```
Maybe `import mock` directly was deprecated for a while and finally removed? All good in any case.

ci/Dockerfile.lambda
sastels: ahh, we added it here when we were trying random things to get New Relic working again 😅 . Ok, let me see if we still need it.

sastels: looks like the app works without it and still sends to New Relic 👍

sastels:
yep, everything still works without that newrelic pip install.

.github/workflows/lambda_review.yml
smcmurtry: I would be in favour of removing this for now and then bringing it back if/when we decide to set up API review apps properly

smcmurtry (changes requested):
Regarding the file `newrelic-layer.zip` - could we just include the unzipped version of that file in our repo so we can track changes to it more easily? Since it is essentially part of our codebase that we are maintaining now.

smcmurtry:
I also checked this out locally and tried out connecting to it with the admin site. Sending email and SMS works, have not found any bugs yet.

smcmurtry:
Okay I didn't realize it was the result of a build process. In that case I agree that it makes sense to keep it in another repo.

smcmurtry (approved):

PR #1712, opened 2022-12-15
Update sending logic to ignore test sends for sms limits

andrewleith:
# Summary | Résumé

This PR updates the sending logic to exclude test sends from services' daily sms limits.  There are 4 places that need to be handled:
1. API one-off
2. API bulk
3. ADMIN one-off
4. ADMIN bulk

And there are some factors to consider:
- API
   1. Test keys 
      - All addresses: should not count toward the limit
   2. Team keys 
      - Test addresses - should not count toward the limit
      - Other addresses - should count towards the limit
   3. Live keys 
      - Test addresses - should not count toward the limit
      - Other addresses - should count towards the limit
- ADMIN
   1. Test addresses - these should not count toward the limit

**Notes:**
- When sending using API Bulk or Admin CSV there exists the possibility of a user mixing testing/non-testing phone numbers.  This is now explicitly disallowed and will return an error from both the API and ADMIN.
- **Except** in the case of an API bulk send with a TEST api key; mixing the numbers is still allowed here

This bug exposed several scenarios that we are currently not handling. New tests have been added for the new scenarios:
- Testing API one-off:
   - Sending using TEST, NORMAL, and TEAM API keys with ALL simulated phone numbers should not count towards limits
- Testing API BULK:
   - Sending using TEST API key with ALL simulated phone numbers should not count towards limits
   - Throw an error if a user mixes testing and non-testing numbers with a LIVE or TEAM key
   - Allow mixing if its a TEST key
- Testing ADMIN one-off:
   - Sending using TEST phone numbers (i.e. +16132532222) should not count towards limits
- Testing ADMIN CSV:
   - Sending using ALL TEST phone numbers (i.e. +16132532222) should succeed and not not count towards limits
   - Sending using test+non-test phone numbers should fail


whabanks:
**Current progress on testing this PR:**
✅  Sending using TEST, NORMAL, and TEAM API keys with ALL simulated phone numbers should not count towards limits
#### **Testing API BULK:**
☑️ Sending using TEST API key with ALL simulated phone numbers should not count towards limits
☑️ Throw an error if a user mixes testing and non-testing numbers with a LIVE or TEAM key
☑️ Allow mixing if its a TEST key
#### **Testing ADMIN one-off:**
✅ Sending using TEST phone numbers (i.e. +16132532222) should not count towards limits
#### Testing ADMIN CSV:
✅ Sending using ALL TEST phone numbers (i.e. +16132532222) should succeed and not not count towards limits
✅ Sending using test+non-test phone numbers should fail

sastels:
Code LGTM - @whabanks are you happy with your testing?

whabanks:
LGTM! Tests pass, no issues from manual testing and prodding.

whabanks (approved):

app/job/rest.py
```diff
@@ -146,7 +147,20 @@ def create_job(service_id):
...
-        check_sms_limit_increment_redis_send_warnings_if_needed(service, recipient_csv.sms_fragment_count)
+        # calculate the number of simulated recipients
+        numberOfSimulated = sum(
+            simulated_recipient(i["phone_number"].data, template.template_type) for i in list(recipient_csv.get_rows())
+        )
+        mixedRecipients = numberOfSimulated > 0 and numberOfSimulated != len(list(recipient_csv.get_rows()))
+
+        # if they have specified testing and NON-testing recipients, raise an error
+        if mixedRecipients:
+            raise InvalidRequest(message="Bulk sending to testing and non-testing numbers is not supported", status_code=400)
```
jimleroyer, line 158: Very nice

app/notifications/validators.py
```diff
@@ -151,13 +150,9 @@ def time_until_end_of_day() -> timedelta:
...
 
 
-def check_sms_limit_increment_redis_send_warnings_if_needed(
-    service: Service, requested_sms=0, key_type: ApiKeyType = KEY_TYPE_NORMAL
-) -> None:
+def check_sms_limit_increment_redis_send_warnings_if_needed(service: Service, requested_sms=0) -> None:
     if not current_app.config["FF_SPIKE_SMS_DAILY_LIMIT"]:
         return
-    if key_type == KEY_TYPE_TEST:
-        return
```
jimleroyer, line 160: We assume at this point that this check is made by the caller?

andrewleith: Yes, I re-worked this logic so that this would not even be called in the case that a test key was used.

PR #1714, opened 2022-12-19
Update the "Generate SBOM" action version

smcmurtry:
# Summary | Résumé

~This action is failing on my latest PR, I want to test if it fails on the main branch as well but need to open a PR to trigger the action.~

Pat has kindly fixed the `CycloneDX` package that was causing this failure. The "Generate SBOM" action is now passing in this PR. We can re-purpose this PR to make the update.


requirements.txt
```diff
@@ -67,8 +67,8 @@ typing-extensions==4.4.0
 
 ## The following requirements were added by pip freeze:
 aiohttp==3.8.3
-aiosignal==1.2.0
```
sastels, line 70: Did the sbom change need these requirements upgrades?

smcmurtry: No, but to trigger the "generate SBOM" action I needed to change requirements.txt, so I ran `make freeze-requirements`.

Now that we know it works, I could remove changes from requirements.txt from this PR if you would like.

smcmurtry: Actually, I'll just go ahead and do that. I don't want to mess up @andrewleith's PR.

sastels: ah, I see. Yeah, I'd say lets remove the requirements changes to keep it just to fixing the SBOM action.

sastels (approved):
:shipit:

PR #1716, opened 2022-12-22
Implement malicious content scanning for attached files

smcmurtry:
# Summary | Résumé

PR to implement malicious content scanning for attached files inside notification-api. This is closely related to https://github.com/cds-snc/notification-document-download-api/pull/106

This PR does the following:
- adds a `check_scan_verdict` method to the `DocumentDownloadClient`, which makes a request to the new endpoint
- adds this check to `send_to_providers.py`, so that the send will not occur if the doc-download-api returns an error which indicates the malware scan verdict of an attached file is `malicious`, `suspicious`, or `in_progress`
- adds an error handling fn `check_for_malware_errors` to update the notification status and raise some new errors that are caught in `provider_tasks.py`. This fn handles the retry behaviour if the verdict is `in_progress`
- added tests to check behaviour is correct for the various cases

1. Check out this branch
2. Add `DOCUMENT_DOWNLOAD_API_HOST` and `DOCUMENT_DOWNLOAD_API_KEY` to your .env and use the staging values
3.  run locally
4. send an email with an attachment file that is not malware via your local API. Make sure it is sent successfully and you see a log like: `scan_verdict for document_id 4968d4e5-dec8-4022-9090-33dcb482aaa0 is {'scan_verdict': 'clean'}`
6. send and email with an EICAR attachment test file that will be detected as malware. Make sure the email is not sent and you see a log like: `Malicious content detected! Download and attachment failed for notification.id: ae96095b-c442-4ae2-a721-8f09247b4727`

Once this is merged and released, the malware scan feature will be live.

      fonctionnalité existante?


tests/app/delivery/test_send_to_providers.py
```diff
@@ -1041,3 +1057,87 @@ def test_is_service_allowed_html(sample_service: Service, notify_api):
...
+        personalisation = {"file": document_download_response()}
+
+        db_notification = save_notification(create_notification(template=sample_email_template, personalisation=personalisation))
+
+        with pytest.raises(DocumentDownloadException) as e:
+            send_to_providers.send_email_to_provider(db_notification)
+            assert db_notification.id in e.value
+        send_mock.assert_not_called()
+
+        assert Notification.query.get(db_notification.id).status == "technical-failure"
```
jimleroyer, line 1143: Do we have a test for when document download does not fail per se, but the scanning did for some reason? What do we do if malware scanning system is down for some reason? 

I added a [related comment on the ADR](https://github.com/cds-snc/notification-adr/pull/30/files#r1106435018) to the effect of what we'd do if the malware scan is down or returns an actual error.

app/celery/provider_tasks.py
```diff
@@ -69,6 +74,8 @@ def deliver_email(self, notification_id):
         current_app.logger.error(f"Cannot send notification {notification_id}, got an invalid direct file url.")
         update_notification_status_by_id(notification_id, NOTIFICATION_TECHNICAL_FAILURE)
         _check_and_queue_callback_task(notification)
+    except (MalwareDetectedException, DocumentDownloadException):
+        _check_and_queue_callback_task(notification)
```
jimleroyer, line 78: When is the `DocumentDownloadException` exception raised? It is new behavior, right? I wonder because before, we'd retry more times via `self.retry` before reporting a failure. Is this exception raised and cannot be recovered at all?

app/delivery/send_to_providers.py
```diff
@@ -128,8 +135,32 @@ def check_file_url(file_info: Dict[str, str], notification_id: UUID):
...
+        # Throw error so celery will retry in sixty seconds
+        raise MalwareScanInProgressException
+
+    if document_download_response_code != 200:
+        document_download_internal_error(notification=notification)
+
+
 def send_email_to_provider(notification: Notification):
-    current_app.logger.info(f"Sending email to provider for notification id {notification.id}")
+    current_app.logger.info(f"Sending email to provider for notification id: {notification.id}")
```
jimleroyer, line 163: A side note that we should probably check our existing queries in CloudWatch and TF when we change logs. Probably fine here I hope.

app/delivery/send_to_providers.py
```diff
@@ -128,8 +135,32 @@ def check_file_url(file_info: Dict[str, str], notification_id: UUID):
         raise InvalidUrlException
 
 
+def check_for_malware_errors(document_download_response_code, notification):
```
jimleroyer, line 138: I love small dedicated functions.

app/delivery/send_to_providers.py
```diff
@@ -146,12 +177,16 @@ def send_email_to_provider(notification: Notification):
...
+            scan_verdict_response = document_download_client.check_scan_verdict(service.id, document_id, sending_method)
+            check_for_malware_errors(scan_verdict_response.status_code, notification)
+            current_app.logger.info(f"scan_verdict for document_id {document_id} is {scan_verdict_response.json()}")
             if sending_method == "attach":
                 try:
-
-                    req = urllib.request.Request(personalisation_data[key]["document"]["direct_file_url"])
+                    req = urllib.request.Request(direct_file_url)
                     with urllib.request.urlopen(req) as response:
+                        check_for_malware_errors(response.getcode(), notification)
```
jimleroyer, line 189: I am confused why we call `check_for_malware_errors` again and with the status code of the file download link. Maybe  talking offline would make this faster with this review.

app/delivery/send_to_providers.py
```diff
@@ -128,8 +135,32 @@ def check_file_url(file_info: Dict[str, str], notification_id: UUID):
...
+        malware_failure(notification=notification)
+
+    # 428 "Precondition Required" response is sent if the scan is still in progress
+    if document_download_response_code == 428:
+        current_app.logger.info(f"Malware scan in progress, could not download files for notification.id: {notification.id}")
+        # Throw error so celery will retry in sixty seconds
+        raise MalwareScanInProgressException
+
+    if document_download_response_code != 200:
+        document_download_internal_error(notification=notification)
```
jimleroyer, line 159: Based on parent's logic flow, this would quickly set a failure on the notification status in the database, without retry strategies. Let's talk if I understand this correctly and if yes, why we might want to give retries a try (pun not intended) for resiliency purpose.

jimleroyer (commented):
Love the test, very clear. I have a few questions on the logic, especially for retries and failure handling.

app/delivery/send_to_providers.py
smcmurtry: Good point - I did not need to make that log change! I'll revert.

app/delivery/send_to_providers.py
smcmurtry: I had intended this as an extra check for safety, since requesting to download a file tagged malware will also return an error code once that is implemented over on the dd-api.

But I agree it is confusing to have the check twice and I can't think of a case where the first check would pass and the 2nd check would fail. So I'll just remove this.

app/delivery/send_to_providers.py
smcmurtry: Yes you are understanding that correctly. I intended this error to happen when the dd-api has a `DocumentStoreError` which results in a 404: https://github.com/cds-snc/notification-document-download-api/blob/main/app/download/views.py#L131-L139

That happens when there is a `BotoClientError`: https://github.com/cds-snc/notification-document-download-api/blob/dce41bbcc5090d07552aab14677e56c34f036864/app/utils/store.py#L126

My guess is that should only happen if the object is not in S3, in which case I don't think we will recover from the error. But I don't know that for sure.

So I think we could either:
a) target this just to 404 errors, or 
b) let the retry logic happen for this error.

What do you think?

tests/app/delivery/test_send_to_providers.py
smcmurtry: If the scanning doesn't happen, then the `av-status` key will not be written to the object in S3 and dd-api will have a `KeyError` which is handled here: https://github.com/cds-snc/notification-document-download-api/blob/dce41bbcc5090d07552aab14677e56c34f036864/app/utils/store.py#L128-L129

So it will look like the scan is "in progress" and the retry logic will handle the error. So in that case we would retry until we reach the max number of retries. 

If the malware scan verdict comes back as `{"av-status": "error"}` or `"unknown"` or `"unable_to_scan"` then the code treats that the same as a `"clean"` verdict. But the `av-status` gets logged so I was thinking we could make a cloudwatch query or alarm based on that log.

andrewleith (commented):
Can this s3 bucket be removed now? 
- https://s3.console.aws.amazon.com/s3/buckets/notification-canada-ca-staging-document-download-scan-test?region=ca-central-1

smcmurtry:
> Can this s3 bucket be removed now?

Yes thanks for the reminder, I just deleted it.

app/delivery/send_to_providers.py
smcmurtry: I guess we don't really know under what circumstances this error will happen. I will change the error handling so this gets retried as per Jimmy's suggestion since that is safer.

jimleroyer (approved):
Awesome work.

PR #1721, opened 2023-01-09
Add feature flag for Freshdesk integration

whabanks:
# Summary | Résumé
Add a feature flag to disable Freshdesk integration. This will help prevent us from hammering Freshdesk with tickets during tasks such as pen testing.

1. Set `FRESH_DESK_ENABLED` to `False` in `.env`.
2. Submit a ticket via the `contact us` page 
3. Check Freshdesk to ensure that a ticket was not created.

None.

      fonctionnalité existante?


app/clients/freshdesk.py
```diff
@@ -80,7 +80,7 @@ def _generate_ticket(self) -> Dict[str, Union[str, int, List[str]]]:
     def send_ticket(self) -> int:
         try:
             api_url = current_app.config["FRESH_DESK_API_URL"]
-            if not api_url:
+            if not api_url or current_app.config["FRESH_DESK_ENABLED"] is False:
```
jzbahrai, line 83: what happens in this case? Does this throw a warning in #notification-ops?
I'm not sure if we want to raise an error for the second half - but maybe just a continue?

jzbahrai (commented):
I see you added the config to conftest, can we also add a test?

app/clients/freshdesk.py
whabanks: We can definitely add tests for this! 
I wasn't sure about the placement of the check here. Originally I had it wrapping the post and just returning a 201 if it was disabled, since the implementation is disabled, it also seemed appropriate to throw a `NotImplementedError`. I didn't see any messages in #notification-ops but if that's a possibility then we should avoid it. This type of flag is only something we'd use in development correct, we wouldn't see use in prod?

jzbahrai: ya if you can add a test that would be great, if its cumbersome dw about it

tests/app/clients/test_freshdesk.py
```diff
@@ -220,3 +221,13 @@ def match_json(request):
                 )
             ).send_ticket()
             assert response == 201
+
+
+def test_send_ticket_freshdesk_integration_disabled(mocker, notify_api: Flask):
+    with set_config_values(notify_api, {"FRESH_DESK_ENABLED": "False"}):
```
whabanks, line 227: Removed the env var set in `conftest.py` since we've got a default value already in `config.py` and this test is a bit of a one off.

jzbahrai (approved):

PR #1722, opened 2023-01-10
Change warning and exception logger for retries

jzbahrai:
# Summary | Résumé

We don't want the API to throw an error for one retry. We will throw a warning when there have been more than 10 retries.

This is from where I assume we have retry data:
https://github.com/celery/celery/blob/main/celery/app/task.py#L704

### DIFF
--- a/app/celery/provider_tasks.py
+++ b/app/celery/provider_tasks.py
@@ -76,7 +76,10 @@ def deliver_email(self, notification_id):
     except Exception as e:
         try:
             current_app.logger.warning(f"The exception is {repr(e)}")
-            current_app.logger.exception("RETRY: Email notification {} failed".format(notification_id))
+            if self.request.retries <= 10:
+                current_app.logger.warning("RETRY {}: Email notification {} failed".format(self.request.retries, notification_id))
+            else:
+                current_app.logger.exception("RETRY: Email notification {} failed".format(notification_id))
             self.retry(queue=QueueNames.RETRY)
         except self.MaxRetriesExceededError:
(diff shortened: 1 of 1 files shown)


app/celery/provider_tasks.py
```diff
@@ -76,7 +76,10 @@ def deliver_email(self, notification_id):
     except Exception as e:
         try:
             current_app.logger.warning(f"The exception is {repr(e)}")
-            current_app.logger.exception("RETRY: Email notification {} failed".format(notification_id))
+            if self.request.retries <= 10:
+                current_app.logger.warning("RETRY {}: Email notification {} failed".format(self.request.retries, notification_id))
+            else:
+                current_app.logger.exception("RETRY: Email notification {} failed".format(notification_id))
```
sastels, line 82: yeah, this is good. I was thinking that we could just let the `MaxRetriesExceededError` cause an error, but we set `max_retries` to 48, and maybe we want to see an error before then 👍

sastels: I was thinking that we should add some tests to make sure this code doesn't cause a problem, but we already have a test around "can't find the notification in the database, just retry it" that will run this new code.
https://github.com/cds-snc/notification-api/blob/main/tests/app/celery/test_provider_tasks.py#L80-L90

sastels (approved):
:shipit:

app/celery/provider_tasks.py
jzbahrai: thanks! ill merge

PR #1723, opened 2023-01-10
feat: update generate SBOM and Docker scan

patheard:
# Summary
Add a nightly Docker vulnerability scan and update the generate SBOM action to only create an SBOM for the api Docker image.

GitHub is now automatically indexing and scanning our manifests so the Dependency-Track SBOM generate workflow is no longer needed.

# Related
- cds-snc/platform-core-services#204


andrewleith (approved):
LGTM 🚀

PR #1724, opened 2023-01-11
fix: take service name into account when checking sms message length

andrewleith:
# Summary | Résumé

This PR updates the sms validator which checks the overall message length to include the length of service name which we always append to every sms message.

### DIFF
--- a/app/notifications/validators.py
+++ b/app/notifications/validators.py
@@ -317,8 +317,10 @@ def validate_and_format_recipient(
         return validate_and_format_email_address(email_address=send_to)
 
 
-def check_sms_content_char_count(content_count):
-    if content_count > SMS_CHAR_COUNT_LIMIT:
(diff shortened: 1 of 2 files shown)


app/notifications/validators.py
```diff
@@ -317,8 +317,10 @@ def validate_and_format_recipient(
         return validate_and_format_email_address(email_address=send_to)
 
 
-def check_sms_content_char_count(content_count):
-    if content_count > SMS_CHAR_COUNT_LIMIT:
+def check_sms_content_char_count(content_count, service_name):
+    if (
+        content_count + len(service_name) + 2 > SMS_CHAR_COUNT_LIMIT
+    ):  # the +2 is to account for the ': ' that is added to the service name
```
jzbahrai, line 323: hahaha thats for the comment I was just going to comment and ask what is the +2 for

jzbahrai (approved):

PR #1725, opened 2023-01-11
Fix/enforce sms max limit part 2

andrewleith:
# Summary | Résumé

This PR adds the too-long-sms check to the API bulk sending, and adds a few more tests.

When the bulk send fails, contextual information will be returned to the user so they can troubleshoot which notification has the error.  It will return a message like:
```
{
    "errors": [
        {
            "error": "BadRequestError",
            "message": "Content for template has a character count greater than the limit of 612. Notification to +16132532222 on row #1 exceeds the maximum size limit."
        }
    ],
    "status_code": 400
}
```


jzbahrai (approved):

app/v2/notifications/post_notifications.py
```diff
@@ -182,6 +182,15 @@ def post_bulk():
 
     check_for_csv_errors(recipient_csv, max_rows, remaining_messages)
 
+    for row in recipient_csv.get_rows():
+        try:
+            validate_template(template.id, row.personalisation, authenticated_service, template.template_type)
+        except BadRequestError as e:
```
jzbahrai, line 188: not for this PR but we should have made another exception for this - except TooLongSMSError or somethng

PR #1726, opened 2023-01-12
Fix/enforce sms limit with prefix sms option

andrewleith:
# Summary | Résumé

This builds on #1724 and #1725 to take into account the `sms_prefix` setting of a service which controls whether or not the service name is including in the content of an SMS message.

Test scenarios: 
- From a service with `sms_prefix` on, messages whose length are greater than 610 - the length of the service name should provide an error message in both the API and in ADMIN
- From a service with `sms_prefix` off, messages whose length are greater than 612 should provide an error message in both the API and in ADMIN

### DIFF
--- a/app/notifications/validators.py
+++ b/app/notifications/validators.py
@@ -317,10 +317,12 @@ def validate_and_format_recipient(
         return validate_and_format_email_address(email_address=send_to)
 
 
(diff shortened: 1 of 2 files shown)


sastels (approved):
tested locally! :shipit:

PR #1728, opened 2023-01-12
add a new test number

sastels:
# Summary | Résumé

Add a test number (613-555-0123) that is treated as a normal number everywhere in the app **except** the corresponding notification  is not sent to AWS for delivery.

We have test api keys and simulated numbers for users to test their usage of our api, but they are not useful for us to test the system, since:
- simulated numbers just return a 201 response and do not create a notification
- test keys use the research queue regardless of the priority of the notification

The new number will allow us to test everything just up to actually sending the SMS.

send some SMS to this number. You should see a celery log message similar to
```
notification 242d485e-e4e4-4bd1-b66b-d4242ed1fbaa sending to internal test number. Not sending to AWS
```
and you should **not** see a celery log message similar to
```
AWS SNS request finished in 2.1
```

Note that phone numbers 555-0100 through 555-0199 have been reserved across all area codes for movies and tv shows to use and are not assigned.


jzbahrai (approved):

andrewleith (approved):
This is a great idea and will come in handy for sure.  Nice work! 

Tested locally, works as advertised! 🚀

PR #1733, opened 2023-01-17
Migrate to poetry

whabanks:
# Summary | Résumé
Initial migration to Poetry.

-  Project Virtual Environments are now fully and automatically managed by Poetry
- `Dockerfiles` and Github actions updated to leverage Poetry 
- `make freeze-requirements` is replaced by: `poetry lock`
- `make test-requirements` is replaced by: `poetry lock --check`
- `requirements*.txt` files have been replaced with their Poetry counterparts:
    -  `requirements.txt` -> `pyproject.toml` 
    -   `requirements_for_test.txt` -> `pyproject.toml` 
    - `requirements-app.txt` -> `poetry.lock`
- Added [Better TOML](https://marketplace.visualstudio.com/items?itemName=bungcip.better-toml) extension to the dev container for sanity when working with `pyproject.toml`

## TODO:
- [ ] Set up coveralls for the `coverage` make task

- [ ] Dev container can be built
- [ ] All `Dockerfiles` can be built with `docker build` commands
- [ ] Github actions do not fail
- [ ] `Makefile` tasks run without issue
- [ ] post devcontainer build, terminal instances should automatically put the user into the appropriate venv ( indicated by the shell prompt being prefixed with the venv name)
    - [ ] Poetry commands work: `poetry install`, `poetry show --tree`, `poetry env info`, `poetry lock --check`
- [ ] [Use either a local or git branch reference to utils](https://github.com/cds-snc/notification-utils/pull/168)
- [ ] Functionality of the running application is unchanged


app/celery/tasks.py
```diff
@@ -269,7 +269,7 @@ def save_smss(self, service_id: Optional[str], signed_notifications: List[Signed
 
         verified_notifications.append(notification)
         notification_id_queue[notification_id] = notification.get("queue")  # type: ignore
-        process_type = template.process_type
+        process_type = str(template.process_type)
```
whabanks, line 272: Kept running into an incompatible type error from MyPy:

```
app/celery/tasks.py:403: error: Argument "priority" to "put_batch_saving_bulk_processed" has incompatible type "Union[Any, MapperProperty[Any], ColumnElement[Any]]"; expected "Optional[str]"
```
In [models.py](https://github.com/cds-snc/notification-api/blob/be4177e240b374e2520cb75bc6957fd26d1925ea/app/models.py#L1067) it appears that `@declared_attr` annotated properties aren't inferred by MyPy without [jumping through hoops](https://docs.sqlalchemy.org/en/14/orm/extensions/mypy.html#using-declared-attr-and-declarative-mixins). It may have something to do with invariance vs covariance as [described in MyPy documentation](https://mypy.readthedocs.io/en/stable/common_issues.html#invariance-vs-covariance). 

`sqlalchemy-stubs` and `sqlalchemy2-stubs` are supposed to provide us with better type inference for SQLAlchemy related types. However enabling the plugin in `mypy.ini` seems to have no effect.

We should revisit some of the typing we do to avoid str() calls just to satisfy MyPy.

.devcontainer/scripts/notify-dev-entrypoint.sh
```diff
@@ -28,18 +28,28 @@ echo -e "complete -F __start_kubectl k" >> ~/.zshrc
...
+
+pip install poetry==${POETRY_VERSION} \
+  && poetry --version
+
+# Initialize poetry autocompletions
+mkdir ~/.zfunc
+touch ~/.zfunc/_poetry
+poetry completions zsh > ~/.zfunc/_poetry
+
+# Warm up git index prior to display status in prompt else it will
```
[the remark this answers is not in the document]
whabanks: Removed!

pyproject.toml
```diff
@@ -2,3 +2,105 @@
...
+
+[tool.poetry.scripts]
+notify-api = ""
+
+[build-system]
+requires = ["poetry>=1.3.2"]
+build-backend = "poetry.core.masonry.api"
+
+[tool.poetry.dependencies]
+python = "^3.10.9"
```
jimleroyer, line 23: We might want the tilde instead of the caret, for accepting patches and nothing more. The caret could upgrade from 3.10 to 3.11 [if read the doc correctly](https://python-poetry.org/docs/dependency-specification/).

```suggestion
python = "~3.10.9"
```

scripts/run_tests.sh
```diff
@@ -20,7 +20,7 @@ function display_result {
   fi
 }
 
-make test-requirements
+poetry lock --check
```
jimleroyer, line 23: Keeping the `Makefile` entry would prevent much changes such as these, as this is used as an abstraction for most targets.

README.md
```diff
@@ -221,11 +204,11 @@ __Solution__: Do not specify a database in your `.env`
...
 
-__Solution__: Ensure `SQLALCHEMY_DATABASE_URI` supplied in pytest.ini or your `.env` file is valid to your 
+__Solution__: Ensure `SQLALCHEMY_DATABASE_URI` supplied in pytest.ini or your `.env` file is valid to your
 local database with user access, (pytest.ini takes precedence)
 
 ---
 
 __Problem__: Messages are in the queue but not sending
 
-__Solution__: Check that `celery` is running.
```
jimleroyer, line 231: Thanks for the cleanup. 🧹

jimleroyer (commented):
That looks real nice. The one blocker for an approval to me is the Python version might bump up of minor version (which is a lot in the Python world).

app/celery/tasks.py
smcmurtry: I think we may need to migrate to sqlalchemy 2.0 to improve how the typing is working - it looks like that adds better support for `Column` objects: https://docs.sqlalchemy.org/en/20/changelog/whatsnew_20.html#new-typing-support-in-core-and-orm-stubs-extensions-no-longer-used

smcmurtry: Does adding a `# type ignore` comment to that line instead of `str()` fix the error? Then we wouldn't have to change the code.

Makefile
```diff
@@ -19,27 +19,6 @@ generate-version-file: ## Generates the app version file
...
-	$$(pwd)/venv-freeze/bin/pip freeze -r <(sed '/^--/d' requirements-app.txt) | sed -n '/The following requirements were added by pip freeze/,$$p' >> requirements.txt
-	rm -rf venv-freeze
-
-.PHONY: test-requirements
-test-requirements:
-	@diff requirements-app.txt requirements.txt | grep '<' \
-	    && { echo "requirements.txt doesn't match requirements-app.txt."; \
-	         echo "Run 'make freeze-requirements' to update."; exit 1; } \
-|| { echo "requirements.txt is up to date"; exit 0; }
-
```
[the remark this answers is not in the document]
whabanks: This makes sense, I hadn't considered the aspect of carrying out common tasks without knowing the underlying tooling as a point of convenience. I will restore this targets and update them.

pyproject.toml
whabanks: Agreed, good catch.

scripts/run_tests.sh
whabanks: Agreed, I'll restore these make targets.

pyproject.toml
```diff
@@ -2,3 +2,105 @@
...
+name = "notification-api"
+version = "0.1.0"
+description = "Public-facing REST API for Notification built on the GOV.UK Notify platform."
+authors = ["Canadian Digital Service"]
+license = "MIT license"
+readme = "README.md"
+packages = []
+
+[tool.poetry.scripts]
+notify-api = ""
```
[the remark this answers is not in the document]
jimleroyer: Hmm interesting, maybe we could go the other way of deprecating Makefile and have targets in poetry instead.

Makefile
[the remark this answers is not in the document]
jimleroyer: Thanks! we could test if we could move everything to poetry afterward, to see if we can make targets in there and rely on Python instead of bash. Just throwing out the idea for the future if we like this idea.

app/celery/tasks.py
whabanks: Removing the `str()` and putting `# type: ignore` on L290 and L403 seem to resolve the issue. We could get some type-ception going on here and use the typed ignore `# type: ignore[arg-type]`.

scripts/run_tests.sh
whabanks: Restored here, as well as in `DDL-API`, and `admin`.

pyproject.toml
whabanks: Fixed - as well in `admin` and `DDL-API`.

Makefile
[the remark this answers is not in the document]
whabanks: Restored, as well as in `admin` and `DDL-API`.

jimleroyer (approved):
Okay let's test this in the CI pipeline. Hope all will be well. 🤞

PR #1734, opened 2023-01-18
Update french product name in system templates

andrewleith:
# Summary | Résumé

This PR updates all of the GC Notify system templates to correct the french product name, as well as establishes a baseline for these templates in case we ever need to re-create an environment from scratch.

- Ensure that the templates being updated have the correct content
- Ensure there is no instances of "GC Notification" in any system templates

### DIFF
--- a/migrations/versions/0425_update_system_templates.py
+++ b/migrations/versions/0425_update_system_templates.py
@@ -1,0 +1,273 @@
+"""
+
+Revision ID: 0425_update_system_templates
+Revises: 0424_sms_templates_in_redacted
+Create Date: 2022-09-21 00:00:00
+
+"""
+from datetime import datetime
+
+from alembic import op
+from flask import current_app
+
+revision = "0425_update_system_templates"
+down_revision = "0424_sms_templates_in_redacted"
+
+templates = [
+    {
+        "id": current_app.config["ALREADY_REGISTERED_EMAIL_TEMPLATE_ID"],
+        "name": "Your Notify account",
+        "template_type": "email",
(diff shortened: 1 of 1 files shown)


sastels:
there are 26 templates in the PR and 29 in [this list](https://github.com/cds-snc/notification-api/blob/main/app/config.py#L228-L256), are there some we don't have to update?

andrewleith:
> there are 26 templates in the PR and ~29~ 27 in [this list](https://github.com/cds-snc/notification-api/blob/main/app/config.py#L228-L256), are there some we don't have to update?
> 
Yeah I [mentioned the missing one in slack](https://gcdigital.slack.com/archives/CV38DBNVA/p1674058941326329), its called `CONTACT_US_TEMPLATE_ID` and doesn't actually appear to be used, nor is it even a template in the GC Notify service 😬 
> hrm, some of them don't have "Notify" in them, that's probably it?
> 
> Also, running the query
> 
> ```
> select count(*) from templates
> where service_id = 'd6aa2c68-a2d9-4437-ab19-3ae8eb202553'
> and content like '%GC Notification%'
> ```
> 
> I see 32 Notify templates with "GC Notification" in them.
I saw the same, I left those one-off ones alone, since they aren't part of the system templates.  I'll go update those manually though just to be thorough!

andrewleith:
Oh, and just to mention, I included all the system templates even if they didn't mention GC Notification so we would have a baseline of content for all.

sastels (approved):
Replies to comments make sense, looks good to go!

PR #1737, opened 2023-01-23
fix/do not log email address

sastels:
# Summary | Résumé

- stop logging the email address when there's a problem
- if there is a problem, check if the email address contains non-ascii characters.

- Add "+é" to the local part of your CDS email address (before the @) and try to send.
- should see the new error logged and not the email address.


whabanks (approved):
LGTM! Logs are writing as expected:
image

PR #1740, opened 2023-01-25
Use correct api key type when checking if a service is over daily limit

smcmurtry:
# Summary | Résumé

We were hard-coding the api key type to be `KEY_TYPE_NORMAL`, which caused a bug where "test" api key sending caused an error in celery if the service was over it's daily limit. 

I have now tested this on both the `/email` and `/sms` endpoints and I am no longer seeing any celery errors when sending with a Test API key after the service has reached its daily limit.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -412,9 +412,9 @@ def save_emails(self, _service_id: Optional[str], signed_notifications: List[Sig
         # service is whatever it was set to last in the for loop above.
         # at this point in the code we have a list of notifications (saved_notifications)
         # which could be from multiple services
-        check_service_over_daily_message_limit(KEY_TYPE_NORMAL, service)
         research_mode = service.research_mode  # type: ignore
         for notification_obj in saved_notifications:
+            check_service_over_daily_message_limit(notification_obj.key_type, service)
             queue = notification_id_queue.get(notification_obj.id) or template.queue_to_use()  # type: ignore
             send_notification_to_queue(
                 notification_obj,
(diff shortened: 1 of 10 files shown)


app/celery/tasks.py
```diff
@@ -412,9 +412,9 @@ def save_emails(self, _service_id: Optional[str], signed_notifications: List[Sig
         # service is whatever it was set to last in the for loop above.
         # at this point in the code we have a list of notifications (saved_notifications)
         # which could be from multiple services
-        check_service_over_daily_message_limit(KEY_TYPE_NORMAL, service)
         research_mode = service.research_mode  # type: ignore
         for notification_obj in saved_notifications:
+            check_service_over_daily_message_limit(notification_obj.key_type, service)
```
jimleroyer, line 417: Should we have a default to `KEY_TYPE_NORMAL` just in case? Can we guarantee this will always be there?

smcmurtry: We are defaulting to `KEY_TYPE_NORMAL` here: https://github.com/cds-snc/notification-api/blob/bug/test-key-sending-blocked/app/celery/tasks.py#L366

So I think the `key_type` will always be there.

andrewleith (commented):
The key type is also hard coded in the `save_smss` function here: https://github.com/cds-snc/notification-api/blob/main/app/celery/tasks.py#L298

smcmurtry:
Good catch @andrewleith! I even left a comment above that line a while ago. I will move that inside the for loop.

app/celery/tasks.py
jimleroyer: Thanks @smcmurtry . I very much like the new SignedNotification and VerifiedNotification types btw.

andrewleith (approved):
LGTM

smcmurtry:
Oh I actually included these changes in this (now merged) PR since it was touching the same lines: https://github.com/cds-snc/notification-api/pull/1744

So this PR can be closed since the fix has been made. (I probably should have merged this 1st to make the changes easier to track)

PR #1744, opened 2023-01-27
Catch and log the "over daily limit" errors if they are raised in Celery

smcmurtry:
# Summary | Résumé

This fixes the issue where errors related to the daily limit are going uncaught if they are raised from Celery, which can cause our on-call alarms to be triggered. 

These errors are also raised as part of the request cycle, and they create the error response that the user sees if they are over their daily limit. These errors are only raised in celery if the user makes more requests than allowed by their daily limit (these would have to happen quickly). This may only be a problem for email, but I applied the fix to the `save_smss` function as well, just to be safe.

I also needed to split out a new function since flake 8 was complaining that `save_notifications` was too complex.

This should be merged after https://github.com/cds-snc/notification-api/pull/1740

Here is the test that I ran:

Test first using prod:
1. Set your service's daily message limit to 5
2. Using a Live API key, Send 10 one-off emails in quick succession
3. These should all receive `201` success responses
4. Look at your celery output - you should see some uncaught errors like `LiveServiceTooManyRequestsError` or `TrialServiceTooManyRequestsError`, depending on what kind of service you are using

Then try these steps again on this branch and verify that these errors appear as `info` level logs rather than uncaught errors:
```
[2023-01-27 21:57:56,734: INFO/ForkPoolWorker-3] Exceeded send limits (20) for today: Email 3fcfb0de-38fd-49a0-85e1-ba7638f5bc6f not created
```

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -76,7 +76,12 @@ from app.notifications.process_notifications import (
 from app.notifications.validators import check_service_over_daily_message_limit
 from app.types import VerifiedNotification
 from app.utils import get_csv_max_rows
-
+from app.v2.errors import (
+    LiveServiceTooManyRequestsError,
+    LiveServiceTooManySMSRequestsError,
+    TrialServiceTooManyRequestsError,
+    TrialServiceTooManySMSRequestsError
+)
 
 @notify_celery.task(name="process-job")
 @statsd(namespace="tasks")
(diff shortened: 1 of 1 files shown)


andrewleith (approved):
Nice! LGTM, nice to prevent an alarm from triggering here since there is no action for us to take.

PR #1747, opened 2023-02-03
Chore/types in celery

smcmurtry:
# Summary | Résumé

This work isn't part of an existing issue, it's just part of my ongoing learning about static typing in Python.

This PR replaces the `NotRequired` keys on the `NotificationDictToSign` type with `Optional` keys. The difference is that now the keys will have to be there, but they can have a value of `None`. This means we will have a known set of keys on objects of type `NotificationDictToSign`, and we can start to remove the `_notification.get(key, None)` logic and replace with `_notification["key"]`. This will simplify a lot of logic and make our Celery code more robust.

Verify the app works as before.


smcmurtry:
Will open a new PR if I get around to working on this again.

PR #1745, opened 2023-02-02
Add logging for synchronous and asynchronous email replies from SES

andrewleith:
# Summary | Résumé

This PR adds logging to capture SES responses during synchronous and asynchronous communication.

- Synchronous email send in `AwsSesClient.send_email()`
- Asynchronous response in `process_ses_results()` celery task

PR #1748, opened 2023-02-06
Revert "Add logging for synchronous and asynchronous email replies from SES"

andrewleith:
Reverts cds-snc/notification-api#1745

I was able to see some interesting information from this but it does log the sending to email address so I'll revert it.


jimleroyer (approved):
LGTM

PR #1758, opened 2023-02-20
Load test sending file attachments

sastels:
# Summary | Résumé

- Add a load test sending file attachments.
- Add more to readme.

play with the locust tests


andrewleith (approved):
Works beautifully, nicely done!

2 questions:
1) Should we save the template ids and api keys to lastpass so each dev who runs these doesn't have to go create new templates (or find one) and api keys?
2) This may be due to my lack of understanding of poetry, but for me to get this to run I need to do `poetry run locust -f ...` and not `locust -f ...`.  Should the README reflect that, or is the correct end state not to have to prefix it with `poetry run` ?

sastels:
> Works beautifully, nicely done!
> 
> 2 questions:
> 
> 1. Should we save the template ids and api keys to lastpass so each dev who runs these doesn't have to go create new templates (or find one) and api keys?
> 2. This may be due to my lack of understanding of poetry, but for me to get this to run I need to do `poetry run locust -f ...` and not `locust -f ...`.  Should the README reflect that, or is the correct end state not to have to prefix it with `poetry run` ?

1. very good idea.
2. yes, good catch!

PR #1759, opened 2023-02-20
Add a 408 response code case

smcmurtry:
# Summary | Résumé

This PR does the following:
- add a 408 response code case, since we want to send that from dd-api when the scan times out
- added a new test for the "scan timed out" case
- target only python 3.10 in our black config so it will not fail on match statements (was more relevant for an earlier iteration of this PR, but though I'd keep it anyway)

To test:
- run this locally along with this branch of the dd-api: https://github.com/cds-snc/notification-document-download-api/pull/109
- modify the dd-api to return the 408 / "scan timed out" response by: a) adding`raise ScanInProgressError("Content scanning is in progress")` to https://github.com/cds-snc/notification-document-download-api/blob/main/app/utils/store.py#L112, and b) modifying [this timeout](https://github.com/cds-snc/notification-document-download-api/blob/main/app/download/views.py#L24) to be 60s: `SCAN_TIMEOUT_SECONDS = 1 * 60`)
- verify that the email and attachment are sent as expected after about 1 minute when the timeout is reached.


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.black]
 line-length = 130
-target-version = ['py37', 'py38', 'py39', 'py310']
+target-version = ['py310']
```
jimleroyer, line 3: Nice!

app/delivery/send_to_providers.py
```diff
@@ -141,22 +141,29 @@ def check_for_malware_errors(document_download_response_code, notification):
...
+            # Update notification that it contains malware
+            malware_failure(notification=notification)
+        # 428 "Precondition Required" response is sent if the scan is still in progress
+        case 428:
+            current_app.logger.info(f"Malware scan in progress, could not download files for notification.id: {notification.id}")
+            # Throw error so celery will retry
+            raise MalwareScanInProgressException
+        # 408 "Request Timeout" response is sent if the scan does is not complete before it times out
+        case 408:
+            current_app.logger.info(f"Malware scan timed out for notification.id: {notification.id}, send anyway")
```
jimleroyer, line 160: We might already have it, but can we have an alarm (in a TF PR) to tell us when scan-files times out?

jimleroyer: CC @andrewleith over here

sastels (approved):
LGTM!

PR #1765, opened 2023-02-27
Fix/smoke test running with poetry

andrewleith:
# Summary | Résumé

This PR updates the makefile to run the smoke tests without refering to poetry and the devconatiner script to run them using poetry.


sastels (approved):
Let's give this a try!

PR #1766, opened 2023-03-01
Task/calculate bounce

jzbahrai:
# Summary | Résumé

Add functions to calculate bounce rate. Two functions added:
1. Find the bounce rate for ALL the services
2. Find the bounce rate for an individual service

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,10 +36,10 @@ clean:
 
 .PHONY: format
 format:
-	isort .
-	black --config pyproject.toml .
-	flake8 .
-	mypy .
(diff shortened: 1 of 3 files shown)


whabanks (approved):
Tests passing locally, code is looking fresh. The fetch bounce rate for all services sets us up for the monitoring we'll need to do later. LGTM!

andrewleith (approved):
Looks great.  As we discussed yesterday, when we add a new column specifically for bounce type we will need to update this so that it looks at that subset of permanent failures instead of all.

PR #1769, opened 2023-03-03
Update dev container extensions

smcmurtry:
# Summary | Résumé

`Pylance` is now the default python formatter in VS code so we can change to that. And adding the `mypy` extension will hopefully make the type hints in the editor work better.


whabanks:
LTGM!

whabanks (approved):

PR #1770, opened 2023-03-06
Feat/capture all bounce types

andrewleith:
# Summary | Résumé

This PR updates the `app/notifications/notifications_ses_callback.py` module to record all possible bounce types defined by AWS.

It also adds tests to ensure they are identified and captured correctly in the database.


app/notifications/notifications_ses_callback.py
```diff
@@ -44,6 +57,50 @@ def _determine_provider_response(ses_message):
...
+        if bounce_subtype == "General":
+            bounce_response["feedback_subtype"] = NOTIFICATION_SOFT_GENERAL
+        if bounce_subtype == "MailboxFull":
+            bounce_response["feedback_subtype"] = NOTIFICATION_SOFT_MAILBOXFULL
+        if bounce_subtype == "MessageTooLarge":
+            bounce_response["feedback_subtype"] = NOTIFICATION_SOFT_MESSAGETOOLARGE
+        if bounce_subtype == "ContentRejected":
+            bounce_response["feedback_subtype"] = NOTIFICATION_SOFT_CONTENTREJECTED
+        if bounce_subtype == "AttachmentRejected":
+            bounce_response["feedback_subtype"] = NOTIFICATION_SOFT_ATTACHMENTREJECTED
```
jimleroyer, line 99: Should we cover unknown types in case the response is unexpected? or we will consider blank feedback subtype as such?

app/notifications/notifications_ses_callback.py
```diff
@@ -44,6 +57,50 @@ def _determine_provider_response(ses_message):
...
+    elif bounce_type == "Permanent":
+        bounce_response["feedback_type"] = NOTIFICATION_HARD_BOUNCE
+        if bounce_subtype == "General":
+            bounce_response["feedback_subtype"] = NOTIFICATION_HARD_GENERAL
+        if bounce_subtype == "NoEmail":
+            bounce_response["feedback_subtype"] = NOTIFICATION_HARD_NOEMAIL
+        if bounce_subtype == "Suppressed":
+            bounce_response["feedback_subtype"] = NOTIFICATION_HARD_SUPPRESSED
+        if bounce_subtype == "OnAccountSuppressionList":
+            bounce_response["feedback_subtype"] = NOTIFICATION_HARD_ONACCOUNTSUPPRESSIONLIST
```
jimleroyer, line 87: Should we cover unknown types in case the response is unexpected? or we will consider blank feedback subtype as such?

jimleroyer (approved):
Looks good! I left a question for edge cases in the conditions.

app/notifications/notifications_ses_callback.py
andrewleith: Yes you make a good point.  It could work with the blanks now, but I think it would be best to be explicit.  I will re-work. this and add a new migration to update the `feedback_types` and `feedback_subtypes`.

andrewleith: Done, have a look when you have a chance!

app/celery/process_ses_receipts_tasks.py
```diff
@@ -61,6 +61,7 @@ def process_ses_results(self, response):
             notification=notification,
             status=notification_status,
             provider_response=aws_response_dict["provider_response"],
+            bounce_response=aws_response_dict["bounce_response"],
```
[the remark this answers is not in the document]
jimleroyer: Yeah, we should usually prefer None safe way to retrieve dictionary values.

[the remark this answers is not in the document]
andrewleith: Nice catch!  I'll fix up both using `.get()` and default to None if the key isn't found, since the update method handles `None` appropriately.

app/dao/notifications_dao.py
```diff
@@ -124,11 +124,16 @@ def country_records_delivery(phone_prefix):
...
+def _update_notification_status(notification, status, provider_response=None, bounce_response=None):
     status = _decide_permanent_temporary_failure(current_status=notification.status, status=status)
     notification.status = status
     if provider_response:
         notification.provider_response = provider_response
+    if bounce_response:
+        notification.feedback_type = bounce_response["feedback_type"]
+        notification.feedback_subtype = bounce_response["feedback_subtype"]
+        notification.ses_feedback_id = bounce_response["ses_feedback_id"]
+        notification.ses_feedback_date = bounce_response["ses_feedback_date"]
```
jimleroyer, line 136: We could have used the dict.get here as well just in case one of these are missing. I assume it should be all present but just in case.

PR #1780, opened 2023-03-09
Add more safety around json data handling

andrewleith:
# Summary | Résumé

This PR adds more checking around getting values out of JSON to avoid key errors.


app/dao/notifications_dao.py
```diff
@@ -131,10 +131,10 @@ def _update_notification_status(notification, status, provider_response=None, bo
...
         notification.provider_response = provider_response
     if bounce_response:
-        notification.feedback_type = bounce_response["feedback_type"]
-        notification.feedback_subtype = bounce_response["feedback_subtype"]
-        notification.ses_feedback_id = bounce_response["ses_feedback_id"]
-        notification.ses_feedback_date = bounce_response["ses_feedback_date"]
+        notification.feedback_type = bounce_response.get("feedback_type", None)
+        notification.feedback_subtype = bounce_response.get("feedback_subtype", None)
+        notification.ses_feedback_id = bounce_response.get("ses_feedback_id", None)
+        notification.ses_feedback_date = bounce_response.get("ses_feedback_date", None)
```
jimleroyer, line 137: I think it is unnecessary to define `None` as the second parameter, as this is the default.

jimleroyer (commented):
This looks good but left a suggestion to remove `None` as a parameter (as it is the default).

app/dao/notifications_dao.py
andrewleith: 🤦

jzbahrai (approved):

PR #1782, opened 2023-03-09
Refactor salts

sastels:
# Summary | Résumé

Following [the key rotation ADR](https://docs.google.com/document/d/1RXlbgIlYzVY5DMWix81vIZluTp-FBLMz8HCYi-v7I7s/edit#heading=h.6z6awvmxvdpk) we change some of the salts to hard coded values. From the [ItsDangerous documentation](https://itsdangerous.palletsprojects.com/en/2.1.x/concepts/#the-salt):

>The salt is combined with the secret key to derive a unique key for distinguishing different contexts. Unlike the secret key, the salt doesn’t have to be random, and can be saved in code. It only has to be unique between contexts, not private.
>
>For example, you want to email activation links to activate user accounts, and upgrade links to upgrade users to a paid accounts. If all you sign is the user id, and you don’t use different salts, a user could reuse the token from the activation link to upgrade the account. If you use different salts, the signatures will be different and will not be valid in the other context.

Basically we need to make sure that we didn't break any functionality. We should run smoke tests, send and view with the admin, and if possible make a test complaint (we can make a test complaint by emailing complaint@simulator.amazonses.com, see [here](https://docs.aws.amazon.com/ses/latest/dg/send-an-email-from-console.html))

Note that the unit tests are pretty thorough (testing sign / verify for new salts) so we should concentrate on ensuring that old stuff still works:
- in admin look at old notifications (check that their personalisation is there)
- through api send file to yourself using an old key
- look at an old callback and make sure you can see the bearer_token

I ran the smoke tests locally and verified that I got all the emails and texts.


app/encryption.py
```diff
@@ -32,19 +32,70 @@ def init_app(self, app):
...
         "A wrapper around the sign fn to define the argument type and return type"
-        return self.sign(notification)
+        return SignedNotification(self._sign(notification, "notification"))
 
     def verify_notification(self, signed_notification: SignedNotification) -> NotificationDictToSign:
         "A wrapper around the verify fn to define the argument type and return type"
-        return self.verify(signed_notification)
+        return self._verify(signed_notification, "notification")
+
+    def sign_personalisation(self, personalisation: dict) -> str:
```
jimleroyer, line 64: It seems like an opportunity to have a more OOP approach where we do not duplicate the methods, but rather have multiple instances of the signer, for each usage of signing where we'd provide the salts as a parameter upon instantiation via the constructor. Do you know what I mean?

app/models.py
```diff
@@ -821,13 +821,13 @@ class ServiceInboundApi(BaseModel, Versioned):
     @property
     def bearer_token(self):
         if self._bearer_token:
-            return signer.verify(self._bearer_token)
+            return signer.verify_bearer_token(self._bearer_token)
```
jimleroyer, line 824: Regarding to my [previous comment](https://github.com/cds-snc/notification-api/pull/1782/files#r1136994473), this would rather be:
```
return signer_bearer_token.verify(self._bearer_token)
```

app/encryption.py
jimleroyer: Practically too, that would minimize our required testing code to get full coverage if we only have the sign/verify methods instead of duplicating these (and every time we'd add a signer).

sastels: 👍 that's a really good idea! Makes code / tests cleaner and when / if we move towards different `SECRET_KEY`s for each it'll work with minimal code changes.

app/__init__.py
```diff
@@ -99,7 +105,15 @@ def create_app(application, config=None):
...
     notify_celery.init_app(application)
-    signer.init_app(application)
+
+    signer_notification.init_app(application, secret_key=application.config["SECRET_KEY"], salt="notification")
+    signer_personalisation.init_app(application, secret_key=application.config["SECRET_KEY"], salt="personalisation")
+    signer_complaint.init_app(application, secret_key=application.config["SECRET_KEY"], salt="complaint")
+    signer_delivery_status.init_app(application, secret_key=application.config["SECRET_KEY"], salt="delivery_status")
+    signer_bearer_token.init_app(application, secret_key=application.config["SECRET_KEY"], salt="bearer_token")
+    signer_api_key.init_app(application, secret_key=application.config["SECRET_KEY"], salt="api_key")
+    signer_inbound_sms.init_app(application, secret_key=application.config["SECRET_KEY"], salt="inbound_sms")
```
jimleroyer, line 115: Ah I forgot about these init_app calls... I don't like them because they leave an object in an invalid state after the ctor was called, waiting for proper initialization. But anyway, good for our purpose..

app/encryption.py
```diff
@@ -28,23 +28,27 @@ class NotificationDictToSign(TypedDict):
 
 
 class CryptoSigner:
-    def init_app(self, app):
-        self.serializer = URLSafeSerializer(app.config.get("SECRET_KEY"))
-        self.salt = app.config.get("DANGEROUS_SALT")
+    def init_app(self, app: Any, secret_key: str | Iterator[str], salt: str) -> None:
```
jimleroyer, line 31: Oh nice usage of the union operator.

app/encryption.py
```diff
@@ -28,23 +28,27 @@ class NotificationDictToSign(TypedDict):
...
+        self.serializer = URLSafeSerializer(secret_key)
+        self.salt = salt
+        self.dangerous_salt = app.config.get("DANGEROUS_SALT")
```
jimleroyer, line 34: So we have two salts to migrate from old one to new one?

app/encryption.py
```diff
@@ -28,23 +28,27 @@ class NotificationDictToSign(TypedDict):
...
-    def verify_notification(self, signed_notification: SignedNotification) -> NotificationDictToSign:
-        "A wrapper around the verify fn to define the argument type and return type"
-        return self.verify(signed_notification)
+    # TODO: get rid of this after everything is signed with the new salts
+    # This is only needed where we look things up by the signed value:
+    #     - get_api_key_by_secret()
+    def sign_dangerous(self, to_sign: str) -> str | bytes:
+        return self.serializer.dumps(to_sign, salt=self.dangerous_salt)
+
+    # NOTE: currently the verify checks agains the default salt as well as the dangerous salt
```
jimleroyer, line 45:
```suggestion
    # NOTE: currently the verify checks against the default salt as well as the dangerous salt
```

app/encryption.py
```diff
@@ -28,23 +28,27 @@ class NotificationDictToSign(TypedDict):
...
+    # TODO: remove this double check once we've removed DANGEROUS_SALT
```
jimleroyer, line 46: ok got it 👍

tests/app/test_encryption.py
```diff
@@ -1,20 +1,43 @@
+import pytest
+from itsdangerous import BadSignature
+
 from app.encryption import CryptoSigner
 
-signer = CryptoSigner()
 
+def test_sign_and_verify(notify_api):
```
jimleroyer, line 7: We don't want to put these tests in test Class?

jimleroyer (approved):
Looks good to me as-is. 👍

app/__init__.py
sastels: yeah, right now the way we do it we need the application configured to get the `SECRET_KEY`. 🙁

app/encryption.py
sastels: In a later PR we'll get rid of `DANGEROUS_SALT` entirely (well, move it to `PASSWORD_SALT` that is only used to salt the user passwords before hashing). All `DANGEROUS_SALT` usage will be removed from signing related code.

tests/app/test_encryption.py
sastels: 🤷 they have their own file. But, fair enough, it will be more consistent with how we want to organize our tests. 👍

PR #1783, opened 2023-03-09
Task/redis add bounce

jzbahrai:
# Summary | Résumé
Everytime we send a notification, we will keep track of the notification in Redis. This PR adds code in API at the last stage (after a notification has been sent to AWS), and puts a count for that notification in REDIS

Tested manually. I can see the for a given service_id, notifications have been added to Redis

Screenshot 2023-03-14 at 12 12 10 PM


whabanks (approved):
Tests look good, I tested this locally and confirmed that when sending a notification the id can be seen in Redis under `total_notifications:<service_id>`.

PR #1786, opened 2023-03-15
Update utils

jzbahrai:
# Summary | Résumé

Update notification utils to 50.1.8, but had to upgrade other packages that were tied to that upgrade.


whabanks (approved):
LGTM!

PR #1788, opened 2023-03-16
fix: sync CI DB Version with actual version

CalvinRodo:
# Summary | Résumé

In production and staging you are running 11.17 and so I'm just syncing with that.

I went with the bullseye version because there is no `11.17` tag and the version you currently have (13.9) is built off of the bullseye base image.
https://hub.docker.com/layers/library/postgres/13.9/images/sha256-f5911875bf2d18c585a5f245f6596d946d28eabddeffd809ac52ca5720c7c5b5?context=explore

I'm not sure if this would have caught the issue but it's in line with this issue that came from the incident https://github.com/cds-snc/notification-planning-core/issues/17


jzbahrai (approved):

ben851 (approved):
You literally beat me to this by an hour!

ben851:
hmm while the change is good, the tests are failing now. I'll take a look

sastels:
looks like this has since been add by Ben [here](https://github.com/cds-snc/notification-api/pull/1802). We good to close this PR I assume?

PR #1789, opened 2023-03-21
Add code to resign database fields

sastels:
# Summary | Résumé

As discussed in the [ADR](https://docs.google.com/document/d/1RXlbgIlYzVY5DMWix81vIZluTp-FBLMz8HCYi-v7I7s/edit#heading=h.6z6awvmxvdpk), we add code to resign the api key secret, callback bearer tokens, inbound sms, and notifications tables.

We also add a wrapper script to call the code.
```
usage: resign_database.py [-h] [--unsafe]

options:
  -h, --help  show this help message and exit
  --unsafe    resign notifications that have a bad signature
```

This code has been added to this repo rather than the attic since;
- it relies on our database models, and
- it is crucial that we have tests for this code, and these tests need the test framework set up in this repo.

The "unsafe" option ignores the signature when resigning. It is only intended to be used if the database gets into a state where some of the signatures cannot be verified. This should only happen if we rotate keys 
```
[k1, k2] -> [k2, k3] -> [k3, k4]
```
**without** resigning after each step.

Note: use a template with a variable

- Assume that your current SECRET_KEY is K1 or K0,K1.
- test that you can POST an email and view sent emails in admin
- make a new key K2 (locally can just use the string "K2")
- set SECRET_KEY to K2,K1 and restart admin / api / celery
- make sure you can view the email you just sent
- test that you can POST an email and view sent emails in admin
- set SECRET_KEY to K1,K2 and restart admin / api / celery
- make sure you can view the emails you already sent
- test that you can POST an email and view sent emails in admin

At this point some records have been signed with K1 and some with K2. You can read all of them since you're verifying against both K1 and K2. But we'd like to stop using K1 in the future, so we need to have everything signed with K2.

- run `scripts/resign_database.py`

Now everything is signed with K2! To verify this:

- set SECRET_KEY to K2 and restart admin / api / celery
- make sure you can view the emails you already sent
- test that you can POST an email and view sent emails in admin

The `resign_database.py` script has an "unsafe" option. This ignores the signature and just resigns the data with the new key. To test this:
- change your SECRET_KEY to K3 and restart admin / api / celery
- verify that everything is broken (cannot POST an email, cannot view emails in admin)
- run `resign_database.py` - crashes complaining about signatures not matching
- run `resign_database.py --unsafe`. The script will report bad signatures but will still resign the values so that the system works again.
- verify that you can once again POST emails and view in admin


sastels:
Moved to draft until after the Bug Bash

app/config.py
```diff
@@ -408,6 +408,28 @@ class Config(object):
...
+            else crontab(day_of_month=3, hour=5, minute=15),
+            "options": {"queue": QueueNames.PERIODIC},
+        },
+        "resign-inbound-sms": {
+            "task": "resign-inbound-sms",
+            "schedule": crontab(minute=10)
+            if os.getenv("NOTIFY_ENVIRONMENT", "development") == "staging"
+            else crontab(day_of_month=3, hour=5, minute=30),
+            "options": {"queue": QueueNames.PERIODIC},
+        },
```
jimleroyer, line 432: Following Ben's recommendation: we could avoid configuration based on environment name and rather use feature flags. In this case though, we could have this tied to a configuration value (via an abstract method existing in children Config classes for example) that store either the cron value or an enum such as hourly or montly (to restrict values and mistakes).

app/dao/api_key_dao.py
```diff
@@ -10,6 +10,14 @@
 from app.models import ApiKey
 
 
+@transactional
+def resign_api_keys():
```
jimleroyer, line 14: Let's add some Python documentation to say what's the intent of this method, we have the what with the code but we can add the why.

tests/app/celery/test_monthly_tasks.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class TestResigning:
+    def test_resign_callbacks(self, sample_service):
+        from app import signer_bearer_token
+
+        signer_bearer_token.serializer = URLSafeSerializer(["k1", "k2"])
+        initial_callback = create_service_callback_api(service=sample_service)
+        bearer_token = initial_callback.bearer_token
+        _bearer_token = initial_callback._bearer_token
```
jimleroyer, line 19: Hmm I don't know how the setter and getter works, but seems like invoking the resign will change the internal value, right?

sastels: `bearer_token = initial_callback.bearer_token` will call the getter, which will verify the `_bearer_token` and return the unsigned (ie verified) value. 
`_bearer_token = initial_callback._bearer_token` will get the value of `_bearer_token` directly, ie no getter called.

Whe we do something like `callback.bearer_token = new_token` then the setter is called, `new_token` is signed and the signed value is stuck in  `_bearer_token`

sastels:
Made a few changes:
- clarify what the resigning is doing
- use feature flag `FF_RESIGN_TASKS_ENABLED` and variable `RESIGN_TASK_FREQUENCY` to set if the resigning tasks run, and if so what's their frequency. Defaults are `False` and `monthly`.

sastels:
Have refactored the code to remove the tasks and run the resigning code manually via a script. This makes the code simpler as well as easier to use. In particular, if we need to quickly rotate the keys we won't have to hack things to run the tasks immediately.

app/config.py
```diff
@@ -356,7 +356,6 @@ class Config(object):
             "schedule": 10,
             "options": {"queue": QueueNames.PERIODIC},
         },
-        ""
```
jimleroyer, line 359: 🤨 hmm

app/dao/api_key_dao.py
```diff
@@ -10,6 +12,26 @@
...
+@transactional
+def resign_api_keys(unsafe: bool = False):
+    # Resign the secret column of the api_keys table
+    # This allows us to rotate the secret key used to sign the api key secret
+    rows = ApiKey.query.all()  # noqa
+    current_app.logger.info(f"Resigning {len(rows)} api keys")
+
+    for row in rows:
+        try:
+            unsigned_secret = getattr(row, "secret")  # unsign the secret
```
jimleroyer, line 24: good comment!

app/dao/api_key_dao.py
```diff
@@ -10,6 +12,26 @@
 from app.models import ApiKey
 
 
+@transactional
+def resign_api_keys(unsafe: bool = False):
+    # Resign the secret column of the api_keys table
+    # This allows us to rotate the secret key used to sign the api key secret
```
jimleroyer, line 18: Let's write a proper python doc string for the function and we can document the parameter as well, i.e. what it does.

app/dao/api_key_dao.py
```diff
@@ -27,20 +49,22 @@ def expire_api_key(service_id, api_key_id):
...
-            .one()
-        )
+    signed_with_all_keys = signer_api_key.sign_with_all_keys(str(secret))
+    for signed_secret in signed_with_all_keys:
+        try:
+            return db.on_reader().query(ApiKey).filter_by(_secret=signed_secret).options(joinedload("service")).one()
+        except NoResultFound:
+            pass
+
+    signed_dangerous_with_all_keys = signer_api_key.sign_dangerous_with_all_keys(str(secret))
```
jimleroyer, line 61: The difference is that we will tie the salt with this signage, right? can we rename `sign_dangerous_with_all_keys` to something like `sign_with_all_salted_keys`?

app/dao/notifications_dao.py
```diff
@@ -57,6 +58,29 @@
 )
 
 
+@transactional
+def resign_notifications(unsafe: bool = False):
```
jimleroyer, line 62: Same applies there with [docstring format](https://peps.python.org/pep-0257/).

app/encryption.py
```diff
@@ -28,20 +28,34 @@ class NotificationDictToSign(TypedDict):
...
+            signed.append(URLSafeSerializer(k).dumps(to_sign, salt=self.salt))
+        return signed
+
     # TODO: get rid of this after everything is signed with the new salts
     # This is only needed where we look things up by the signed value:
     #     - get_api_key_by_secret()
     def sign_dangerous(self, to_sign: str) -> str | bytes:
         return self.serializer.dumps(to_sign, salt=self.dangerous_salt)
 
+    def sign_dangerous_with_all_keys(self, to_sign: str | NotificationDictToSign) -> List[str | bytes]:
```
jimleroyer, line 53: I'd rename that one to mean that we are signing with the old salt, as saying sign dangerous is not obvious on what is happening.

app/encryption.py
```diff
@@ -50,6 +64,10 @@ def verify(self, to_verify: str | bytes) -> Any:
         except BadSignature:
             return self.serializer.loads(to_verify, salt=self.dangerous_salt)
 
+    # This ignores the signature so should be used with caution
```
jimleroyer, line 67: Let's put this comment as a docstring too please. 🙏 Along with parameters documentation.

scripts/resign_database.py
```diff
@@ -0,0 +1,46 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--unsafe", default=False, action='store_true', help="resign notifications that have a bad signature")
+    args = parser.parse_args()
+
+    load_dotenv()
+    application = Flask("resign_database")
+    create_app(application)
+    application.app_context().push()
+
+    resign_all(args.unsafe)
```
jimleroyer, line 46: Well done, all the signing functions are in their appropriate modules and that command line is simply calling these. The scope and isolation is good.

tests/app/dao/notification_dao/test_notification_dao.py
```diff
@@ -1797,3 +1800,46 @@ def test_bulk_insert_notification_duplicate_ids(self, sample_template):
...
+            initial_notification = create_notification(sample_template_with_placeholders, personalisation={"Name": "test"})
+            save_notification(initial_notification)
+            personalisation = initial_notification.personalisation
+            _personalisation = initial_notification._personalisation
+
+        with set_signer_secret_key(signer_personalisation, ["k2", "k3"]):
+            resign_notifications()
+            notification = Notification.query.get(initial_notification.id)
+            assert notification.personalisation == personalisation
+            assert notification._personalisation != _personalisation
```
jimleroyer, line 1819: This does not make the test obvious in what's happening and expected.

tests/app/test_encryption.py
```diff
@@ -37,3 +37,35 @@ def test_should_sign_dicts(self, notify_api):
...
+        assert signer12.sign_with_all_keys("this") == [signer2.sign("this"), signer1.sign("this")]
+
+    def test_sign_dangerous_with_all_keys(self, notify_api):
+        signer1 = CryptoSigner()
+        signer1.init_app(notify_api, "s1", "salt1")
+        signer2 = CryptoSigner()
+        signer2.init_app(notify_api, "s2", "salt2")
+        signer12 = CryptoSigner()
+        signer12.init_app(notify_api, ["s1", "s2"], "salt3")
+        assert signer12.sign_dangerous_with_all_keys("this") == [signer2.sign_dangerous("this"), signer1.sign_dangerous("this")]
```
jimleroyer, line 71: Very nice set of tests. 👏

ben851 (approved):
I ran through the test steps locally and validated that all works as expected.

app/dao/api_key_dao.py
sastels: yeah, that's a bit clearer 👍

jimleroyer (approved):
Great changes! :shipit:

PR #1795, opened 2023-03-23
treat SECRET_KEY as a list

sastels:
# Summary | Résumé

In preparation for our ability to rotate these signing keys without downtime, here we just read `SECRET_KEY` into a list of strings rather than just a string. This will have no affect the app.

Note that `env.list()` takes a comma separated list and splits into a list of strings. Currently the key is one string without commas so this will become a list of one key.

`SECRET_KEY` is used in admin and api to set the key for the serializer we import from the itsDangerous library (and this is the only way we use it).

As discussed in [our signing library's documentation](https://itsdangerous.palletsprojects.com/en/2.1.x/concepts/#key-rotation):
>Instead of passing a single key, you can pass a list of keys, oldest to newest. When signing the last (newest) key will be used, and when validating each key will be tried from newest to oldest before raising a validation error.

Everything should behave exactly as before.

note: smoke tested locally.


ben851 (approved):
LGTM!

PR #1799, opened 2023-03-24
sign with DANGEROUS_SALT for now

sastels:
# Summary | Résumé

We had an issue when we merged in #1782 where during the lambda redeploy, old lambdas got bad signature errors trying to read notifications signed with new lambdas 😞 

In this PR we switch signing back to the old DANGEROUS_SALT (and keep the same "verify with either salt" code)

So there shouldn't be this old lambda / new lambda issue when this is released to production.

We will do a subsequent PR / release to prod where we switch back to signing with the new salts, but since the "verify with either" code will already be in place there won't be an "old lambda / new lambda" issue then either.

See [this Slack discussion](https://gcdigital.slack.com/archives/CV38DBNVA/p1679681730197659?thread_ts=1679666261.261539&cid=CV38DBNVA) for more info.

Everything should still work. (even stuff signed with the new salts from #1782)


ben851 (approved):
LGTM

PR #1800, opened 2023-03-27
feat: Salesforce Engagement create

patheard:
# Summary
Create a Salesforce Engagement when a new Notify service is created.  

The Notify user that creates the Engagement will have their associated Salesforce Contact updated.  If there is not already a Salesforce Contact for the Notify user, one will be created.

# Related
- cds-snc/platform-core-services#282

### DIFF
--- a/app/clients/salesforce/salesforce_engagement.py
+++ b/app/clients/salesforce/salesforce_engagement.py
@@ -1,0 +1,77 @@
+from __future__ import annotations
+
+from datetime import datetime
+from typing import TYPE_CHECKING, Optional
+
+from flask import current_app
+from simple_salesforce import Salesforce
+
+from .salesforce_utils import parse_result
+
+if TYPE_CHECKING:
+    from app.models import Service
+
+ENGAGEMENT_PRODUCT = "GC Notify"
+ENGAGEMENT_TEAM = "Platform"
+ENGAGEMENT_TYPE = "New Business"
+ENGAGEMENT_STAGE_ACTIVATION = "Activation"
+ENGAGEMENT_STAGE_LIVE = "Live"
+ENGAGEMENT_STAGE_TRIAL = "Trial Account"
+
+
+def create(
+    session: Salesforce, service: Service, stage_name: str, account_id: Optional[str], contact_id: Optional[str]
+) -> Optional[str]:
+    """Create a Salesforce Engagement for the given Notify service
+
+    Args:
+        session (Salesforce): Salesforce session to perform the operation.
+        service (Service): The service's details for the engagement.
+        stage_name (str): The service's stage name.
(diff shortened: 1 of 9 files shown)


tests/app/clients/test_salesforce_account.py
smcmurtry, line 1: Nice tests!

tests/app/clients/test_salesforce_engagement.py
```diff
@@ -0,0 +1,72 @@
...
+                "PricebookEntryId": "the ring",
+                "Product2Id": "my precious",
+                "Quantity": 1,
+                "UnitPrice": 0,
+            },
+            headers={"Sforce-Duplicate-Rule-Header": "allowSave=true"},
+        )
+
+
+def test_create_no_engagement_id(mocker, notify_api, service):
```
smcmurtry, line 58: Does `OpportunityLineItem` not get called because the `notify_api.config` variables aren't set?

app/clients/salesforce/salesforce_engagement.py
```diff
@@ -0,0 +1,77 @@
...
+    Args:
+        session (Salesforce): Salesforce session to perform the operation.
+        service (Service): The service's details for the engagement.
+        stage_name (str): The service's stage name.
+        account_id (Optional[str]): Salesforce Account ID to associate with the Engagement.
+        contact_id (Optional[str]): Salesforce Contact ID to associate with the Engagement.
+
+    Returns:
+       Optional[str]: Newly created Engagement ID or None if the operation failed.
+    """
```
smcmurtry, line 36: I love the detailed doc strings!

app/service/rest.py
```diff
@@ -260,6 +260,8 @@ def create_service():
     try:
         # try-catch; just in case, we don't want to error here
         ZenDeskSell().send_create_service(valid_service, user)
+        if current_app.config["FF_SALESFORCE_CONTACT"]:
+            salesforce_client.engagement_create(valid_service, user)
```
smcmurtry, line 264: It's great that you've been able to keep all the logic inside the salesforce client, the changes to the existing api code are so minimal 😍

smcmurtry (approved):
Looks great Pat!

app/service/rest.py
patheard: Thanks!  Having the `salesforce_client` also made it easier to group stuff compared to our initial work in the admin PR review env.

tests/app/clients/test_salesforce_engagement.py
patheard: No, in this case it's not being created because the `Engagement` create is failing in this test - it's based on the newly created `engagement_id` being returned from the Salesforce API.

PR #1801, opened 2023-03-27
Revert "sign with DANGEROUS_SALT for now (#1799)"

sastels:
This reverts commit 7d91af1e075930904a2bca202ae6b7f1e33d82ef.

# Summary | Résumé

Now that prod has the "verify with old or new salt" code, we can change the signing code back to using the new salt.

Should be no differences.

In particular, should be able to view both old notifications and new ones and send using new or old api keys.


ben851 (approved):
Looks good to go!

PR #1802, opened 2023-03-27
Modifying pgsql docker image to reflect what is in production

ben851:
# Summary | Résumé

Locally, and in Github Actions, we were running Postgres Version 13.9, but in production we are running 11.17. I've changed the docker-compose and github actions to the correct postgres version. This broke one of the tests because the wording of the error was different between versions. I've corrected the assertion to have the proper wording of 11.17.

---

poetry run make test

---

None, this impacts development and github actions only.

# Necessary steps to perform before and after the deployment of these changes.

**Note: I had to clear my dev container volume cache to get this to work locally. This is because postgres will complain that the database was initialized with the old (13.9) version. Simply run docker-compose down --volumes to remove the cached volumes**


jimleroyer (approved):
LGTM!

jzbahrai (approved):

PR #1803, opened 2023-03-27
feat: Salesforce Engagement go live flow

patheard:
## Summary
Update the Salesforce Engagement stage as a Notify service goes through the `go live` flow.

This involves setting the Engagement stage to `Activiation` when the request to go live is made.  Once the service is set as live, the Engagement stage is then set to `Live`.

### With env var `FF_SALESFORCE_CONTACT = True`
1. Login and create a new service.
1. Check Salesforce and expect to see a new Engagement with `stage = Trial`
1. Request for the service to go live.
1. Check Salesforce and expect the service Engagement to have `stage = Activation`
1. Set the service as live.
1. Check Salesforce and expect the service Engagement to have `stage = Live`

### With env var `FF_SALESFORCE_CONTACT = False`
1. Repeat the above Notify steps and expect no Salesforce Engagement creation or stage updates.

## Related
- cds-snc/platform-core-services#285

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -16,6 +16,10 @@ from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm.exc import NoResultFound
 
 from app import redis_store, salesforce_client
+from app.clients.salesforce.salesforce_engagement import (
+    ENGAGEMENT_STAGE_ACTIVATION,
+    ENGAGEMENT_STAGE_LIVE,
+)
 from app.config import QueueNames
 from app.dao import fact_notification_status_dao, notifications_dao
 from app.dao.api_key_dao import (
@@ -78,6 +82,7 @@ from app.dao.services_dao import (
     dao_fetch_all_services_by_user,
     dao_fetch_live_services_data,
     dao_fetch_service_by_id,
+    dao_fetch_service_creator,
     dao_fetch_todays_stats_for_all_services,
     dao_fetch_todays_stats_for_service,
     dao_remove_user_from_service,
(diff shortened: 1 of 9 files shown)


app/service/rest.py
```diff
@@ -303,6 +308,22 @@ def update_service(service_id):
...
+            # Two scenarios, if there is a user that has requested to go live, we will use that user
+            # to create a Contact/Engagment pair between Notify and Salesforce.
+            # If by any chance there is no tracked request to a user, Notify will try to identify the user
+            # that created the service and then create a Contact/Engagment relationship.
+            if service.go_live_user_id:
+                user = get_user_by_id(service.go_live_user_id)
+            else:
+                user = dao_fetch_service_creator(service.id)
+
+            stage_name = ENGAGEMENT_STAGE_LIVE if service_going_live else ENGAGEMENT_STAGE_ACTIVATION
```
smcmurtry, line 322: I may be missing something, but I don't think this code will run when a "go live" request is made. "Go live" requests hit a different endpoint: https://github.com/cds-snc/notification-api/blob/main/app/user/rest.py#L438

smcmurtry: However this code will get executed when a platform admin user makes a service live.

tests/app/clients/test_salesforce_engagement.py
smcmurtry, line 1: Tests look good!

app/service/rest.py
patheard: In my local testing, the request to "go live" was also hitting that endpoint because of this service update call:
https://github.com/cds-snc/notification-admin/blob/d587ea1169f6813c17a3c6a86c76720fd620a357/app/main/views/service_settings.py#L356

smcmurtry (approved):
Code looks great, love the doc strings and modular approach!

PR #1805, opened 2023-03-28
Freshdesk failover

sastels:
# Summary | Résumé

If the Freshdesk integration fails (say, during a key rotation 😅 ) send the "Contact Us" form via email.

1. Run `flask db upgrade` to add the new template
2. Set the env var `CONTACT_FORM_EMAIL_ADDRESS`
3. Make sure you env doesn't have a valid `FRESH_DESK_API_KEY`
4. Submit a Contact Us form.
5. Get an email that looks like
```
Freshdesk integration has failed. Sending the following contact us form:
{
“product_id”: 61111111,
“subject”: “Get technical support”,
“description”: “The internet is slow”,
“email”: “[asdfa@google.com](mailto:asdfa@google.com)”,
“priority”: 1,
“status”: 2,
“tags”: []
}

L’intégration de Freshdesk a échoué. Envoi du formulaire de contact suivant :

{
“product_id”: 61111111,
“subject”: “Get technical support”,
“description”: “The internet is slow”,
“email”: “[asdfa@google.com](mailto:asdfa@google.com)”,
“priority”: 1,
“status”: 2,
“tags”: []
}
```


app/clients/freshdesk.py
```diff
@@ -96,8 +104,34 @@ def send_ticket(self) -> int:
...
-        except requests.RequestException as e:
+        except requests.RequestException:
             content = json.loads(response.content)
-            current_app.logger.error(f"Failed to create Freshdesk ticket: {content['errors']}")
-            raise e
+            current_app.logger.error(f"Failed to create Freshdesk ticket: {content}")
+            self.email_freshdesk_ticket(self._generate_ticket())
+            return 201
+
+    def email_freshdesk_ticket(self, content: dict) -> None:
```
jimleroyer, line 115: Do you think we should catch further errors that might trigger from sending the notification and log?

app/clients/freshdesk.py
```diff
@@ -96,8 +104,34 @@ def send_ticket(self) -> int:
...
+        template = dao_get_template_by_id(current_app.config["CONTACT_FORM_DIRECT_EMAIL_TEMPLATE_ID"])
+        notify_service = dao_fetch_service_by_id(current_app.config["NOTIFY_SERVICE_ID"])
+
+        if current_app.config["CONTACT_FORM_EMAIL_ADDRESS"] is None:
+            current_app.logger.info("Cannot email contact us form, CONTACT_FORM_EMAIL_ADDRESS is empty")
+        else:
+            current_app.logger.info("Emailing contact us form to {}".format(current_app.config["CONTACT_FORM_EMAIL_ADDRESS"]))
```
jimleroyer, line 122: Good to log in here on what's happening. 👍

app/clients/freshdesk.py
sastels: If there are further errors saving or sending the ticket we want that to surface as an error.

Hrm. we could log the contact info in this case. Might have PII in it, but we could delete the log stream later. Likely better than throwing away the contact data. Probably we'd want to raise a critical alarm as well, some sort of "Contact data could not be sent" so we'd know to check the logs.

Thoughts?

jimleroyer: Sounds good to me. I think we can drop the re-raise though, else we'll get this error twice in the logs (and we are not sure how/what the parent exception handler might show). The exception handled, so we can consider it properly managed and move on. We could then add alarms on these logs to surface it up to #notification-ops . Does that make sense?

tests/app/clients/test_freshdesk.py
```diff
@@ -1,233 +1,262 @@
...
-
-def test_send_ticket_freshdesk_integration_disabled(mocker, notify_api: Flask):
-    with set_config_values(notify_api, {"FRESH_DESK_ENABLED": "False"}):
-        mocked = mocker.patch("requests.post")
-
-        with notify_api.app_context():
-            response = Freshdesk(ContactRequest(email_address="test@email.com")).send_ticket()
-            mocked.assert_not_called()
-            assert response == 201
+class TestSendTicket:
```
jimleroyer, line 13: That is great -- leaving a cleaner state. 👍

app/clients/freshdesk.py
sastels: 👍  yeah. And the main thing is that we're logging that the contact email failed.

jimleroyer (approved):
:shipit:

PR #1806, opened 2023-03-28
feat: add timeout to Salesforce API requests

patheard:
# Summary
Update the Salesforce API requests to timeout after 10 seconds. This will prevent excessive waiting for Notify users if the Salesforce API is unresponsive.

### DIFF
--- a/app/clients/salesforce/salesforce_auth.py
+++ b/app/clients/salesforce/salesforce_auth.py
@@ -2,12 +2,14 @@ import requests
 from flask import current_app
 from simple_salesforce import Salesforce
 
+SALESFORCE_TIMEOUT_SECONDS = 10
+
 
 class TimeoutAdapter(requests.adapters.HTTPAdapter):
     """Custom adapter to add a timeout to Salesforce API requests"""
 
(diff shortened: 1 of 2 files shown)


app/clients/salesforce/salesforce_auth.py
```diff
@@ -1,7 +1,16 @@
+import requests
 from flask import current_app
 from simple_salesforce import Salesforce
 
 
+class TimeoutAdapter(requests.adapters.HTTPAdapter):
+    """Custom adapter to add a timeout to Salesforce API requests"""
+
+    def send(self, *args, **kwargs):
+        kwargs["timeout"] = 10
```
smcmurtry, line 10: Let's make this an all caps constant defined at the top, something like `SALESFORCE_TIMEOUT_SECONDS`

smcmurtry (approved):
Looks good, one comment to address before merging

app/clients/salesforce/salesforce_auth.py
patheard: Good idea - just added

PR #1807, opened 2023-03-29
feat: close engagement when service deleted

patheard:
# Summary 
Add Salesforce integration that closes an Engagement if a linked Notify Service is deleted.

Update the Engagement create operation to include the other organisation name if it is provided.

### DIFF
--- a/app/clients/salesforce/salesforce_client.py
+++ b/app/clients/salesforce/salesforce_client.py
@@ -66,7 +66,7 @@ class SalesforceClient:
             service (Service): The Notify service to retrieve the account from.
             user (User): The Notify user to update the Salesforce Contact for.  If a contact does not exist, one will be created.
         """
-        account_name = salesforce_account.get_account_name_from_org(service.organisation_notes)
+        account_name = salesforce_account.get_org_name_from_notes(service.organisation_notes)
         account_id = salesforce_account.get_account_id_from_name(session, account_name, self.generic_account_id)
         contact_id = salesforce_contact.update_account_id(session, user, account_id)
         return account_id, contact_id
@@ -84,7 +84,7 @@ class SalesforceClient:
         """
         session = self.get_session()
         account_id, contact_id = self.contact_update_account_id(session, service, user)
-        salesforce_engagement.create(session, service, salesforce_engagement.ENGAGEMENT_STAGE_TRIAL, account_id, contact_id)
+        salesforce_engagement.create(session, service, {}, account_id, contact_id)
         self.end_session(session)
 
     def engagement_update_stage(self, service: Service, user: User, stage_name: str):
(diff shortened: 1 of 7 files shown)


app/clients/salesforce/salesforce_client.py
```diff
@@ -98,5 +98,22 @@ def engagement_update_stage(self, service: Service, user: User, stage_name: str)
         """
         session = self.get_session()
         account_id, contact_id = self.contact_update_account_id(session, service, user)
-        salesforce_engagement.update_stage(session, service, stage_name, account_id, contact_id)
+        stage_update = {"StageName": stage_name}
+        salesforce_engagement.update(session, service, stage_update, account_id, contact_id)
+        self.end_session(session)
+
+    def engagement_close(self, service: Service):
```
smcmurtry, line 105: Since we're annotating these we could specify what the function returns as well
```suggestion
    def engagement_close(self, service: Service) -> None:
```

app/clients/salesforce/salesforce_client.py
```diff
@@ -98,5 +98,22 @@ def engagement_update_stage(self, service: Service, user: User, stage_name: str)
...
+        """Closes a Salesforce Engagement for the given Notify service.
+
+        Args:
+            service (Service): Notify Service to close an Engagement for.
+        """
+        session = self.get_session()
+        engagement = salesforce_engagement.get_engagement_by_service_id(session, str(service.id))
+        if engagement:
+            close_update = {
+                "CDS_Close_Reason__c": "Trial deleted by user",
```
smcmurtry, line 115: This could also happen to a live account, correct?
```suggestion
                "CDS_Close_Reason__c": "Service deleted by user",
```

app/clients/salesforce/salesforce_engagement.py
```diff
@@ -37,19 +39,23 @@ def create(
...
+                "CDS_Opportunity_Number__c": str(service.id),
+                "Notify_Organization_Other__c": get_org_name_from_notes(service.organisation_notes, ORG_NOTES_OTHER_NAME_INDEX),
+                "CloseDate": datetime.today().strftime("%Y-%m-%d"),
+                "RecordTypeId": current_app.config["SALESFORCE_ENGAGEMENT_RECORD_TYPE"],
+                "StageName": ENGAGEMENT_STAGE_TRIAL,
+                "Type": ENGAGEMENT_TYPE,
+                "CDS_Lead_Team__c": ENGAGEMENT_TEAM,
+                "Product_to_Add__c": ENGAGEMENT_PRODUCT,
+            }
+            field_values = field_default_values | field_updates
```
smcmurtry, line 56: I just tried running this code and it works, very cool! I'm more used to `{**field_default_values, **field_updates}`  but this is good too.

smcmurtry (approved):
Looks good! A couple minor comments to address.

app/clients/salesforce/salesforce_engagement.py
patheard: heh same, I'd always used `{**field_default_values, **field_updates}` this in the past as well

PR #1810, opened 2023-03-30
feat: update Salesforce contact if profile changes

patheard:
# Summary
Update a Notify user's linked Salesforce Contact if the user profile is updated.

This change also applies the same Salesforce Contact refactor that was made to the Engagements, which will make it easier to support future update requirements.

# Related
- Closes cds-snc/platform-core-services#304

### DIFF
--- a/app/clients/salesforce/salesforce_contact.py
+++ b/app/clients/salesforce/salesforce_contact.py
@@ -56,10 +56,10 @@ def update(session: Salesforce, user: User, field_updates: dict[str, Optional[st
     Args:
         session (Salesforce): Salesforce session used to perform the operation.
         user (User): Notify User object for the linked Contact to update
-        field_updates (dict[str, str]): The contact fields to update.
+        field_updates (dict[str, Optional[str]]): The contact fields to update.
 
     Returns:
-         contact_id (str): ID of the updated Contact or None if the operation failed
+        contact_id (Optional[str]): ID of the updated Contact or None if the operation failed
     """
     contact_id = None
     try:
--- a/app/clients/salesforce/salesforce_contact.py
+++ b/app/clients/salesforce/salesforce_contact.py
@@ -16,13 +16,13 @@ if TYPE_CHECKING:
     from app.models import User
 
 
-def create(session: Salesforce, user: User, account_id: Optional[str]) -> Optional[str]:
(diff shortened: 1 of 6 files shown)


app/user/rest.py
```diff
@@ -178,6 +178,13 @@ def update_user_attribute(user_id):
 
         send_notification_to_queue(saved_notification, False, queue=QueueNames.NOTIFY)
 
+    if current_app.config["FF_SALESFORCE_CONTACT"]:
+        try:
+            updated_user = get_user_by_id(user_id=user_id)
```
smcmurtry, line 183: We do have the updated user data from the request without having to make another db call. But making the db query is less complicated and I don't think this will create any performance issues since we wouldn't have a high volume of requests to this endpoint. 👍

app/clients/salesforce/salesforce_contact.py
```diff
@@ -49,14 +50,13 @@ def create(session: Salesforce, user: User, account_id: Optional[str]) -> Option
...
-    """Update the Account ID of a Contact.  If the Contact does not
-    exist, it is created.
+def update(session: Salesforce, user: User, field_updates: dict[str, Optional[str]]) -> Optional[str]:
+    """Update a Contact's details.  If the Contact does not  exist, it is created.
 
     Args:
         session (Salesforce): Salesforce session used to perform the operation.
         user (User): Notify User object for the linked Contact to update
-        account_id (str): ID of the Salesforce Account
+        field_updates (dict[str, str]): The contact fields to update.
```
smcmurtry, line 59:
```suggestion
        field_updates (dict[str, Optional[str]]): The contact fields to update.
```

app/clients/salesforce/salesforce_contact.py
```diff
@@ -49,14 +50,13 @@ def create(session: Salesforce, user: User, account_id: Optional[str]) -> Option
...
 
     Returns:
          contact_id (str): ID of the updated Contact or None if the operation failed
```
smcmurtry, line 62:
```suggestion
         contact_id (Optional[str]): ID of the updated Contact or None if the operation failed
```

smcmurtry (approved):
Looks good! A couple minor type annotation things to fix

app/user/rest.py
patheard: Yeah, I waffled on this a bit as well since it's a bit wasteful to re-query the database when we have the data in a dict.

PR #1813, opened 2023-03-31
fix: Salesforce Contact one word names

patheard:
# Summary
Fix the Salesforce Contact create if the Notify user's name is only one word.  Salesforce requires the `LastName` field to be set, but the `FirstName` field is optional.

# Related
- Closes cds-snc/platform-core-services#308


smcmurtry (approved):
Looks good!

PR #1816, opened 2023-04-03
fix/reduce no row found server errors

sastels:
# Summary | Résumé

Don't raise so many "No row found for one" errors.

run admin / api and request info on on-existant notifications or jobs. Should not see the server error.

Just adding this:
```
method [sqlalchemy.orm.Query.](https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query)first()[¶](https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.first)
Return the first result of this Query or None if the result doesn’t contain any row.

first() applies a limit of one within the generated SQL, so that only one primary entity row is generated on the server side (note this may consist of multiple result rows if join-loaded collections are present).
```


jzbahrai (approved):
I love this PR, thank you for the changes!

PR #1818, opened 2023-04-03
Add script to run celery purge

sastels:
# Summary | Résumé

Add a script to purge the celery queues. Useful when testing locally and adding new periodic tasks.

- Run the celery beat script on its own for a bit.
- Run the purge script and you should see how many jobs were removed from the queues.


scripts/run_celery_beat.sh
```diff
@@ -2,4 +2,5 @@
 
 set -e
 
+celery -A run_celery.notify_celery purge -f || true
```
sastels, line 5: The `|| true` ensures that if the purge fails the beat on the next line will still start.

jimleroyer: I think we should be careful with adding purge to a production script. It seems this command will delete all pending tasks. My question would be, would that delete only tasks that we consider invalid, or valid ones too? How do we differentiate these?

jimleroyer (dismissed):

scripts/run_celery_beat.sh
sastels: oh very good catch, we definitely don't want to purge in prod.

sastels:
Now need to add the keyword `purge` to trigger the purge. staging / prod will be unaffected.

sastels:
Moving to draft until we see exactly what will be purged (do devs share queues?)

sastels:
Moved to a new script to make it clearer what the scripts are doing.

scripts/run_celery_purge.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+set -e
+
```
jimleroyer, line 4: Could we add a warning to not run this against production, with confirmation prompt? 😅

sastels: Added confirmation:
```
$ scripts/run_celery_purge.sh

--------------------------------------------------

                   WARNING!!!!

    This script is for local development only!

  It will delete everything in the celery queues.

--------------------------------------------------

Are you sure you want to continue?
If so, type 'purge'> 
```

andrewleith (approved):
LGTM, purge works as expected and prompt works like a charm too ✅

jimleroyer (approved):

PR #1822, opened 2023-04-05
Fix the trial service report query

smcmurtry:
# Summary | Résumé

draft for now


smcmurtry:
Will re-open if we end up needing this - it actually changes the query to return data for services that have not sent notifications, which may or may not be desired.

PR #1823, opened 2023-04-06
fix: contact is `go live` request if condition

patheard:
# Summary
Update the check for contact `go live` request to properly call the model's method.  Also updates the unit tests to assert that the Salesforce client method is not invoked for other contact types.

# Related
- cds-snc/platform-core-services#313

### DIFF
--- a/app/user/rest.py
+++ b/app/user/rest.py
@@ -464,15 +464,12 @@ def send_contact_request(user_id):
         pass
 
     # Update the engagement stage in Salesforce for go live requests
-    if contact and contact.is_go_live_request and current_app.config["FF_SALESFORCE_CONTACT"]:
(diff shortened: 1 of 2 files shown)


app/user/rest.py
```diff
@@ -464,15 +464,12 @@ def send_contact_request(user_id):
         pass
 
     # Update the engagement stage in Salesforce for go live requests
-    if contact and contact.is_go_live_request and current_app.config["FF_SALESFORCE_CONTACT"]:
+    if contact and contact.is_go_live_request() and current_app.config["FF_SALESFORCE_CONTACT"]:
```
smcmurtry, line 467: Wow, that was the cause?! yikes

smcmurtry (approved):
Nice find Pat!

app/user/rest.py
patheard: haha, I know - @andrewleith came in clutch here

PR #1817, opened 2023-04-03
Populate Freshdesk organisation in "go live" requests using `service.organisation_notes`

smcmurtry:
# Summary | Résumé

Populate Freshdesk organisation in "go live" requests using `service.organisation_notes`

PR #1826, opened 2023-04-12
fix: handle cases where no organisation notes

patheard:
# Summary
Update the `go live` contact request to handle the case where a service does not have `organisation_notes` specified.  This will be the case for all services created before the Salesforce integration was turned on.

Currently this missing `organisation_notes` column is breaking the `go live` request flow and preventing the Freshdesk ticket from being created.

# Related
- #1817


jimleroyer (approved):
LGTM!

PR #1827, opened 2023-04-13
Bump utils version to 50.2.0

whabanks:
# Summary | Résumé
Bumping utils version to 50.2.0


jzbahrai (approved):
Need to fix this as well: .github/workflows/test.yaml

    - name: Checks for new endpoints against AWS WAF rules
      uses: cds-snc/notification-utils/.github/actions/waffles@50.1.8

PR #1834, opened 2023-04-17
Task/add notifications seeded

jzbahrai:
# Summary | Résumé

Add code to seed notifications in Redis.

- Added DB functions to find information regarding count/ per hour for total notifications and hard bounces
- Take the data and post it to redis
- Added tests


app/config.py
```diff
@@ -512,6 +512,7 @@ class Config(object):
 
     # Feature flags for bounce rate
     FF_BOUNCE_RATE_V1 = env.bool("FF_BOUNCE_RATE_V1", False)
+    FF_BOUNCE_RATE_SEED_EPOCH_TIMESTAMP = os.getenv("FF_BOUNCE_RATE_SEED_TIMESTAMP", False)
```
smcmurtry, line 515: I'm not seeing `FF_BOUNCE_RATE_SEED_EPOCH_TIMESTAMP` anywhere else in this PR - is that intentional?

Also 2 suggestions:
- let's make the variable names match (`FF_BOUNCE_RATE_SEED_EPOCH_TIMESTAMP` vs `FF_BOUNCE_RATE_SEED_TIMESTAMP`), since that is the convention for the other env variables and it makes it easier to find them in the project
- can we use a name for the FF that will make it more obvious what it's doing? maybe `FF_CALCULATE_BOUNCE_RATE` or just `FF_BOUNCE_RATE` if this is for the whole feature

app/celery/tasks.py
```diff
@@ -803,3 +806,28 @@ def acknowledge_receipt(notification_type: Any, process_type: Any, receipt: UUID
         return
     else:
         current_app.logger.error(f"acknowledge_receipt: receipt {receipt} not found in any queue")
+
+
+@notify_celery.task(name="seed-bounce-rate-in-redis")
+@statsd(namespace="tasks")
+def seed_bounce_rate_in_redis(service_id: str, interval: int = 24):
```
smcmurtry, line 813: I guess this is going to be called from `app/celery/scheduled_tasks.py` in a future PR?

app/dao/notifications_dao.py
smcmurtry, line 1: Queries look good to me!

tests/app/dao/notification_dao/test_notification_dao_bounce_rate.py
smcmurtry, line 1: looks good!

smcmurtry (approved):
Looks good! A couple questions/comments.

app/celery/tasks.py
jzbahrai: I was thinking about this and added in the latest commit that will initialize the seeded data

app/config.py
jzbahrai: ah this was for the latest commit that I hadn't pushed up. I just pushed. Happy to change the name though its not a flag... its actually epoch. So unsure if we want to say FF for a non-boolean.

app/v2/notifications/post_notifications.py
```diff
@@ -135,6 +140,24 @@ def post_precompiled_letter_notification():
...
+
+    if 0 <= time_difference_ms <= twenty_four_hours_ms:
+        # We are in the 24 hour window to seed bounce rate data
+        # Check if the seeded data key exists in redis
+        if not redis_store.get(_total_notifications_seeded_key(service_id)):
+            seed_bounce_rate_in_redis.apply_async(service_id)
+            current_app.logger.info("Seeding bounce rate for service_id {}".format(service_id))
+        else:
+            current_app.logger.info("Bounce rate already seeded for service_id {}".format(service_id))
+    current_app.logger.info("Not in the time period to seed bounce rate {}".format(service_id))
```
smcmurtry, line 157: Should this be in an else?
```suggestion
    else:
        current_app.logger.info("Not in the time period to seed bounce rate {}".format(service_id))
```

app/v2/notifications/post_notifications.py
```diff
@@ -135,6 +140,24 @@ def post_precompiled_letter_notification():
     return jsonify(resp), 201
 
 
+def _seed_bounce_data(epoch_timestamp: int, service_id: str):
+    current_time_ms = int(1000.0 * datetime.now().timestamp())
+    time_difference_ms = current_time_ms - epoch_timestamp
+
+    twenty_four_hours_ms = 86400000
```
smcmurtry, line 147: Let's write the math for this so it's easier to see where it comes from
```suggestion
    TWENTY_FOUR_HOURS_MS = 24 * 60 * 60 * 1000
```

app/v2/notifications/post_notifications.py
```diff
@@ -135,6 +140,24 @@ def post_precompiled_letter_notification():
...
+
+    if 0 <= time_difference_ms <= twenty_four_hours_ms:
+        # We are in the 24 hour window to seed bounce rate data
+        # Check if the seeded data key exists in redis
+        if not redis_store.get(_total_notifications_seeded_key(service_id)):
+            seed_bounce_rate_in_redis.apply_async(service_id)
+            current_app.logger.info("Seeding bounce rate for service_id {}".format(service_id))
+        else:
+            current_app.logger.info("Bounce rate already seeded for service_id {}".format(service_id))
```
smcmurtry, line 156: What if the key was already there with some old data that hasn't expired yet? I'm thinking if we need to run this 2 times in a 24 hour period

app/v2/notifications/post_notifications.py
```diff
@@ -135,6 +140,24 @@ def post_precompiled_letter_notification():
     return jsonify(resp), 201
 
 
+def _seed_bounce_data(epoch_timestamp: int, service_id: str):
```
smcmurtry, line 143: Just want to make sure I understand the use cases for this function:
1. When the feature is first launched we would set `epoch_timestamp = current_time - 24_hours`?
2. If we lose all our redis data we would also set this to 24 hours to recover?
3. If there is some kind of outage and we haven't been able to write to redis for, say 5 hours, we would set `epoch_timestamp = current_time - 5_hours`?

app/v2/notifications/post_notifications.py
```diff
@@ -144,6 +167,11 @@ def post_bulk():
 
     max_rows = current_app.config["CSV_MAX_ROWS"]
     check_sms_limit = current_app.config["FF_SPIKE_SMS_DAILY_LIMIT"]
+    epoch_seeding_bounce = current_app.config["FF_BOUNCE_RATE_SEED_EPOCH_MS"]
+    bounce_rate_v1 = current_app.config["FF_BOUNCE_RATE_V1"]
+    if bounce_rate_v1 and epoch_seeding_bounce:
+        _seed_bounce_data(epoch_seeding_bounce, str(authenticated_service.id))
```
smcmurtry, line 173: This is just going to seed the bounce rate in the event that a service tries to send. What about if the service does not send but is looking at their dashboard? Then they wouldn't see a valid bounce rate.
Would it make sense to move the "seeding" logic to a celery task that will get run for all services regardless of if they are sending?

smcmurtry (changes requested):

app/config.py
```diff
@@ -512,6 +512,7 @@ class Config(object):
 
     # Feature flags for bounce rate
     FF_BOUNCE_RATE_V1 = env.bool("FF_BOUNCE_RATE_V1", False)
+    FF_BOUNCE_RATE_SEED_EPOCH_MS = os.getenv("FF_BOUNCE_RATE_SEED_EPOCH_MS", False)
```
jzbahrai, line 515: put a comment explaining this.

app/v2/notifications/post_notifications.py
smcmurtry: This is actually not such a big problem since we will only use this in the event that we lose data from Redis - definitely an edge case.

app/v2/notifications/post_notifications.py
smcmurtry: Not a concern since this will only be used if there is no data in Redis

app/v2/notifications/post_notifications.py
smcmurtry: Case 3 is not relevant - we will just use this if we lose all data from Redis, and write the last 24 hours.

app/celery/tasks.py
```diff
@@ -803,3 +806,45 @@
...
+    over a given interval (default 24 hours)
+
+    Args:
+        service_id (str): The service id to seed bounce rate for
+        interval: The number of hours to seed bounce rate for
+    """
+    # TODO: uncomment once this method is implemented in notification-utils
+    # if bounce_rate_client.get_seeding_complete(service_id):
+    #     current_app.logger.info("Bounce rate already seeded for service_id {}".format(service_id))
+    #     return
```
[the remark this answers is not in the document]
jzbahrai: delete the total notifications keyand hard bounces key

app/celery/tasks.py
```diff
@@ -803,3 +806,46 @@ def acknowledge_receipt(notification_type: Any, process_type: Any, receipt: UUID
...
+def seed_bounce_rate_in_redis(service_id: str, interval: int = 24):
+    """
+    Function to seed both the total_notifications and total_hard_bounces in Redis for a given service
+    over a given interval (default 24 hours)
+
+    Args:
+        service_id (str): The service id to seed bounce rate for
+        interval: The number of hours to seed bounce rate for
+    """
+    # TODO: uncomment once this method is implemented in notification-utils
```
smcmurtry, line 822: Could delete this comment now

PR #1836, opened 2023-04-19
Fix incorrect timezone on lower bound of date range filter

whabanks:
# Summary | Résumé
This fix corrects [discrepancies between notification counts](https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/860) on the dashboard and bulk job page. We have two queries that were using date range boundaries with differing timezones resulting in the differing counts between the two pages.


andrewleith (approved):
:shipit:

PR #1837, opened 2023-04-19
Check bounce rate when sending one off and bulk emails

whabanks:
# Summary | Résumé
- Add bounce rate parameters
- Add validator for bounce rate
- One off and bulk email sends through the API now check the bounce rate and log if a service is `warning` or `critical` BR status. In V2 this will also include raising a `BadRequestError` if the service's BR is `critical`.


tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -1011,6 +1011,7 @@ def test_post_notification_with_document_upload(
         template = create_template(service=service, template_type="email", content=content)
 
         statsd_mock = mocker.patch("app.v2.notifications.post_notifications.statsd_client")
+        mocker.patch("app.notifications.validators.check_service_over_bounce_rate")
```
whabanks, line 1014: I am not certain that these `patch` calls are necessary. Tests were passing fine without them locally but [failing in CI](https://github.com/cds-snc/notification-api/actions/runs/4755087192/jobs/8449991855). When I initially added these I was getting failures but different failures. Re-ran the tests and they passed 🤷.

tests/app/notifications/test_validators.py
```diff
@@ -618,6 +619,29 @@ def test_check_service_sms_sender_id_where_sms_sender_is_not_found(sample_servic
     assert e.value.message == "sms_sender_id {} does not exist in database for service id {}".format(fake_uuid, sample_service.id)
 
 
+def test_check_service_over_bounce_rate_critical(mocker, fake_uuid):
+    mocker.patch("app.bounce_rate_client.get_bounce_rate", return_value=current_app.config["BR_CRITICAL_PERCENTAGE"])
```
andrewleith, line 623: Neat idea, very nice! 🏅

tests/app/v2/notifications/test_post_notifications.py
andrewleith: Hmm not too sure why it was inconsistent like that but it would make sense to me you would need to either mock this now that its being called in the post or else ensure whatever it needs to run is present when it does (i.e. `bounce_rate_client`, etc.)

app/notifications/validators.py
```diff
@@ -281,6 +281,16 @@ def service_can_send_to_recipient(send_to, key_type: ApiKeyType, service: Servic
         raise BadRequestError(message=message, status_code=400)
 
 
+def check_service_over_bounce_rate(service_id: str):
+    if current_app.config["FF_BOUNCE_RATE_V1"]:
+        bounce_rate = bounce_rate_client.get_bounce_rate(service_id)
+        if bounce_rate >= current_app.config["BR_CRITICAL_PERCENTAGE"]:
+            # TODO: Bounce Rate V2, raise a BadRequestError when bounce rate meets or exceeds critical threshold
```
andrewleith, line 288: This makes definitely makes sense, and we may also want to have something further up the call stack to deny notifications before they even get to this point, too.

andrewleith (approved):
LGTM!

app/config.py
```diff
@@ -609,6 +615,8 @@ class Test(Development):
 
     TEMPLATE_PREVIEW_API_HOST = "http://localhost:9999"
 
+    FF_BOUNCE_RATE_V1 = True
```
jzbahrai, line 618: we have this twice on line 576 as well

whabanks:
Temporarily blocked until Utils is updated & tagged.

app/notifications/validators.py
```diff
@@ -281,6 +281,16 @@ def service_can_send_to_recipient(send_to, key_type: ApiKeyType, service: Servic
...
+            current_app.logger.info(f"Service: {service_id} has met or exceeded a critical bounce rate threshold of 10%")
+        elif bounce_rate >= current_app.config["BR_WARNING_PERCENTAGE"]:
```
smcmurtry, line 290:
```suggestion
        elif bounce_rate_status == "warning":
```

smcmurtry (changes requested):
Here's a few suggested changes now that we have implemented some more things in utils

app/notifications/validators.py
whabanks: Will use enum.

whabanks:
Bumping utils version up to 50.3.0 along with the changes.

smcmurtry (approved):

PR #1848, opened 2023-04-28
Chore/upgrade flask-sqlalchemy 2.4.0

smcmurtry:
# Summary | Résumé

It looks like we haven't upgraded flask-sqlalchemy in 4 years, since someone pinned it to a github sha in order to use an unreleased feature: https://github.com/cds-snc/notification-api/commit/47c403f6ab17a108773c6d0229438c6b9edeb2d3

The feature is the addition of a `count: bool` arg to the `paginate` function. This feature is now live in flask-sqlalchemy 3.0.0, but we haven't been able to take an incremental approach to upgrading to it because we are stuck on this weird, unreleased version of 2.3.2.

We are only setting `count=False` in one place as far as I can tell: https://github.com/cds-snc/notification-admin/blob/main/app/notify_client/notification_api_client.py#L67

And that is on the api page where the most recent notifications (i.e. "page 1") are listed. So this change may make that page load slowly for services with a lot of notifications - we will need to test that. If that is a problem, we could add an argument that would allow the admin to specify how many notifications should be returned in the query.

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -234,7 +234,6 @@ def get_notifications_for_service(
     filter_dict=None,
     page=1,
     page_size=None,
-    count_pages=True,
     limit_days=None,
     key_type=None,
     personalisation=False,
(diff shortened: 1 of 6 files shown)


smcmurtry:
Will re-open if needed

PR #1851, opened 2023-05-01
Add ticket info logging when creating freshdesk tickets

whabanks:
# Summary | Résumé
This PR adds logging when we send Freshdesk tickets so we can better track them when troubleshooting Freshdesk related issues.

### DIFF
--- a/app/clients/freshdesk.py
+++ b/app/clients/freshdesk.py
@@ -104,7 +104,8 @@ class Freshdesk(object):
                     timeout=5,
                 )
                 response.raise_for_status()
-
+                if response.status_code == 201:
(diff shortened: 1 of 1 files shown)


app/clients/freshdesk.py
```diff
@@ -104,7 +104,8 @@ def send_ticket(self) -> int:
                     timeout=5,
                 )
                 response.raise_for_status()
-
+                if response.status_code == 201:
```
jimleroyer, line 107: I wonder if a scenario could happen where we wouldn't receive a `201` and skip this log?

jimleroyer (approved):
Let's try this out!

PR #1853, opened 2023-05-02
Fix for bug where `redis_client.active=False` for the running app

smcmurtry:
# Summary | Résumé

This is a bug fix for an issue where methods on the `BounceRateClient` were not actually saving things to Redis because they were failing this `if` check: https://github.com/cds-snc/notification-utils/blob/main/notifications_utils/clients/redis/redis_client.py#L140

We need to initialize the `BounceRateClient` with `redis_store` since that is what we are calling `.init_app()` on here: https://github.com/cds-snc/notification-api/blob/main/app/__init__.py#L130

and that sets `self.active = True`.


jimleroyer (approved):
Glad you found the issue! 🙇

PR #1857, opened 2023-05-04
Edit the Smoke test command to use poetry

jzbahrai:
# Summary | Résumé

Edit the smoke-test command to use the poetry env.


Makefile
```diff
@@ -43,7 +43,7 @@ format:
 
 .PHONY: smoke-test
 smoke-test:
-	cd tests_smoke && python smoke_test.py
+	cd tests_smoke && poetry run python smoke_test.py
```
andrewleith, line 46: I am still confused when this is necessary.  For example, my smoke tests work both ways whether the `poetry run` command precedes this or not 😕

andrewleith (approved):
Approving, with a little note below..

Makefile
whabanks: It's necessary when we run commands that depend on anything in our app's Python code. `poetry run` just executes the command passed to it inside of the project's virtual env. 
You're seeing it work correctly both with and without because Poetry tries to activate the project's virtual env on terminal start (in vscode).
There were some cases when we first started using Poetry where the env failed to activate on terminal start but I have not seen this happen recently.
Theoretically, in our dev containers, we might not need this but that rests on the assumption that the project's venv is activated prior to running commands that depend on our code so explicitly pointing to the venv helps ensure consistent behaviour.

whabanks (approved):
LGTM!

PR #1858, opened 2023-05-04
Update notification-utils to 50.3.2

smcmurtry:
# Summary | Résumé

- ran `poetry add git+https://github.com/cds-snc/notifier-utils.git#50.3.2` and updated waffles in test.yaml
- updated the calls to `set_sliding_hard_bounce` and `set_sliding_notifications` to include the notification id, and updated associated tests
- updated the timestamps used for seeding bounce rate data to be in seconds, and updated associated tests
- updated seeding logic to use notification ids
- added a call to the new `get_debug_data` function and log this

CI tests should pass.


app/celery/tasks.py
```diff
@@ -833,17 +833,21 @@ def seed_bounce_rate_in_redis(service_id: str, interval: int = 24):
...
 
     for hour, total_notifications in total_seeded_notifications:
-        hour_timestamp_ms = int(hour.timestamp() * 1000.0)
-        # generate a list of tuples of (timestamp, timestamp) that will be used to seed Redis
-        email_data = [(hour_timestamp_ms + n, hour_timestamp_ms + n) for n in range(total_notifications)]
+        # set the timestamp to the start of the hour + 1 second to ensure the notification
+        # will be counted in the correct hour
+        hour_timestamp_s = int(hour.timestamp()) + 1
+        # generate a list of tuples of (UUID, timestamp) that will be used to seed Redis
+        email_data = [(str(uuid4()), hour_timestamp_s) for _ in range(total_notifications)]
```
whabanks, line 840: Should we update `total_hard_bounces_grouped_by_hour` to also return the `notification_id` here now that we are using the `notification_id` in the sorted set?

whabanks (changes requested):

app/celery/tasks.py
smcmurtry: That would make the seeded data in Redis more realistic but I think it would come at a big performance cost for services that send a lot notifications. With the current logic, the database is doing a lot of the heavy lifting to sum notifications by hour: https://github.com/cds-snc/notification-api/pull/1858/files#diff-cca5917c620894fe770faa88835e9e186285d5b91a55fca992eb94d5dc87e635R832-R833

PR #1859, opened 2023-05-05
Use normal queue for one-off admin sends

andrewleith:
This PR makes it so one-off sends from admin use the normal queue even if the template is assigned to the bulk queue.  One-off sends for templates in the priority or normal queues are unaffected.


app/service/send_notification.py
```diff
@@ -109,7 +110,7 @@ def send_one_off_notification(service_id, post_data):
         send_notification_to_queue(
             notification=notification,
             research_mode=service.research_mode,
-            queue=template.queue_to_use(),
+            queue=QueueNames.NORMAL if template.queue_to_use() == QueueNames.BULK else template.queue_to_use(),
```
jimleroyer, line 113: Can we please leave a comment on why we want this?

andrewleith: Done, please have a look!

jimleroyer (approved):
Looks good! :shipit:

PR #1861, opened 2023-05-08
Fix bounce rate logging

smcmurtry:
# Summary | Résumé

Moved the bounce rate check to `send_to_providers.py` in celery. This allows us to call it from one place, and when we later suspend the service from here, it should work since we do not send to the provider if the service is suspended.

Tests should pass, we can QA the logs in Staging.

### DIFF
--- a/app/delivery/send_to_providers.py
+++ b/app/delivery/send_to_providers.py
@@ -173,7 +173,7 @@ def check_service_over_bounce_rate(service_id: str):
     bounce_rate = bounce_rate_client.get_bounce_rate(service_id)
     bounce_rate_status = bounce_rate_client.check_bounce_rate_status(service_id)
     debug_data = bounce_rate_client.get_debug_data(service_id)
-    current_app.logger.info(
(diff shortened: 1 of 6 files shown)


whabanks (approved):
LGTM! Just one small change to consider. :tada:

app/delivery/send_to_providers.py
```diff
@@ -165,6 +166,25 @@ def check_for_malware_errors(document_download_response_code, notification):
...
 
 
+def check_service_over_bounce_rate(service_id: str):
+    if not current_app.config["FF_BOUNCE_RATE_V1"]:
+        return
+
+    bounce_rate = bounce_rate_client.get_bounce_rate(service_id)
+    bounce_rate_status = bounce_rate_client.check_bounce_rate_status(service_id)
+    debug_data = bounce_rate_client.get_debug_data(service_id)
+    current_app.logger.debug(f"Bounce Rate: {bounce_rate} Bounce Status: {bounce_rate_status}, Debug Data: {debug_data}")
```
andrewleith, line 176: Change as per @smcmurtry in slack:
```suggestion
    current_app.logger.info(f"Service id: {service_id} Bounce Rate: {bounce_rate} Bounce Status: {bounce_rate_status}, Debug Data: {debug_data}")
```

whabanks (approved):

PR #1863, opened 2023-05-11
upgrade utils to 50.3.3 and rerun poetry lock

sastels:
# Summary | Résumé

api uses the token generating and checking code from utils to create and check team invitation urls.

Here we switch api to use the [new token code from utils](https://github.com/cds-snc/notification-utils/pull/204) that uses "token" for the salt when generating tokens, but still accepts allows tokens signed with salt DANGEROUS_SALT.

After this has been in production long enough for all old tokens to expire (min 2 days) we will push another PR to api that stops sending the salt to the utils token code. (after that, we can update the utils token code to not accept a salt as an argument)

- run with the old utils (ie api/main)
- send an invitation to your service (say, to you+testutils1@cds-snc.ca)
- verify that if you click the link you get to the "create an account" page
- Don't make a new account: if you get to this page, the link was generated and verified correctly.

- switch to this branch, and run `poetry install`
- go to your email and verify that the old link still works
- log in and sent another invitation (you+testutils2@cds-snc.ca)
- log out and go verify that the new link works

So it's possible that there could be a race condition where during deployment someone generates an invitation link with a lambda running the new code, and sends it, and the person clicks on it immediately and connects to a lambda running the old code. In this case the link would fail. The link would work if they tried a few minutes later when the system is fully deployed. Given how infrequent we generate team invites I do not think it's worth it to try to prevent this situation.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -126,14 +126,14 @@ frozenlist = ">=1.1.0"
 
 [[package]]
 name = "alembic"
-version = "1.10.3"
+version = "1.10.4"
 description = "A database migration tool for SQLAlchemy."
 category = "main"
 optional = false
 python-versions = ">=3.7"
 files = [
-    {file = "alembic-1.10.3-py3-none-any.whl", hash = "sha256:b2e0a6cfd3a8ce936a1168320bcbe94aefa3f4463cd773a968a55071beb3cd37"},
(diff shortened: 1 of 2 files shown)


jimleroyer (approved):
LGTM

PR #1864, opened 2023-05-12
Remove unneeded `s3.remove_job_from_s3` logic

smcmurtry:
# Summary | Résumé

Calling `s3.remove_job_from_s3` from the nightly task is unnecessary. We already have a policy on the bucket that all files will expire after 7 days: https://github.com/cds-snc/notification-terraform/blob/main/aws/common/s3.tf#LL17C1-L23C1

This may also have been contributing to our 4am celery errors where memory consumption would spike. 

Tests should pass in CI. You can try logging in to s3 to verify that files in the `notification-canada-ca-{env}-csv-upload` buckets have the 7 day expiry set.

### DIFF
--- a/app/aws/s3.py
+++ b/app/aws/s3.py
@@ -1,5 +1,6 @@
 import uuid
 from datetime import datetime, timedelta
+from typing import Optional
 
 import botocore
 import pytz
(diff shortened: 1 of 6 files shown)


smcmurtry:
Well this turned out to be a dead end. I was hoping to set the correct "expiry" when the file was uploaded to s3, but this turned out just to be some metadata, not the actual expiry that we want to set. I'll close this since the current method of doing this via a nightly job should work.

Actual expiry that controls when the file is deleted from s3, 7 days out from today:
Screenshot 2023-05-16 at 2 25 34 PM


Metadata expiry that is settable when the object is created in s3, 3 days out from today:
Screenshot 2023-05-16 at 2 25 45 PM

PR #1866, opened 2023-05-16
Add check to CI to ensure poetry.lock is consistent with pyproject.toml

whabanks:
# Summary | Résumé
Current we do not have safe guards in place to prevent an outdated `poetry.lock` file from being committed to main. When this occurs it can lead to our dev containers running into issues. This PR adds a check during the test CI job to ensure that the lock file is up to date and consistent with `pyproject.toml`.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -35,6 +35,8 @@ jobs:
       env:
         POETRY_VERSION: "1.3.2"
       run: pip install poetry==${POETRY_VERSION} && poetry --version
+    - name: Check poetry.lock aligns with pyproject.toml
+      run: poetry lock --check
     - name: Install requirements
       run: poetry install --with test
     - name: Run tests
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ build-backend = "poetry.core.masonry.api"
 [tool.poetry.dependencies]
 python = "~3.10.9"
 apig-wsgi = "2.15.0"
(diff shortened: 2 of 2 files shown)


.github/workflows/test.yaml
```diff
@@ -35,6 +35,8 @@ jobs:
       env:
         POETRY_VERSION: "1.3.2"
       run: pip install poetry==${POETRY_VERSION} && poetry --version
+    - name: Check poetry.lock aligns with pyproject.toml
+      run: poetry lock --check
```
andrewleith, line 39: Very nice! So this throws an error and causes this workflow to fail?  Can we test it in this branch somehow? 🤔

whabanks: It would be nice to be able to test this, it should cause an error though. I went under the assumption that the CI uses the exit code of a command to determine success or failure. When we run `poetry lock --check` and the lock file is out of sync the exit code, `echo $?`, returns 1, which maps to `Operation not permitted`. It's the same for test failures with `make test`.

whabanks: That's a good idea! I'll try it out.

whabanks: [Looks like it does the trick!](https://github.com/cds-snc/notification-api/actions/runs/4995443383/jobs/8947313443?pr=1866)

andrewleith: Nice!

andrewleith (approved):
LGTM

PR #1870, opened 2023-05-25
Improve api and callback status messages

whabanks:
# Summary | Résumé
This PR improves the language used in notifications when fetched via the API or as an API callback.


jzbahrai (approved):
it looks good to me. Is this tested in anyway? Or we can get it on staging today and try and test it ?

whabanks:
> it looks good to me. Is this tested in anyway? Or we can get it on staging today and try and test it ?

Let's merge and test on staging today since we cannot test callbacks locally.

PR #1872, opened 2023-05-28
Provide a direct link to the failed run when failure

jimleroyer:
# Summary | Résumé

If we have a failure during the execution of the `build_and_push_performance_test` GitHub action, we notify Slack with a link to the repository. This PR changes the link to refer to the failed run ID directly so that we do not have to search through our actions manually.

Make the perf workflow fail and notice the link in the Slack notification.

It`s possible it might not work. If that does not, we can replicate the strategy adopted for the URL shortener:
https://github.com/cds-snc/url-shortener/blob/main/.github/workflows/workflow_failure.yml


sastels (approved):
Nice

PR #1875, opened 2023-05-30
Improve api status messages

whabanks:
# Summary | Résumé
This PR is a follow up to https://github.com/cds-snc/notification-api/pull/1870. Instead of replacing the `status` returned for a notification with the `formatted_status` we have added a new property called `status_description` to avoid causing breakage for clients who rely on the `status` as it is currently

1. Send notifications via api with various statuses: Delivered, permanent-failure, blocked, etc
2. Fetch these notifications via the API and note that the `status_description` matches what we display for statuses in the UI
3. Set up a Callback on a service. Send various notifications either via the API or UI, observe that the `status_description` returned matches what we display in the UI.


app/models.py
```diff
@@ -1906,6 +1910,7 @@ def serialize(self) -> dict:
             "postcode": None,
             "type": self.notification_type,
             "status": self.get_letter_status() if self.notification_type == LETTER_TYPE else self.status,
+            "status_description": self.formatted_status,
```
andrewleith, line 1913: Did we want to wrap this in the feature flag as well, or maybe we dont care?

app/notifications/callbacks.py
```diff
@@ -22,6 +22,7 @@ def create_delivery_status_callback_data(notification, service_callback_api):
         "notification_client_reference": notification.client_reference,
         "notification_to": notification.to,
         "notification_status": notification.status,
+        "notification_status_description": notification.formatted_status,
```
andrewleith, line 25: I didn't realize that the data returned for the callback is not the same as the data returned in the API. Good to know.

andrewleith (approved):
Looks good - one question about whether or not we wrap this in the feature flag, but strictly speaking i dont think we HAVE to.

app/models.py
smcmurtry: I think it's okay without a feature flag - I believe the plan is just to release it right away.

smcmurtry (approved):
Looks good!

PR #1876, opened 2023-05-30
Add `province_or_territory` organisation type

smcmurtry:
# Summary | Résumé

Add `province_or_territory` organisation type.

This will allow us to identify if a service is for a P/T or not, and apply some different rules for template redaction and data retention. 

We'll also need to add it in notification-admin before we are able to set this type for an organisation.

Tests should pass, new row should be visible in the `organisation_types` table.


andrewleith (approved):
LGTM

whabanks (approved):
LGTM!

PR #1878, opened 2023-05-30
remove DANGEROUS_SALT related transition code

sastels:
# Summary | Résumé

We added code to help migrate from using DANGEROUS_SALT everywhere to using situational salts. Now that the move is complete we can delete the transitional code.

Basically kick the tires and ensure that admin still works with this api, (sending and viewing notificaiotns), and you can POST to the api.


app/dao/api_key_dao.py
```diff
@@ -56,21 +56,13 @@ def expire_api_key(service_id, api_key_id):
...
         except NoResultFound:
             pass
-
-    signed_dangerous_with_all_keys = signer_api_key.sign_with_all_dangerously_salted_keys(str(secret))
-    for signed_secret in signed_dangerous_with_all_keys:
-        try:
-            return db.on_reader().query(ApiKey).filter_by(_secret=signed_secret).options(joinedload("service")).one()
-        except NoResultFound:
-            pass
     raise NoResultFound()
```
jimleroyer, line 66: Seems like we simplify the code now:
```
        return db.on_reader().query(ApiKey).filter_by(_secret=signed_secret).options(joinedload("service")).one()
```

and remove the try...except along with the manual raise at the end.

app/encryption.py
```diff
@@ -67,35 +66,17 @@ def sign_with_all_keys(self, to_sign: str | NotificationDictToSign) -> List[str
...
-        Args:
-            to_sign (str | NotificationDictToSign): The string or dict to sign.
-
-        Returns:
-            List[str | bytes]: A list of signed values.
-        """
-        signed = []
-        for k in reversed(self.secret_key):  # reversed so that the default key is last
-            signed.append(URLSafeSerializer(k).dumps(to_sign, salt=self.dangerous_salt))
-        return signed
```
jimleroyer, line 82: 💣

app/encryption.py
```diff
@@ -67,35 +66,17 @@ def sign_with_all_keys(self, to_sign: str | NotificationDictToSign) -> List[str
...
-
     def verify(self, to_verify: str | bytes) -> Any:
         """Checks the signature of a signed value and returns the original value.
-        Currently checks against both self.salt and self.dangerous_salt.
-        After everything is signed with the new salts, this will only check against self.salt.
+        Raises BadSignature if the signature is invalid.
```
jimleroyer, line 71: I thought there was a better format for documenting raises, but it's not part of the [PEP 257](https://peps.python.org/pep-0257/). There are different conventions though depending on the community and tool as described in this [stackoverflow entry](https://stackoverflow.com/questions/3898572/what-are-the-most-common-python-docstring-formats).

app/encryption.py
```diff
@@ -107,7 +88,7 @@ def verify_unsafe(self, to_verify: str | bytes) -> Any:
         Returns:
             Any: Original value that has been signed
         """
-        return self.serializer.loads_unsafe(to_verify, salt=self.dangerous_salt)[1]
+        return self.serializer.loads_unsafe(to_verify)[1]
```
jimleroyer, line 91: 👍

app/dao/api_key_dao.py
sastels: Our SECRET_KEY is a list of 2 keys, and we have to support the api_key in the database being signed with either of them. (to continue working during the time between us changing the SECRET_KEY array to rotate a new key in and us running the resigning script).

So getting a `NoResultFound` from the first one is still ok, we could match the second time. If neither gets a match then we do have a legitimate `NoResultFound`.

(We should probably rename this to something like SIGNING_KEYS to make it less confusing)

app/encryption.py
sastels: Yeah, true enough, other places in the code such as [here](https://github.com/cds-snc/notification-api/blob/main/app/dao/notifications_dao.py#L63-L71) we put the Raise in its own section, I've made that change here.

app/dao/api_key_dao.py
jimleroyer: oh I missed the loop somehow 👍

jimleroyer (approved):

PR #1879, opened 2023-05-31
feat: update Salesforce errors with metric token

patheard:
# Summary
Add a unique token to all Salesforce error logs so that dedicated CloudWatch alarms can be created to alert the Salesforce team.

# Related
- cds-snc/platform-core-services#370


smcmurtry (approved):
Looks good!

PR #1881, opened 2023-05-31
feat: sync Salesforce Engagement ContactRoles

patheard:
# Summary
Update the Salesforce API so that as Notify service team members are added and removed, associated Engagement ContactRoles are also added or removed for each team member.

# Testing
1. With `FF_SALESFORCE_CONTACT=true`, start the local API and admin.
2. Create a new service as user `A`.
3. Add a new team member to the service (user `B`) with the permission to manage service team members.
4. Accept the invitation and login to Notify with the new team member `B`.
5. In Salesforce, expect to see ContactRoles on the service's Engagement for users `A` and `B`.
6. As user `B` in Notify, remove user `A` from the service.
7. In Salesforce, there should now only be one ContactRole on the Engagement for user `B`.

# Related
- cds-snc/platform-core-services#344

### DIFF
--- a/app/clients/salesforce/salesforce_client.py
+++ b/app/clients/salesforce/salesforce_client.py
@@ -132,3 +132,27 @@ class SalesforceClient:
             }
             salesforce_engagement.update(session, service, close_update, None, None)
         self.end_session(session)
+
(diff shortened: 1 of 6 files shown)


smcmurtry (approved):
Looks good!

PR #1882, opened 2023-06-02
use timedelta to subtract time in test

sastels:
# Summary | Résumé

Test was failing between midnight and 2 AM UTC because `now.hour() - 2` was negative. Here we subtract hours using datetime objects instead.

### DIFF
--- a/tests/app/celery/test_tasks.py
+++ b/tests/app/celery/test_tasks.py
@@ -2222,8 +2222,8 @@ class TestSeedBounceRateData:
     def test_seed_bounce_rate_data(self, mocker, notify_api):
         now = datetime.utcnow()
         current_hour = datetime(now.year, now.month, now.day, now.hour, 0)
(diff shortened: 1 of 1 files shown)


tests/app/celery/test_tasks.py
```diff
@@ -2222,8 +2222,8 @@ class TestSeedBounceRateData:
     def test_seed_bounce_rate_data(self, mocker, notify_api):
         now = datetime.utcnow()
         current_hour = datetime(now.year, now.month, now.day, now.hour, 0)
-        current_hour_minus_1 = datetime(now.year, now.month, now.day, now.hour - 1, 0)
-        current_hour_minus_2 = datetime(now.year, now.month, now.day, now.hour - 2, 0)
+        current_hour_minus_1 = current_hour - timedelta(hours=1)
+        current_hour_minus_2 = current_hour - timedelta(hours=2)
```
jimleroyer, line 2226: Nice nice 👍

PR #1884, opened 2023-06-02
Remove incorrect types from models.py

smcmurtry:
# Summary | Résumé

The type annotations that were included here (mostly by me) are actually incorrect. These should all be `sqlalchemy.Column` types like this:
Screenshot 2023-06-02 at 4 09 44 PM

They are not just `UUID` or `int` or `str` - they have lots of special methods that can be used in queries, and we will get type errors inside our DAO files if we don't use the correct type.

Tests should pass in CI.


smcmurtry:
@andrewleith I don't think there's much to test except that the type check passes in CI - you can run it locally too with `poetry run mypy ./` from the root directory.

andrewleith (approved):
:shipit:

PR #1885, opened 2023-06-02
Add better types to models.py

smcmurtry:
# Summary | Résumé

Add better types to models.py. 

How this works:
- the `db` object in models.py is an instance of the `RoutingSQLAlchemy` class defined here: https://github.com/cds-snc/notification-api/blob/main/app/dbsetup.py#L70
- `RoutingSQLAlchemy` has `flask_sqlalchemy.SQLAlchemy` as a base class
- `flask_sqlalchemy.SQLAlchemy` has the following in its docstring: 
> This class also provides access to all the SQLAlchemy functions and classes from the `sqlalchemy` and `sqlalchemy.orm` modules
- most of the things that are missing types in models.py can be found in these two paths, for instance `db.Column` is of type `sqlalchemy.Column`, `db.Boolean` is of type `sqlalchemy.Boolean`, `db.relationship` is of type `sqlalchemy.orm.relationship`
- so I just imported all the classes from these two paths into a new file, app/db_type.py, and assigned the `db` object to be that type
- that populates most of the types in models.py:

Before:
Screenshot 2023-06-02 at 4 33 30 PM
Screenshot 2023-06-02 at 4 36 10 PM

After:
Screenshot 2023-06-02 at 4 33 54 PM
Screenshot 2023-06-02 at 4 35 13 PM

Tests should pass in CI, check out and try the new type hints locally!


app/db_type.py
```diff
@@ -0,0 +1,2 @@
+from sqlalchemy import *  # noqa: F401,F403
```
jimleroyer, line 1: Let's please document the purpose of this file with a docstring, explaining why this is necessary.

app/models.py
```diff
@@ -50,6 +51,7 @@
 
 TEMPLATE_TYPES = [SMS_TYPE, EMAIL_TYPE, LETTER_TYPE]
 
+db = cast(db_type, db)  # type: ignore
```
jimleroyer, line 54: And a comment to explain why we need to cast it to the module's type. 👍

smcmurtry:
Will re-open if needed

PR #1888, opened 2023-06-05
Fix bug where template redaction for P/Ts is flaky

smcmurtry:
# Summary | Résumé

After testing creating new templates on staging for P/T services, I observed that the templates were not always auto-redacted and I found `Template not found for redaction.` in the logs. This is logged when we try to fetch the template by id, and we get a `NoResultFound` error: https://github.com/cds-snc/notification-api/blob/main/app/template/rest.py#L106

This indicated that the template has not been created in the db yet and that is causing the redaction to fail. Instead, we can just set the `redact_personalisation` property to `True` when we create the template, and that seems to work.

I removed the optional `tempate_id` argument from `create_template`, since I had added that in the original PR for this feature and it is no longer needed: https://github.com/cds-snc/notification-api/pull/1873

Local test:
- run this branch of notification-api locally, and the main branch of notification-admin locally
- log in as a platform admin and create a new organisation "Province of Ontario"
- go to the settings page for this new org and set it as type "Provincial or Territorial Government"
- create a new service
- visit the service's settings page and set the org as "Province of Ontario"
- create a new template for the service and save it
- observe that the template page shows the text "Recipients' information will be redacted from system"

### DIFF
--- a/app/template/rest.py
+++ b/app/template/rest.py
@@ -96,15 +96,8 @@ def create_template(service_id):
 
     check_reply_to(service_id, new_template.reply_to, new_template.template_type)
 
-    template_id = uuid.uuid4()
-    dao_create_template(new_template, template_id=template_id)
-
-    if service_owned_by_a_province_or_territory(service_id):
-        try:
-            fetched_template = dao_get_template_by_id_and_service_id(template_id=template_id, service_id=service_id)
-            dao_redact_template(fetched_template, template_json["created_by"])
-        except NoResultFound:
-            current_app.logger.error(f"Template not found for redaction. service_id: {service_id}, template_id: {template_id}")
+    redact_personalisation = service_owned_by_a_province_or_territory(service_id)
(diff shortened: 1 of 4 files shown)


app/template/rest.py
```diff
@@ -60,12 +58,11 @@ def validate_parent_folder(template_json):
...
 
 
-def service_owned_by_a_province_or_territory(service_id: str) -> bool:
-    organisation = dao_get_organisation_by_service_id(service_id=service_id)
+def should_template_be_redacted(organisation: Organisation) -> bool:
     try:
         return organisation.organisation_type == "province_or_territory"
-    except AttributeError:
-        # service has no organisation
+    except Exception as e:
```
sastels, line 64: probably should change this back to `AttributeError`

sastels (changes requested):

sastels:
Works locally, just had that one change request

sastels (approved):
🚀

PR #1889, opened 2023-06-06
fix: errors when adding/removing ContactRoles

patheard:
# Summary
Update the Salesforce logic that adds or removes
Engagement ContactRoles to only attempt to parse
the request result if an operation has been performed.

This will fix the Salesforce errors being logged by the API when
it was attempting to parse results that were in an 
unexpected format.

# Related
- cds-snc/platform-core-services#344


smcmurtry (approved):
Looks good!

PR #1892, opened 2023-06-08
bug fix KeyErrors from AWS sending domain verification

sastels:
# Summary | Résumé

Getting lots of KeyErrors raised here
https://github.com/cds-snc/notification-api/blob/main/app/notifications/notifications_ses_callback.py#L39

Believe it is because of the recently added new sending domains. AWS does verification calls when you first create them.

This causes us to try to process an AWS response of "AmazonSnsSubscriptionSucceeded" which our code cannot understand so it throws an error.

Note that if these are related to the new sending domains then these are not associated with notifications, and so our system should not be processing them.

This PR moves the query for the associated notification in front of the response parsing, so if there is no related notification we will not do the parsing. This means that in this case we will not get to the parsing code and not throw a KeyError.

### DIFF
--- a/app/celery/process_ses_receipts_tasks.py
+++ b/app/celery/process_ses_receipts_tasks.py
@@ -50,9 +50,7 @@ def process_ses_results(self, response):
                 )
                 self.retry(queue=QueueNames.RETRY)
             except self.MaxRetriesExceededError:
-                current_app.logger.warning(
-                    f"notification not found for SES reference: {reference}. Giving up."
-                )
+                current_app.logger.warning(f"notification not found for SES reference: {reference}. Giving up.")
             return
 
         aws_response_dict = get_aws_responses(ses_message)
--- a/app/celery/process_ses_receipts_tasks.py
+++ b/app/celery/process_ses_receipts_tasks.py
@@ -38,9 +38,6 @@ def process_ses_results(self, response):
             _check_and_queue_complaint_callback_task(*handle_complaint(ses_message))
             return True
 
-        aws_response_dict = get_aws_responses(ses_message)
-
-        notification_status = aws_response_dict["notification_status"]
         reference = ses_message["mail"]["messageId"]
 
         try:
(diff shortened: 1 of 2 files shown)


app/celery/process_ses_receipts_tasks.py
```diff
@@ -38,26 +38,23 @@ def process_ses_results(self, response):
...
                 self.retry(queue=QueueNames.RETRY)
             except self.MaxRetriesExceededError:
-                current_app.logger.warning(
-                    f"notification not found for SES reference: {reference} (update to {notification_status}). Giving up."
-                )
+                current_app.logger.warning(f"notification not found for SES reference: {reference}. Giving up.")
             return
 
+        aws_response_dict = get_aws_responses(ses_message)
+        notification_status = aws_response_dict["notification_status"]
```
jimleroyer, line 57: In any case, late parsing fields that might not be present is much better. 👍

app/celery/process_ses_receipts_tasks.py
```diff
@@ -38,26 +38,23 @@ def process_ses_results(self, response):
...
-        notification_status = aws_response_dict["notification_status"]
         reference = ses_message["mail"]["messageId"]
 
         try:
             notification = notifications_dao.dao_get_notification_by_reference(reference)
         except NoResultFound:
             try:
                 current_app.logger.warning(
-                    f"RETRY {self.request.retries}: notification not found for SES reference {reference} (update to {notification_status}). "
+                    f"RETRY {self.request.retries}: notification not found for SES reference {reference}. "
```
jimleroyer, line 48: Arr losing these log info though... we could try a `get` instead with a default param value, to get an overridden None?

sastels: yeah this annoyed me too. Figured making the smallest hot fix for now might be best, and followup with maybe thinking about how we might want to deal with the unparsed response type later.

sastels: The exception is happening here
https://github.com/cds-snc/notification-api/blob/main/app/notifications/notifications_ses_callback.py#L39
basically if the response isn't Delivery", "Complaint", or "Bounce" then we don't know how to treat it and we raise an exception.

jimleroyer: Yes let's bring a follow up PR please to do this properly. 🙏

jimleroyer (approved):

PR #1893, opened 2023-06-14
Add status_description to the callback data

smcmurtry:
# Summary | Résumé

The "status_description" wasn't appearing in the data sent to the callback url. This PR should fix that.

# Testing

Callbacks can't be tested locally because we don't get the receipts from AWS. So we will have to do manual testing in Staging and rely on our test suite here.


andrewleith (approved):
LGTM!

PR #1894, opened 2023-06-15
add dimension values to metric to track any expired in-flights

sastels:
# Summary | Résumé

re [Test and fix the in-flight expired alarms](https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/136)

Add "any" values to our metric around expired inflights. This will allow us to add alarms that notify us if there have been any expired in-flights.

# Test instructions | Instructions pour tester la modification

When in staging, [trigger an expired in-flight](https://docs.google.com/document/d/16LLelZ7WEKrnbocrl0Az74JqkCv5DBZ9QILRBUFJQt8/edit#heading=h.uh6oq7aubepd) and see if you can see it in the "any, any" metric!

# Release Instructions | Instructions pour le déploiement

This will allow us to fix the expired in-flights alarm.

PR #1895, opened 2023-06-15
Fix expired inflight metric

sastels:
# Summary | Résumé

[This](https://github.com/cds-snc/notification-api/pull/1894) didn't work - I think because we already flushed() the metric, we need to use a new one.

try it again on staging...


ben851 (approved):
To the moon!

PR #1896, opened 2023-06-15
Add an email character limit

smcmurtry:
# Summary | Résumé

Draft for now.

TBD


whabanks (approved):
Looks good, tested locally with the instructions and it works as expected! Are we waiting on translations for the new error message in Admin?

PR #1897, opened 2023-06-16
Log the exception when there's a problem sending a file

sastels:
# Summary | Résumé

Had a celery error
https://gcdigital.slack.com/archives/C012W5K734Y/p1686873733012959

Would be good to see what the actual exception is in this case.


whabanks (approved):
LGTM! Thanks for throwing this together!

PR #1898, opened 2023-06-16
feat: update engagement when service name changes

patheard:
# Summary
Update the Salesforce Engagement name when its linked Notify service name changes.

# Testing
1. Create a new Notify service and check that a linked Salesforce Engagement was created.
2. Update the service's name in Notify.
3. Check that the Salesforce Engagement name was also updated.

# Related
- https://github.com/cds-snc/platform-core-services/issues/343


smcmurtry (approved):
Looks good 👍

PR #1904, opened 2023-06-20
Enforce email character limit at send time

smcmurtry:
# Summary | Résumé

This PR:
- adds a step to validators.py where the email character length is checked

Try sending some emails through the API that are more than 50k characters. You won't be allowed to create a template this long, so instead you can create a template with a single variable, and make the variable content more than 50k characters:

1. Run this branch locally and connect to locally running Admin
2. create a template that is a single variable
3. send a post request to `'http://host.docker.internal:6011/v2/notifications/email'` containing a a variable that is longer than 50k characters, like: `data = {"email_address": "you@cds-snc.ca", "template_id": "xyz", "personalisation": {"variable": "x" * 50001}}`
4. observe that you receive the response:
```{
  "errors": [
    {
      "error": "BadRequestError",
      "message": "Content for template has a character count greater than the limit of 50000"
    }
  ],
  "status_code": 400
}
```


app/notifications/validators.py
```diff
@@ -351,10 +358,11 @@ def validate_template(template_id, personalisation, service: Service, notificati
     check_template_is_for_notification_type(notification_type, template.template_type)
     check_template_is_active(template)
 
-    template_with_content: Template = create_content_for_notification(template, personalisation)
+    template_with_content = create_content_for_notification(template, personalisation)
```
jimleroyer, line 361: Does this hit the database on every notification?

smcmurtry: Not this line, but higher up in the `validate_template` function: https://github.com/cds-snc/notification-api/blob/main/app/notifications/validators.py#L350

Which gets the template from the db here: https://github.com/cds-snc/notification-api/blob/main/app/notifications/validators.py#L365

jimleroyer: Oh so we are not adding one more retrieval, as the template was already retrieved from the database prior to these changes?

smcmurtry: Yes that's correct!

smcmurtry:
Closing for now - this is probably overkill.

PR #1905, opened 2023-06-20
Add some logging so we know when requests are made to the V1 api

smcmurtry:
# Summary | Résumé

Add some logging so we know when requests are made to the V1 api. Is anyone using it? We will find out!


jimleroyer (approved):
Thanks @smcmurtry !

PR #1906, opened 2023-06-22
Add logging when the email char limit is exceeded

smcmurtry:
# Summary | Résumé

Add logging when the email char limit is exceeded

### DIFF
--- a/app/template/rest.py
+++ b/app/template/rest.py
@@ -101,6 +101,8 @@ def create_template(service_id):
         char_limit = SMS_CHAR_COUNT_LIMIT if new_template.template_type == SMS_TYPE else EMAIL_CHAR_COUNT_LIMIT
         message = "Content has a character count greater than the limit of {}".format(char_limit)
         errors = {"content": [message]}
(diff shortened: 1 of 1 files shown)


app/template/rest.py
```diff
@@ -101,6 +101,8 @@ def create_template(service_id):
         char_limit = SMS_CHAR_COUNT_LIMIT if new_template.template_type == SMS_TYPE else EMAIL_CHAR_COUNT_LIMIT
         message = "Content has a character count greater than the limit of {}".format(char_limit)
         errors = {"content": [message]}
+        if new_template.template_type == EMAIL_TYPE:
+            current_app.logger.info({"error": "email_char_count_exceeded", "message": message, "service_id": service_id})
```
jimleroyer, line 105: This could be a `warn` level log. We could log it for SMS as well, if we want to build stats on these from logs.

jimleroyer (approved):
LGTM -- left a comment on log level and sms inclusion if we want to as well.

PR #1911, opened 2023-06-26
Sending capacity tests

sastels:
# Summary | Résumé

- add a email address that, at the last moment, will not be sent to SES but just marked as "sending" and a mock response sent to the callback task for processing. (we already had a phone number to do the same for SMS)

- added a new locust test to POST repeatedly to `/email` or `/sms` so that we can flood the system and see our send rates without any worries about being slowed down by AWS.

Run locally with something like
```
locust -f ./internal_stress_test.py --host=http://localhost:6011 --type=email
```
Note that in the celery logs you should see lines such as
```
[2023-06-26 19:46:31,189: INFO/ForkPoolWorker-3] notification 4b1247f1-7fb3-4651-a94d-223beee8d91c sending to internal test email address. Not sending to AWS
```


ben851 (approved):
Looks fun!

PR #1912, opened 2023-06-26
Prevent duplicate notifications during job re-processing

whabanks:
# Summary | Résumé
This PR implements a potential solution to eliminate duplicate notifications from being sent. These duplicates result from jobs that slow down due to high load on notify. More info in the [card for this issue](https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/79)

- A new task, `update-job`, runs every minute that sets the job's `updated_at` field to the `updated_at` field of the most recently `sent` notification within that job.
- `check_job_status` now relies on the job's `updated_at` field to determine if it's timeout period has been reached and the job needs to be aborted and retried.
- `check_job_status` timeout period reduced from 2 hours to 1 as the increase to 2 hours was an intermediary fix until the root cause could be identified and resolved.

Using the `processing_started` time when calculating if a job is stuck is not accurate as job processing may slow in relation to the load on Notify. If the job is aborted and retried mid processing then an edge case where the notification has been sent but not yet persisted results in the retry task picking up an resending that notification. 
By using the time of the last notification that a job sent as the start time for the timeout period, we can reliably make the distinction between job processing that is truly stuck, and processing that is just slow. 

# Testing
## Verify reproducibility of duplicate emails
1. Check out main 
2. In `scheduled_tasks.py` edit L141 & 142 and reduce the timeout to something like 2 or 3 minutes
3. Prepare a bulk send via the API using a template that contains a variable and set the variable to 1, 2, 3...etc. This is helpful  for identifying the duplicate
4. Set a breakpoint on L191 of `notifications_dao.py`
5. Send the bulk job
6. When the breakpoint is hit, wait however long you set the timeout to in step 2 then continue.
7. Note that you received all of the emails in the bulk send + 1 extra

If the above process resulted in a duplicate email on main, check out this feature branch and repeat the steps. You should receive no duplicate emails. If unable to reproduce the duplicate email with the above steps, then we may need to revisit how to properly test this scenario.


smcmurtry:
Okay I've made a bit of progress on testing this with the debugger:

By the time we get to L191 of notifications_dao.py, the job is already marked as finished in the job table in the db. This seems weird since no notifications have been sent at this point. Because the job has already been marked as finished at this point, pausing here will not trigger the `check_job_status` in scheduled_tasks.py to restart the job since that task only checks for jobs that are still in progress.

smcmurtry:
It looks like this is a bug - the job is being marked as "finished" too early.
In `process_job` and `process_incomplete_job` in `tasks.py`, we have the following flow:

1. call `process_rows` on all chunk of the total rows for the job
2.  mark the job as complete with `job_complete`

This probably worked correctly if `process_rows` used to not be async. But now `process_rows` is async because it calls `save_emails.apply_async` or `save_smss.apply_async`. So those processes can still be in progress at the time we mark the job as complete.

smcmurtry (changes requested):
I don't think this is going to work unless we first fix the bug where the job is marked `"complete"` too early.

smcmurtry:
Also - I was not able to reproduce the "duplicate notifications" bug using the steps above.

smcmurtry (approved):
Code looks good. I created a new issue for [jobs getting marked finished to soon](https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/1275).

app/celery/tasks.py
```diff
@@ -87,6 +88,15 @@
 )
 
 
+@notify_celery.task(name="update-job")
+@statsd(namespace="tasks")
+def update_job(job_id):
```
jimleroyer, line 93: The celery task receives a job ID here..

PR #1919, opened 2023-06-30
Set `--concurrency=4` for celery debugger

smcmurtry:
# Summary | Résumé

I needed this to debug this PR: https://github.com/cds-snc/notification-api/pull/1912

We probably want to run the celery debugger like this all the time since that is how we are running it in production (I think?).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,7 +66,7 @@ docs/_build/
 # PyBuilder
 target/
 .idea/
-.vscode
+.vscode/settings.json
 
 # Mac
 *.DS_Store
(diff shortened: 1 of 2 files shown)


whabanks (approved):
LGTM! Celery defaults this config to the # of cores that the host machine's CPU has. In my case that would be 10. I am curious why 4 specifically?

smcmurtry:
@whabanks I wanted it to match the setting here: https://github.com/cds-snc/notification-api/blob/main/scripts/run_celery.sh#L7

I'm not sure if that is actually what we're using in prod - I couldn't find that code

PR #1921, opened 2023-07-05
Upgrade flask and werkzeug

sastels:
# Summary | Résumé

upgrade flask / werkzeug / utils

See https://github.com/cds-snc/notification-utils/pull/222

There were 2 main issues with api:
- flask got stricter about when `before_request` can be called
- the way we tweaked `json.dumps()` to encode `Row`s is no longer supported.

- POST some emails / sms. Works with or without newest admin / dd-api


app/__init__.py
```diff
@@ -158,6 +158,15 @@ def create_app(application, config=None):
     return application
 
 
+def register_notify_blueprint(application, blueprint, auth_function, prefix=None):
+    if not blueprint._got_registered_once:
+        blueprint.before_request(auth_function)
+        if prefix:
+            application.register_blueprint(blueprint, url_prefix=prefix)
+        else:
+            application.register_blueprint(blueprint)
```
sastels, line 167: So before if a blueprint was already registered and you called `before_request()` on it, the `before_request()` would basically be ignored. Now it's an explicit error (which is probably better than just ignoring people's code).

The app doesn't care, this code only runs once on boot, but the tests run this repeatedly and, essentially, the second test that got run would fail with the error.

app/__init__.py
```diff
@@ -194,80 +203,55 @@ def register_blueprint(application):
     )
     from app.user.rest import user_blueprint
 
-    service_blueprint.before_request(requires_admin_auth)
-    application.register_blueprint(service_blueprint, url_prefix="/service")
+    register_notify_blueprint(application, service_blueprint, requires_admin_auth, "/service")
```
sastels, line 206: So if I added all the `_got_registered_once` checks in this file, the formatter complains that the method is too complex. Which is fair. So I refactored out how we create blueprints to the above function.

It is very important that in the lines below we keep the correct auth and prefex for the blueprint. Another set of 👀 going over each of these would be appreciated.

andrewleith:
In addition to manual testing, testing this with the cypress suite of tests, and they are all passing:
image
image

Going to look at the PR in more details now...

andrewleith (approved):
All tests are passing for me, everything seems to work properly locally using all 4 branches.  Moving on to the dd-api PR.

PR #1927, opened 2023-07-12
Use Celery chains, ensuring job_complete is called after a job completes processing

whabanks:
# Summary | Résumé
This PR implements [Celery Chains](https://docs.celeryq.dev/en/stable/reference/celery.html#celery.chain) to ensure, when we call the `save_smss|emails` tasks during job processing, that we only call `job_complete` once the aforementioned save tasks have completed. This should prevent jobs from being marked complete before they actually are.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -1,8 +1,10 @@
 import json
+from ast import Str
 from collections import namedtuple
 from datetime import datetime
 from typing import Any, Dict, List, Optional
 from uuid import UUID, uuid4
+from celery import chain
 
 from flask import current_app
 from itsdangerous import BadSignature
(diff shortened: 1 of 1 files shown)


app/celery/tasks.py
```diff
@@ -192,15 +195,17 @@ def process_rows(rows: List, template: Template, job: Job, service: Service):
     # the same_sms and save_email task are going to be using template and service objects from cache
     # these objects are transient and will not have relationships loaded
     if encrypted_smss:
-        save_smss.apply_async(
```
whabanks, line 195: Here we're wrapping the call to the `save_...` tasks in a chain with the `job_complete` task such that they are executed sequentially. By calling `apply_async()` on the chain itself we can preserve the async behaviour we had previously.

app/celery/tasks.py
```diff
@@ -192,15 +195,17 @@ def process_rows(rows: List, template: Template, job: Job, service: Service):
...
+        chain(save_smss.s(
             (str(service.id), encrypted_smss, None),
-            queue=choose_database_queue(str(template.process_type), service.research_mode, job.notification_count),
-        )
+            queue=choose_database_queue(str(template.process_type), service.research_mode, job.notification_count)),
+            job_complete.s(job.id)
```
smcmurtry, line 201: `process_rows` will get called multiple times per job for larger jobs, once on each chunk of 500 rows. So we can't call the job complete after 1 chunk has finished processing.

whabanks:
Closing as this work has been superseded by: https://github.com/cds-snc/notification-api/pull/1931

PR #1928, opened 2023-07-12
Add some type annotations for celery tasks

smcmurtry:
# Summary | Résumé

When we create a celery task using the decorator like this:
```
@notify_celery.task(bind=True, name="save-smss", max_retries=5, default_retry_delay=300)
def save_smss( ...
```

It does not have the correct type information when we import it in another file, making it hard to verify that we are using methods correctly. This PR uses `typing.cast` to set it to the correct type.

The intellisense should now be working for `save_smss` and `save_emails` in `shceduled_tasks.py`:
Screenshot 2023-07-12 at 4 23 02 PM


app/celery/scheduled_tasks.py
```diff
@@ -42,6 +42,11 @@
 )
 from app.notifications.process_notifications import send_notification_to_queue
 from app.v2.errors import JobIncompleteError
+from celery import Task
+
+# https://stackoverflow.com/questions/63714223/correct-type-annotation-for-a-celery-task
+save_smss = cast(Task, save_smss)
```
andrewleith, line 48: It seems wrong to me to need to add a library into our production code to make vscode give correct type hints in python 🤷

PR #1931, opened 2023-07-17
New celery task for marking jobs as complete

smcmurtry:
# Summary | Résumé

This PR:
- removes the `job_complete` calls from `process_job` and `process_incomplete_job` since these calls were happening before all notifications had been added to the db, and update the associate tests
- adds a new periodic celery task that runs every minute called `mark-jobs-complete`. This queries for "in progress" jobs and checks if the notifications for that job exist in the notifications table yet. If the do exist, the job is marked "complete"
- tests for `mark_jobs_complete`
- adds a new function to `notifications_dao.py` called `get_notification_count_for_job`, as well as tests for this

1. Run `./scripts/run_celery_purge.sh` locally
2. run notification-api locally using this branch (`fix/celery-task-job-complete`), also run celery
3. point notification-admin to your local api and run it locally too
4. open a Postgres client like Postico and connect to your local db, open the `jobs` table
5. send a bulk job via the api of maybe 10 emails (needs to be via api because CSV upload doesn't work locally)
6. refresh the `jobs` table in Postico and observe that a new row has appeared for the job you just sent via API
7. observe that the `job_status` column is set to either `pending` or `in progress`
8. wait 1 minute
9. refresh the `jobs` table and observe that `job_status` now has a value of `finished`

### DIFF
--- a/app/dao/notifications_dao.py
+++ b/app/dao/notifications_dao.py
@@ -283,6 +283,11 @@ def get_notifications_for_job(service_id, job_id, filter_dict=None, page=1, page
 
 
 @statsd(namespace="dao")
+def get_notification_count_for_job(service_id, job_id):
+    return Notification.query.filter_by(service_id=service_id, job_id=job_id).count()
+
+
+@statsd(namespace="dao")
 def get_notification_with_personalisation(service_id, notification_id, key_type):
     filter_dict = {"service_id": service_id, "id": notification_id}
     if key_type:
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -16,7 +16,7 @@ from app import (
     sms_priority,
     zendesk_client,
 )
-from app.celery.tasks import process_job, save_emails, save_smss
+from app.celery.tasks import job_complete, process_job, save_emails, save_smss
 from app.config import QueueNames, TaskNames
 from app.dao.invited_org_user_dao import (
     delete_org_invitations_created_more_than_two_days_ago,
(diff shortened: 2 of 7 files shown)


app/dao/notifications_dao.py
```diff
@@ -282,6 +282,11 @@ def get_notifications_for_job(service_id, job_id, filter_dict=None, page=1, page
     return query.order_by(asc(Notification.job_row_number)).paginate(page=page, per_page=page_size)
 
 
+@statsd(namespace="dao")
+def get_notification_count_for_job(service_id, job_id):
+    return Notification.query.filter_by(service_id=service_id, job_id=job_id).count()
```
jimleroyer, line 287: I checked as I was curious and this can actually use the two [separate indexes](https://use-the-index-luke.com/sql/explain-plan/postgresql/operations) on service_id and job_id:

```
Aggregate  (cost=43376.36..43376.37 rows=1 width=8)
  ->  Bitmap Heap Scan on notifications n  (cost=36194.63..43319.72 rows=22656 width=0)
        Recheck Cond: (job_id = '551beb20-3561-4051-b548-a7dab97f1b63'::uuid)
        Filter: (service_id = '1c31507f-a7ec-4102-9cbe-c648388c1e8a'::uuid)
        ->  Bitmap Index Scan on ix_notifications_job_id  (cost=0.00..36188.97 rows=49673 width=0)
              Index Cond: (job_id = '551beb20-3561-4051-b548-a7dab97f1b63'::uuid)
```

But most likely won't use these, as the jobs are 50K rows and too small, whereas an seq scan [will be less costly](https://www.gojek.io/blog/the-case-s-of-postgres-not-using-index).

jimleroyer:
This is a good catch, as we deferred the saving of notifications into async celery tasks (which it was not before), marking the jobs complete could trigger too early then. 👏

whabanks (approved):
LGTM! Tested locally following the provided steps. In addition I introduced a delay during job processing to test the new task a bit more. The job status was not marked complete early.

PR #1936, opened 2023-07-18
Prevent duplicate notifications during job re-processing

whabanks:
# Summary | Résumé
This PR amends the issues discovered with the [initial implementation](https://github.com/cds-snc/notification-api/pull/1912).
- `update_jobs` -> `update_in_progress_jobs`, no longer takes in a `job_id`, and is called outside the `dao_set_scheduled_jobs_to_pending` loop in `run_scheduled_jobs`
- `update_in_progress_jobs` now checks all jobs with a status of `in-progress` in the last 5 minutes, i.e. since the last time the `run_scheduled_jobs` task is called by the heartbeat.


app/celery/scheduled_tasks.py
```diff
@@ -137,30 +143,30 @@ def check_job_status():
...
     and template_type in ('sms', 'email')
-    and scheduled_at or created_at is older that 120 minutes.
+    and scheduled_at or created_at is older than 60 minutes.
     if any results then
         raise error
         process the rows in the csv that are missing (in another task) just do the check here.
     """
-    minutes_ago_120 = datetime.utcnow() - timedelta(minutes=120)
-    minutes_ago_125 = datetime.utcnow() - timedelta(minutes=125)
+    minutes_ago_60 = datetime.utcnow() - timedelta(minutes=60)
```
whabanks, line 151: I reverted this from 120 minutes back to 60. Originally this was 30 minutes and we bumped it up to [60](https://github.com/cds-snc/notification-api/commit/db955f3038742b641af5997cd2a1ed1e96bd9520) then to [120](https://github.com/cds-snc/notification-api/commit/6f71d29a8c60dcadee9562ea624d0c34171a41e9) as a bandaid fix for the issue this PR solves. Perhaps at some point we could revert back to 30 minutes as well?

jimleroyer: We could do 30 minutes now to eagerly fail: if this is deemed to fail even with the new fix, we'd know sooner than later.

app/celery/scheduled_tasks.py
```diff
@@ -53,6 +58,7 @@
 @statsd(namespace="tasks")
 def run_scheduled_jobs():
     try:
+        update_in_progress_jobs.apply_async(queue=QueueNames.JOBS)
```
jimleroyer, line 61: We really want this to run async? the updates won't take effect for the code that follows. I am not sure myself here, as I didn't dig deep into the process, but if following code depends the job latest updated values, it's possible it won't get it.

app/celery/tasks.py
```diff
@@ -87,6 +88,16 @@
 )
 
 
+@notify_celery.task(name="update-job")
+@statsd(namespace="tasks")
+def update_in_progress_jobs():
+    jobs = dao_get_in_progress_jobs(datetime.utcnow() - timedelta(minutes=5))
```
jimleroyer, line 94: I wonder if that 5 minutes restriction could not bite us later. What if a beat task is skipped, due to redeployment? would that still be ok? What is the reasoning for picking 5 minutes?

whabanks: The idea was to speed things up by only looking in the time window since the last heartbeat, the last time it was called. However in testing this morning I did run into the issue you are describing. I'll remove this for now.

app/celery/scheduled_tasks.py
whabanks: Agreed, if the task to re-process the job runs before this task finishes updating the `updated_at` field of the job we'll end up in the same position. I'll change this to use `delay()` instead.

app/celery/scheduled_tasks.py
whabanks: Reverted back to 30 minutes!

app/celery/tasks.py
```diff
@@ -87,6 +88,17 @@
...
```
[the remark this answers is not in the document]
whabanks: fixed!

app/celery/tasks.py
smcmurtry: Yeah I think we should run this on all "in progress" jobs. Or if we want to get fancy, we could run it on "in progress" jobs where the last update time is more than 10 or 20 minutes ago. That way we would catch the jobs this feature is concerned with - the ones that may be passed to `process_incomplete_jobs`

tests/app/dao/notification_dao/test_notification_dao.py
```diff
@@ -570,6 +572,27 @@ def test_get_all_notifications_for_job(sample_job):
     assert len(notifications_from_db) == 5
 
 
+def test_get_latest_sent_notification_for_job_partially_processed_job(sample_job):
```
smcmurtry, line 575: This is a good test but it took me a little while to understand what the test is doing. Would be a bit easier to understand with the data written out in an array:

```
one_s = timedelta(seconds=1)
now = datetime.utcnow()
test_data = [
   (now - 5 * one_s, "sent"),
   (now - 4 * one_s, "sent"),
   (now - 3 * one_s, "sent"),
   (now - 2 * one_s, "pending"),
   (now - 1 * one_s, "pending"),
   (now, "sent"),
]
for updated_at, status in test_data:
   ...

Comment (whabanks) [reply tests/app/dao/notification_dao/test_notification_dao.py:575]: Great idea, I'll adjust this test.

File: app/dao/notifications_dao.py [hunk 6]
```diff
@@ -751,6 +751,16 @@ def dao_get_total_notifications_sent_per_day_for_performance_platform(start_date
     )
 
 
+@statsd(namespace="dao")
+def get_latest_sent_notification_for_job(job_id):
+    return (
+        Notification.query.filter(Notification.job_id == job_id, Notification.status == "sent")
```
smcmurtry, line 757: I think we don't need to filter on `Notification.status == "sent"` here. If the notification has been added to the db then that is good enough

whabanks: Fixed!

app/celery/scheduled_tasks.py
```diff
@@ -54,6 +60,7 @@
 @statsd(namespace="tasks")
 def run_scheduled_jobs():
     try:
+        update_in_progress_jobs.delay(queue=QueueNames.JOBS)
```
smcmurtry, line 63: It looks like `.delay` actually does an async call to the task: https://docs.celeryq.dev/en/stable/userguide/calling.html

I think we just want to call `update_in_progress_jobs()` directly. In fact, we could just make `update_in_progress_jobs` a regular function and not a celery task.

app/celery/scheduled_tasks.py
```diff
@@ -165,30 +172,30 @@ def check_job_status():
...
-    and scheduled_at or created_at is older that 120 minutes.
+    and scheduled_at or created_at is older than 60 minutes.
     if any results then
         raise error
         process the rows in the csv that are missing (in another task) just do the check here.
     """
-    minutes_ago_120 = datetime.utcnow() - timedelta(minutes=120)
-    minutes_ago_125 = datetime.utcnow() - timedelta(minutes=125)
+    minutes_ago_30 = datetime.utcnow() - timedelta(minutes=30)
+    minutes_ago_35 = datetime.utcnow() - timedelta(minutes=35)
```
smcmurtry, line 181: Can we go with 60 and 65 minutes here?

whabanks: I reverted this from 120 minutes back to 30. Originally this was 30 minutes and we bumped it up to [60](https://github.com/cds-snc/notification-api/commit/db955f3038742b641af5997cd2a1ed1e96bd9520) then to [120](https://github.com/cds-snc/notification-api/commit/6f71d29a8c60dcadee9562ea624d0c34171a41e9) as bandaid fixes for the issue this PR is solving.

smcmurtry: Okay makes sense! 👍

app/celery/scheduled_tasks.py
whabanks: Sounds good! As discussed, we can call this at the start of `check_job_status` since it runs on a schedule already and the functionality of `update_in_progress_jobs` really only needs to run when we do this check, and not periodically on its own.

smcmurtry (approved):
I found some bugs in nearby code while I was testing this and made some fixes here: https://github.com/cds-snc/notification-api/pull/1940

This code seems to be working as expected.

jimleroyer (approved):
:shipit:

PR #1937, opened 2023-07-19
Backend: Task to institute a Daily email limit

jzbahrai:
# Summary | Résumé

Previous to this code being merged in we would count sms limits + email limits in one "message limit"
After merging this code "message limit" on the Service object (in the DB) purely indicates an "email limit"

This Pr does a lot of things and should have been split into multiple PRs, but here is a changelog:

1. Added validators to check if the email_limit was breached or not
2. Updated notification-utils to use new redis keys as we will now cache email limits
3. Moved validation checks to post notification
4. Added tests for all the code changes.
5. Emails to use to send for emails nearing the limit/ over the limit

# Test:
Keep `FF_EMAIL_DAILY_LIMIT` = False

- [x] Update the message limit through settings - you get an email saying "the email limit" has been updated. 
Subject "Daily sending limit"
- [x] Send emails upto the limit. You should get two emails - subject "Action required: 80% of Daily sending limit" and "Action required: Daily sending limit reached for test"
- [x] Try sending 1 more email - you should be blocked an get an error

When `FF_EMAIL_DAILY_LIMIT`=True

- [x] Update the message limit through settings - you get an email saying "the email limit" has been updated. 
Subject "Daily email sending limit"
- [x] Send emails upto the limit. You should get two emails - subject "Action required: 80% of Daily email sending limit" and "Action required: Daily email sending limit reached for test.
- [x] You should get an error if you try and send 1 more email... It will say "technical error" as the frontend changes are not in yet. This is what needs to be changed: https://github.com/cds-snc/notification-admin/blob/main/app/main/views/send.py#L1053 and when I do it locally I get (I will put up this branch)
Screenshot 2023-07-21 at 10 18 09 AM

- [x] You will NOT get an error if you send an SMS as the message limit is not tied to the SMS limit


tests/app/notifications/test_validators.py
```diff
@@ -61,50 +65,55 @@ def enable_redis(notify_api):
...
         app.notifications.validators.redis_store.set.assert_not_called()
         assert not app.notifications.validators.services_dao.mock_calls
 
     @pytest.mark.parametrize(
-        "limit_type, key_type",
-        [("all", "test"), ("all", "team"), ("all", "normal"), ("sms", "test"), ("sms", "team"), ("sms", "normal")],
+        "limit_type",
+        ["email", "sms"],
     )
     def test_check_service_message_limit_in_cache_under_message_limit_passes(
```
whabanks, line 115: More of a suggestion for later consideration:
Test names could be a little more specific, we don't really have a naming convention in Notify for our tests but perhaps something like: `given_service_under_message_limit_and_cache_key_exists_when_checking_daily_limit_then_key_fetched_from_redis`

It's long as hell but when I read this I immediately have a good grasp of what is under test.

whabanks (approved):
LGTM, tested with provided instructions and it went as outlined!

tests/app/notifications/test_validators.py
jzbahrai: this is fair, we should standardize this going forward.

PR #1939, opened 2023-07-20
`mark-jobs-complete` should also run on jobs with the error status

smcmurtry:
# Summary | Résumé

For jobs that are not complete after 2 hours, the `job_status` gets updated to `error` here: https://github.com/cds-snc/notification-api/blob/main/app/celery/scheduled_tasks.py#L192.

Those jobs are then set to `process-incomplete-jobs` and the remaining rows are added to the `notifications` table. Once all the rows are added to the notifications table, these jobs can be marked `complete`.


whabanks (approved):
LGTM, nice catch on this one!

PR #1940, opened 2023-07-21
Fix 2 bugs in `process_incomplete_jobs` 

smcmurtry:
# Summary | Résumé

Bug 1: `process_incomplete_jobs` was changing the `job_status` from `error` back to `in progress`. In my testing, I observed the scheduled task `check_job_status` picking up the same job twice and sending it to `process_incomplete_jobs` twice, resulting in duplicate rows being added to the notification table. Removing the `job.job_status = JOB_STATUS_IN_PROGRESS` fixes that.

Bug 2: `process_incomplete_job` had a bug in the logic where it wasn't running on the correct rows. I simplified the logic here so that it just runs on the rows from `resume_from_row` until the end of the list.
 
1. `git checkout 8426c9e27f9713aa33a63cea90fd20c6a391d698` to get this branch with some extra code to make it easier to test this locally (chunk size = 10, always fail in `process_job` on the second chunk of rows, run `process_incomplete_jobs` on jobs that are up to 5 minutes old instead of 2 hours). 
2. Run the api and admin locally
3. open a postgres app like Postico and look at the jobs table
4. do a bulk send via the API of 15 emails to `success@simulator.amazonses.com`
5. refresh the jobs table, observe that the new job goes from `job_status` "in progress" to "error"
6. wait for a few minutes
7. observe that the `job_status` is "finished"
8. visit the job page in admin and observe that there are 15 notifications as expected

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -127,15 +127,11 @@ def process_job(job_id):
     csv = get_recipient_csv(job, template)
 
     rows = csv.get_rows()
-    i = 0
     for result in chunked(rows, Config.BATCH_INSERTION_CHUNK_SIZE):
-        if i > 0:
-            break
         process_rows(result, template, job, service)
         put_batch_saving_bulk_created(
             metrics_logger, 1, notification_type=db_template.template_type, priority=db_template.process_type
         )
-        i += 1
 
 
 def job_complete(job: Job, resumed=False, start=None):
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -1,6 +1,7 @@
 import json
 from collections import namedtuple
 from datetime import datetime
+from itertools import islice
 from typing import Any, Dict, List, Optional
 from uuid import UUID, uuid4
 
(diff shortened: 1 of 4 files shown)


app/celery/tasks.py
```diff
@@ -644,18 +644,11 @@ def process_incomplete_job(job_id):
...
-            first_row.append(row)
-            for result in chunked(rows, Config.BATCH_INSERTION_CHUNK_SIZE):
-                result = first_row + result
-                process_rows(result, template, job, job.service)
-                put_batch_saving_bulk_created(
-                    metrics_logger, 1, notification_type=db_template.template_type, priority=db_template.process_type
-                )
-                first_row = []
-            break
+    for result in chunked(islice(rows, resume_from_row, None), Config.BATCH_INSERTION_CHUNK_SIZE):
```
smcmurtry, line 647: `islice(rows, resume_from_row, None)` is just like `rows[resume_from_row:]` but for an iterator

whabanks (approved):
LGTM! Tested with the provided instructions. I was able to create a job, observed the first chunk of notifications (10) be added. Then the job was set to error and shortly after when `process_incomplete_job` is called, the remaining notifications (5) were added to the db and the job was marked as complete.

PR #1943, opened 2023-07-25
Safety PR: Revert marshmallow major upgrade

jimleroyer:
# Description

This is a safety PR to revert the cds-snc/notification-api#1844 in case there is 🔥 after a merge to production. Use this to quickly pull the plug.

We thoroughly tested these changes so most likely, we will not use this. We should close this PR after 1 week of the marshmallow changes deployed to production.


jimleroyer:
Closing as the world didn't catch fire. Revert isn't necessary.

PR #1947, opened 2023-07-27
Double number of concurrent celery workers (DO NOT RELEASE TO PROD)

jimleroyer:
# Summary | Résumé

Double the number of celery concurrent workers. The current value of `4` was probably selected for the number of CPU cores present on the machines running our workers. But as our workers are probably not CPU bound (we do not do numbers crunching) but most likely bound by the network, we can test with more celery concurrent workers, as these would wait on either some AWS service calls, the database or Redis.

### **This is change should not end in production, this is meant to test the configuration.**

Testing in a scratch environment might be preferable.


ben851 (approved):
LGTM

PR #1950, opened 2023-07-28
Remove FF_SPIKE_SMS_DAILY_LIMIT  and FF_SMS_PARTS_UI

jzbahrai:
# Summary | Résumé

Remove feature flags:
FF_SPIKE_SMS_DAILY_LIMIT 
FF_SMS_PARTS_UI


app/dao/fact_notification_status_dao.py
```diff
@@ -253,8 +253,6 @@ def fetch_notification_status_for_service_for_today_and_7_previous_days(service_
                     else_=FactNotificationStatus.billable_units,
                 ).label("count")
             ]
-            if current_app.config["FF_SMS_PARTS_UI"]
-            else [FactNotificationStatus.notification_count.label("count")]
```
andrewleith, line 257: We want to continue to show total messages here, not fragments, right? If so, I think we need to go the other way on this, and keep: `FactNotificationStatus.notification_count.label("count")` and remove the bit about `billable_units`.  This code is VERY unreadable (sorry - I wrote this 😭)

andrewleith (changes requested):
A few questions around desired behaviour where removing the `FF_SMS_PARTS_UI` flag.

andrewleith (approved):
LGTM!

PR #1952, opened 2023-07-31
use CELERY_CONCURRENCY

sastels:
# Summary | Résumé

Use the new `CELERY_CONCURRENCY` env var.

Note that if the variable is not defined, the concurrency will be the default value of 4.

1. run `make run-celery` without the env var. Note the concurrency is 4. Kill celery and see 4 workers stopped.
2. set `export CELERY_CONCURRENCY=8` and run `make run-celery` again. This time the concurrency is 8.

### DIFF
--- a/scripts/run_celery.sh
+++ b/scripts/run_celery.sh
@@ -2,6 +2,6 @@
 
 set -e
 
-echo "Start celery"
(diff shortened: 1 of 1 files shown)


ben851 (approved):
To the moon we go!

PR #1958, opened 2023-08-10
send sms using 3 new queues

sastels:
# Summary | Résumé

https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/186

Move sms sending (ie `deliver_sms()`) to their own queues (send-sms-high, send-sms-medium, send-sms-low)

note that these queues are still all services with the same celery works as the other queues.

- Send some sms
  - one-off, bulk
  - priority, normal, and bulk templates
- see what delivery queues are being used by looking for lines in the celery logs like "... sent to the .... queue for delivery"
- ensure the correct queue was used in each case

- We will need alarms added to terraform for these queues
- also need to add "age of oldest message" charts to the SMS dashboard

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -383,5 +383,5 @@ def beat_inbox_sms_priority():
 
     while list_of_sms_notifications:
         save_smss.apply_async((None, list_of_sms_notifications, receipt_id_sms), queue=QueueNames.PRIORITY_DATABASE)
-        current_app.logger.info(f"Batch saving with Bulk Priority: SMS receipt {receipt_id_sms} sent to in-flight.")
+        current_app.logger.info(f"Batch saving with Priority: SMS receipt {receipt_id_sms} sent to in-flight.")
         receipt_id_sms, list_of_sms_notifications = sms_priority.poll()
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -308,7 +308,7 @@ def save_smss(self, service_id: Optional[str], signed_notifications: List[Signed
         try:
             if not current_app.config["FF_EMAIL_DAILY_LIMIT"]:
                 check_service_over_daily_message_limit(notification_obj.key_type, service)
-            queue = notification_id_queue.get(notification_obj.id) or get_delivery_queue_for_template(template)  # type: ignore
+            queue = notification_id_queue.get(notification_obj.id) or get_delivery_queue_for_template(template)
             send_notification_to_queue(
                 notification_obj,
                 service.research_mode,
@@ -436,7 +436,7 @@ def try_to_send_notifications_to_queue(notification_id_queue, service, saved_not
         try:
             if not current_app.config["FF_EMAIL_DAILY_LIMIT"]:
                 check_service_over_daily_message_limit(notification_obj.key_type, service)
-            queue = notification_id_queue.get(notification_obj.id) or get_delivery_queue_for_template(template)  # type: ignore
+            queue = notification_id_queue.get(notification_obj.id) or get_delivery_queue_for_template(template)
             send_notification_to_queue(
                 notification_obj,
                 research_mode,
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -697,19 +697,19 @@ def choose_sending_queue(process_type: str, notif_type: str, notifications_count
     queue: Optional[str] = process_type
 
     if notifications_count >= large_csv_threshold:
-        queue = QueueNames.BULK
+        queue = QueueNames.SEND_SMS_LOW if notif_type == SMS_TYPE else QueueNames.BULK
     # If priority is slow/bulk, but lower than threshold, let's make it
     # faster by switching to normal queue.
     elif process_type == BULK:
-        queue = QueueNames.SEND_NORMAL_QUEUE.format(notif_type)
+        queue = QueueNames.SEND_SMS_MEDIUM if notif_type == SMS_TYPE else QueueNames.SEND_NORMAL_QUEUE.format(notif_type)
     else:
         # If the size isn't a concern, fall back to the template's process type.
         if process_type == PRIORITY:
-            queue = QueueNames.PRIORITY
+            queue = QueueNames.SEND_SMS_HIGH if notif_type == SMS_TYPE else QueueNames.PRIORITY
         elif process_type == BULK:
-            queue = QueueNames.BULK
+            queue = QueueNames.SEND_SMS_LOW if notif_type == SMS_TYPE else QueueNames.BULK
         else:
-            queue = QueueNames.SEND_NORMAL_QUEUE.format(notif_type)
+            queue = QueueNames.SEND_SMS_MEDIUM if notif_type == SMS_TYPE else QueueNames.SEND_NORMAL_QUEUE.format(notif_type)
     return queue
 
 
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -704,12 +704,7 @@ def choose_sending_queue(process_type: str, notif_type: str, notifications_count
         queue = QueueNames.DELIVERY_QUEUES[notif_type][Priorities.MEDIUM]
     else:
         # If the size isn't a concern, fall back to the template's process type.
-        if process_type == PRIORITY:
-            queue = QueueNames.DELIVERY_QUEUES[notif_type][Priorities.HIGH]
-        elif process_type == BULK:
-            queue = QueueNames.DELIVERY_QUEUES[notif_type][Priorities.LOW]
-        else:
-            queue = QueueNames.DELIVERY_QUEUES[notif_type][Priorities.MEDIUM]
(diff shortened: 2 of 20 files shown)


app/celery/scheduled_tasks.py
```diff
@@ -383,5 +383,5 @@ def beat_inbox_sms_priority():
 
     while list_of_sms_notifications:
         save_smss.apply_async((None, list_of_sms_notifications, receipt_id_sms), queue=QueueNames.PRIORITY_DATABASE)
-        current_app.logger.info(f"Batch saving with Bulk Priority: SMS receipt {receipt_id_sms} sent to in-flight.")
+        current_app.logger.info(f"Batch saving with Priority: SMS receipt {receipt_id_sms} sent to in-flight.")
```
jzbahrai, line 386: why is this priority? It can be either normal/ bulk or priority?

app/celery/tasks.py
```diff
@@ -697,19 +697,19 @@ def choose_sending_queue(process_type: str, notif_type: str, notifications_count
     queue: Optional[str] = process_type
 
     if notifications_count >= large_csv_threshold:
-        queue = QueueNames.BULK
+        queue = QueueNames.SEND_SMS_LOW if notif_type == SMS_TYPE else QueueNames.BULK
```
jzbahrai, line 700: can you also link the terraform pr that creates these queues? are we making them fifo as well?

app/config.py
```diff
@@ -52,8 +52,12 @@ class QueueNames(object):
     # pretty quickly.
     SEND_NORMAL_QUEUE = "send-{}-tasks"  # notification type to be filled in the queue name
 
-    # Queue for sending all SMS, except long dedicated numbers.
-    # TODO: Deprecate to favor priority queues instead, i.e. bulk, normal, priority.
+    # Queues for sending all SMS, except long dedicated numbers.
+    SEND_SMS_HIGH = "send-sms-high"
```
jzbahrai, line 56: I am not asking to change this - but asking a question:

why did we choose sms_high, sms_medium, sms_low instead of sms_priority, sms_bulk, sms_normal so it maps perfectly to our current system?

app/notifications/process_notifications.py
```diff
@@ -283,7 +283,7 @@ def send_notification_to_queue(notification, research_mode, queue=None):
         raise
 
     current_app.logger.info(
-        "{} {} sent to the {} queue for delivery".format(notification.notification_type, notification.id, queue)
+        "------ ---- {} {} sent to the {} queue for delivery".format(notification.notification_type, notification.id, queue)
```
jzbahrai, line 286: why did we need this? is this for checking the logs when this goes live?

app/celery/scheduled_tasks.py
jimleroyer: Reading quickly the code, it seems this function is for the SMS high priority notifications? Steve probably removed `Bulk` as this seems a bad copy paste of line 368.

app/celery/tasks.py
jimleroyer: Ideally BULK/LOW queues would not be FIFO if we want to treat high volume for these, but we could check the numbers to see if we'd want to make it. So far, only db saving ops for high priority notifications are FIFO if I am not mistaken.

app/config.py
jimleroyer: We want to refactor internally for some time the names because these are confusing. The current names at the moment are more aligned with the use case rather than their speed intent (bulk vs low priority, it's a use case vs speed name, or saying _bulk priority_ in itself when bulk has no priority, so saying low priority would be more sound, or saying _it is priority_ to mean high priority but we already have normal and bulk priority, etc..). Steve most likely named these in anticipation for the future refactoring.

What do you think about these? We can have a chat and sync on this.

app/celery/scheduled_tasks.py
sastels: yea, exactly, I think we just made beat_inbox_sms_priority() but copying the above beat_inbox_sms_bulk() but forgot the change the log.

I purposely didn't change the bulk / normal / priority in this PR since I figured I'd keep it simple and just add the queues, but 💯 we should change them everywhere to (IMO) low / medium / high

app/celery/tasks.py
sastels: currently the other queues used to send are not created in terraform but when they are used by api / celery, so I just copied that approach.

app/config.py
sastels: yeah, I was going to just copy what we did before (as I did for much of how the new code is written) but really really hate how we do it (basically what jimmy said), so thought I'd just go with high / med / low and spark the conversation 😁

app/notifications/process_notifications.py
sastels: yeah, this log is just to ensure that we can 💯 see how a message was sent, but we definitely don't need all the dashes, that was just to help me see the lines locally. 😅  I shall get rid of them.

app/config.py
jzbahrai: spoke offline. this looks good.

app/celery/tasks.py
```diff
@@ -697,19 +697,19 @@ def choose_sending_queue(process_type: str, notif_type: str, notifications_count
...
         if process_type == PRIORITY:
-            queue = QueueNames.PRIORITY
+            queue = QueueNames.SEND_SMS_HIGH if notif_type == SMS_TYPE else QueueNames.PRIORITY
         elif process_type == BULK:
-            queue = QueueNames.BULK
+            queue = QueueNames.SEND_SMS_LOW if notif_type == SMS_TYPE else QueueNames.BULK
         else:
-            queue = QueueNames.SEND_NORMAL_QUEUE.format(notif_type)
+            queue = QueueNames.SEND_SMS_MEDIUM if notif_type == SMS_TYPE else QueueNames.SEND_NORMAL_QUEUE.format(notif_type)
     return queue
```
jimleroyer, line 713: This is getting more complex and would be worth having its own data structure (like a dictionary) or class to express such logic.

tests/app/test_config.py
```diff
@@ -27,7 +27,7 @@ def reload_config():
 def test_queue_names_all_queues_correct():
     # Need to ensure that all_queues() only returns queue names used in API
     queues = QueueNames.all_queues()
-    assert len(queues) == 17
+    assert len(queues) == 20
```
jimleroyer, line 30: Hmm interesting we have a test for the number of queues..

jimleroyer (approved):
LGTM. I provided a potential improvement though, but this could be merged.

app/models.py
```diff
@@ -1084,11 +1085,7 @@ def process_type(cls):
         )
 
     def queue_to_use(self):
-        return {
-            NORMAL: QueueNames.NORMAL,
-            PRIORITY: QueueNames.PRIORITY,
-            BULK: QueueNames.BULK,
-        }[self.process_type]
```
sastels, line 1091: Essentially, we move this logic to app.utils to avoid circular import errors 😞

app/celery/tasks.py
sastels: yeah... I spent far too much time refactoring a bit and adding a `Priorities` class and a dict `DELIVERY_QUEUES` to the `QueueNames` class. I think this makes it easier to understand what queues are going to be used.

app/config.py
```diff
@@ -93,6 +114,24 @@ class QueueNames(object):
...
+            Priorities.LOW: SEND_SMS_LOW,
+            Priorities.MEDIUM: SEND_SMS_MEDIUM,
+            Priorities.HIGH: SEND_SMS_HIGH,
+        },
+        "email": {
+            Priorities.LOW: BULK,
+            Priorities.MEDIUM: SEND_EMAIL,
+            Priorities.HIGH: PRIORITY,
+        },
+        "letter": {
```
jzbahrai, line 128: can we remove this ?

app/config.py
```diff
@@ -10,13 +10,34 @@
...
+class Priorities(object):
+    LOW = "low"
+    MEDIUM = "medium"
+    HIGH = "high"
+    BULK = "bulk"
+    NORMAL = "normal"
+    PRIORITY = "priority"
+
+    @staticmethod
+    def to_lmh(priority: str) -> str:
```
jimleroyer, line 30: Let's add some documentation to say what it converts from and to please. 🙏

app/config.py
```diff
@@ -89,6 +114,24 @@ class QueueNames(object):
     # Queue for delivery receipts such as emails sent through AWS SES.
     DELIVERY_RECEIPTS = "delivery-receipts"
 
+    DELIVERY_QUEUES = {
```
jimleroyer, line 117: Thank youuu!

app/config.py
jimleroyer: I wonder if this is meant for tests maybe? Else I'd vote to not have it too.

app/models.py
jimleroyer: Makes sense!

app/config.py
sastels: 😞 yeah, exactly, a ton of tests fail without this.

PR #1965, opened 2023-08-28
Refactor dao_get_template_by_id to no longer return a tuple

whabanks:
# Summary | Résumé
This PR refactors `dao_get_template_by_id` so that it returns either a `Template` or `TemplateHistory` object rather than a tuple.

- Added additional fallbacks for fetching the `reply_to_text` to `save_smss` and `save_emails` when either then template or service is fetched from the cache and relationship data is not populated.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -247,16 +247,10 @@ def save_smss(self, service_id: Optional[str], signed_notifications: List[Signed
         reply_to_text = ""  # type: ignore
         if sender_id:
             reply_to_text = dao_get_service_sms_senders_by_id(service_id, sender_id).sms_sender
-            if isinstance(template, tuple):
-                template = template[0]
-        # if the template is obtained from cache a tuple will be returned where
-        # the first element is the Template object and the second the template cache data
-        # in the form of a dict
-        elif isinstance(template, tuple):
-            reply_to_text = template[1].get("reply_to_text")  # type: ignore
-            template = template[0]
+        elif template.service:
+            reply_to_text = template.get_reply_to_text()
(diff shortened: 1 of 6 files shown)


jzbahrai (approved):

app/celery/tasks.py
```diff
@@ -247,16 +247,10 @@ def save_smss(self, service_id: Optional[str], signed_notifications: List[Signed
...
-        # if the template is obtained from cache a tuple will be returned where
-        # the first element is the Template object and the second the template cache data
-        # in the form of a dict
-        elif isinstance(template, tuple):
-            reply_to_text = template[1].get("reply_to_text")  # type: ignore
-            template = template[0]
+        elif template.service:
+            reply_to_text = template.get_reply_to_text()
         else:
-            reply_to_text = template.get_reply_to_text()  # type: ignore
```
sastels, line 259: This PR brings me joy. One question: why change this line?

whabanks: Glad to bring you some joy! I changed this up because when the template is pulled from the cache it's associated service object is not populated. Since `get_reply_to_text()` just calls `service.get_default_sms_sender()` under the hood I changed the logic to:
1. Try and fetch the reply to text from the DB if a sender_id is present
2. If the template's service object is set, i.e. we didn't fetch it from the cache, then call `get_reply_to_text()`
3. Otherwise use the service object we fetched earlier to `get_default_sms_sender()`

sastels: NIce!

sastels (approved):
Awesome!! 🎉

PR #1966, opened 2023-08-30
Auto assign the notify organisation when a service is first created

smcmurtry:
# Summary | Résumé

Auto assign the notify organisation when a service is first created.
This required some changes to the gc-organisations repo to add the Notify organisation ID to the dataset: https://github.com/cds-snc/gc-organisations/pull/47

I also added some logic to set the 3 day data retention if the service belongs to a Prov/Territory.

# Testing

To test this locally you will need to do some set up.

## Set up:
1. check the branch out locally
2. Add the 2 new environment variables to your .env file: `CRM_GITHUB_PERSONAL_ACCESS_TOKEN` and `CRM_ORG_LIST_URL`. You can find the values in the notification-manifest repo
3. run the api and admin locally
4. log in as a platform admin and create an organisation called "Government of Ontario - Local Test"
5. open your local db via the Postico App and find the new record in the `organisation` table
6. edit and save a new `id` value for your new organisation: `c4be8612-85c3-454b-ace1-309352220cf5`. This is so it matches up the organisation ID in prod
7. visit the page http://localhost:6012/organisations/c4be8612-85c3-454b-ace1-309352220cf5 to verify that your new id has been applied correctly

## Test 1:
1. Create a new service and select "Government of Ontario" from the dropdown. 
2. Enter "test" in the box for "Enter any other names for your group (Optional)"
3. Create the service
4. go to the service settings page and verify that the organisation has been set correctly as "Government of Ontario - Local Test"

## Test 2:
Same steps as Test 1, but go through the service creation flow in french

## Test 3:
Same steps as Test 1 but select "Government of British Columbia". No organisation should be set automatically since there is no local organisation with the same ID.


whabanks (changes requested):
I tested this locally with `CRM_GITHUB_PERSONAL_ACCESS_TOKEN` and `CRM_ORG_LIST_URL` pulled from manifests, and the org id edited to match prod. 
On step 3 when saving the org I am getting a 500. In `get_organisation_id_from_crm_org_notes`, it looks like the dict containing the org data does not have the key `notify_organisation_id`:
```
Traceback (most recent call last):
  File "/home/vscode/.cache/pypoetry/virtualenvs/notification-api-xS3fZVNL-py3.10/lib/python3.10/site-packages/flask/app.py", line 1484, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/vscode/.cache/pypoetry/virtualenvs/notification-api-xS3fZVNL-py3.10/lib/python3.10/site-packages/flask/app.py", line 1469, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/workspace/app/service/rest.py", line 262, in create_service
    get_organisation_id_from_crm_org_notes(data["organisation_notes"])
  File "/workspace/app/service/utils.py", line 96, in get_organisation_id_from_crm_org_notes
    en_dict[item["name_eng"]] = item["notify_organisation_id"]
KeyError: 'notify_organisation_id'
```

app/dao/services_dao.py
```diff
@@ -284,7 +289,10 @@ def dao_create_service(
     if service_permissions is None:
         service_permissions = DEFAULT_SERVICE_PERMISSIONS
 
-    organisation = dao_get_organisation_by_email_address(user.email_address)
+    if organisation_id:
+        organisation = dao_get_organisation_by_id(organisation_id)
```
whabanks, line 293: I suspect we wouldn't come across this case in production, but is there a way we could fail more gracefully if the `org_id` we fetch from the CRM org notes doesn't exist in the DB? Currently we just get a "technical difficulties" screen in this case.
Maybe we can still create the service and not add the org? Adding some logging here or in `service/rest.py` could be helpful as well.

whabanks (changes requested):

app/dao/services_dao.py
smcmurtry: Great catch Will! I thought I had tested that case but I am seeing an error when the organisation id is missing from the DB. I have pushed a fix - could you re-test?

whabanks: Re-tested and it looks great! The service gets created without issue when the org doesn't exist yet in the DB. ✅

whabanks (approved):
Looks great, nice work!

PR #1925, opened 2023-07-10
Add celery worker make targets

sastels:
# Summary | Résumé

Add some make targets for running the celery scripts to make local running a little easier.

# Test instructions | Instructions pour tester la modification

Run the celery workers with `make run-celery-beat` and either `make run-celery` or `make run-celery-clean`

PR #1972, opened 2023-09-06
tweak celery scripts

sastels:
# Summary | Résumé

As part of the process of moving sms sending to its own pod deployment, we need a new script for those pods to run!
Our code is all over the place, so this is the process:

- created new celery queues for sms sending (and associated alarms) **DONE**
- create new scripts for 
  - the new SMS celery pod deployment (**this PR**)
  - the old celery pod deployment (ie not containing the send-sms-* queues) (**this PR**)
- create the new deployment that uses this script in staging ([manifest repo PR](https://github.com/cds-snc/notification-manifests/pull/1980))

I've also added some documentation to the scripts, as well as a new script for running all the queues locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,17 +50,13 @@ smoke-test:
 run: ## Run the web app
 	flask run -p 6011 --host=0.0.0.0
 
-.PHONY: run-celery
-run-celery-local: ## Run the celery workers
+.PHONY: run-celery-local
+run-celery-local: ## Run the celery workers with all the queues
 	./scripts/run_celery_local.sh
 
-.PHONY: run-celery-clean
-run-celery-local-clean: ## Run the celery workers but filter out common scheduled tasks
+.PHONY: run-celery-local-filtered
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -51,12 +51,12 @@ run: ## Run the web app
 	flask run -p 6011 --host=0.0.0.0
 
 .PHONY: run-celery
-run-celery: ## Run the celery workers
-	./scripts/run_celery.sh
+run-celery-local: ## Run the celery workers
+	./scripts/run_celery_local.sh
```
jimleroyer, line 55: We can keep the run-celery make target? we'd have to change/duplicate the `.PHONY` name too I think.

sastels: These make targets were added in #1925 to make it clearer how to run locally, so I think we can just move the make targets to the new local scripts. Need tup update the `.PHONY`s though 👍

jimleroyer (approved):

PR #1973, opened 2023-09-08
Add an internal api so that we might revoke api keys

jzbahrai:
# Summary | Résumé

Add an internal API so that SRE can call it and revoke API keys

- [x] Added a new column in api_keys and api_keys history to save the information that SRE sends up
- [x] Added a function to save the extra information into the api_keys table
- [x] Added the api that does the actual revokation of an exposed token
- [x] Added a function to send an email implying a key has been revoked - but this is not tested - we will wait for content to test this function.
- [x] Tests for the above

### DIFF
--- a/app/api_key/rest.py
+++ b/app/api_key/rest.py
@@ -1,8 +1,9 @@
+from datetime import datetime
+
 import werkzeug
 from flask import Blueprint, current_app, jsonify, request
 
 from app import DATETIME_FORMAT
-from app.config import QueueNames
 from app.dao.api_key_dao import (
     expire_api_key,
     get_api_key_by_secret,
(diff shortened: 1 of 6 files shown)


whabanks (approved):
Code looks good, dev container builds & DB migrations are working. LGTM!

app/api_key/rest.py
```diff
@@ -59,3 +67,94 @@ def get_api_keys_ranked(n_days_back):
...
+        data = request.get_json()
+    except werkzeug.exceptions.BadRequest as errors:
+        raise InvalidRequest(errors, status_code=400)
+
+    # Step 1
+    for api_key_data in data:
+        try:
+            # take last 36 chars of string so that it works even if the full key is provided.
+            api_key_token = api_key_data["token"][-36:]
+            api_key = get_api_key_by_secret(api_key_token)
```
andrewleith, line 135: I expected this method to receive and validate a shared secret but instead it uses the API key itself for auth?

It seems weird to use the thing that was exposed publicly, but I suppose if the hacker is using the leaked key to expire it that isn't really a problem...

andrewleith (commented):
Looks good, just a clarification on the deviation from the authentication mechanism we discussed initially

app/api_key/rest.py
jzbahrai: I am missing something... I assumed that the user posted "gc-notify-dkfjsdkjfksdjfksd" which is the actual key on the open internet. We would need to find the key in the DB by using get_api_key_secret? 

Let me know what you are thinking

andrewleith: Who is calling this endpoint? SRE?  I thought we discussed them providing a secret/token during this request to authenticate the call?

PR #1976, opened 2023-09-12
Update daily notification stats to use 12am utc 

andrewleith:
# Summary | Résumé

This PR fixes the daily stats query to use midnight UTC which aligns with the way we handle daily limits.

### DIFF
--- a/app/dao/fact_notification_status_dao.py
+++ b/app/dao/fact_notification_status_dao.py
@@ -1,6 +1,6 @@
-import pytz 
 from datetime import datetime, time, timedelta
 
+import pytz
 from flask import current_app
 from notifications_utils.timezones import convert_local_timezone_to_utc
 from sqlalchemy import Date, case, func
--- a/app/dao/fact_notification_status_dao.py
+++ b/app/dao/fact_notification_status_dao.py
@@ -1,6 +1,5 @@
 from datetime import datetime, time, timedelta
 
-import pytz
 from flask import current_app
 from notifications_utils.timezones import convert_local_timezone_to_utc
 from sqlalchemy import Date, case, func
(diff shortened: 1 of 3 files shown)


app/dao/fact_notification_status_dao.py
```diff
@@ -239,7 +241,11 @@ def fetch_notification_status_for_service_for_day(bst_day, service_id):
 
 
 def fetch_notification_status_for_service_for_today_and_7_previous_days(service_id, by_template=False, limit_days=7):
-    start_date = midnight_n_days_ago(limit_days)
+    if limit_days == 1:
+        start_date = get_midnight(datetime.now(tz=pytz.utc))
+    else:
+        start_date = midnight_n_days_ago(limit_days)
```
smcmurtry, line 247: I think this will return the correct datetime for the case we are trying to fix, but I think there are a couple issues with this change:
- `start_date` is a timezone aware datetime if `limit_days = 1`, and it is timezone naive if `limit_days > 1`. That makes the code harder to understand
- the way `start_date` is calculated is very different if `limit_days = 1` vs if `limit_days > 1`. This might cause issues in the future when we forget this detail and we are calling the endpoint via notification-admin

Could we just always use utc midnight for this function? I'm thinking something like:
```
def utc_midnight_n_days_ago(number_of_days):
    """
    Returns utc midnight a number of days ago.
    """
    return get_midnight(datetime.utcnow() - timedelta(days=number_of_days))
```
in `app/utils.py`, and then:
```
start_date = utc_midnight_n_days_ago(limit_days)
```
in this file.

andrewleith: Yeah for sure, I specifically targeted the 1 day case, but you make good points here.  Let's chat!

PR #1979, opened 2023-09-13
use config variable for deliver_sms rate limiting

sastels:
# Summary | Résumé

https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/80

make the celery worker rate limit for `deliver_sms` a config variable so we can keep prod as it is while we play with staging.

- make sure sms still send
- run another `INTERNAL_TEST_NUMBER` perf test in staging and verify that using the default we still get approx 3 sms / sec


ben851 (approved):

jimleroyer (approved):
I like this change. 👍

PR #1980, opened 2023-09-13
Handle the `UnsupportedMediaType` error

smcmurtry:
# Summary | Résumé

This PR adds error handling so that we don't cause a 500 error when and API caller doesn't set the `Content-Type` header to `application/json`. You can reproduce the error on staging with the following command:
```
curl -i -X POST \
   -H "Authorization:ApiKey-v1  gcntfy-talend_key-3********" \
   -H "Content-Type:" \
   -d \
'{
  "email_address": "stephen.astels@cds-snc.ca",
  "template_id": "eb3a6c36-2e4d-4e55-9e25-ed4b314ff3e9",
  "personalisation": {
        "var": "test /email"
    }
}' \
 'https://api.staging.notification.cdssandbox.xyz/v2/notifications/email'
```

Check out this branch locally and try the above curl command. You should get a 415 error back now instead of a 500 error.

### DIFF
--- a/app/v2/notifications/post_notifications.py
+++ b/app/v2/notifications/post_notifications.py
@@ -251,6 +251,11 @@ def post_notification(notification_type: NotificationType):
             message="Error decoding arguments: {}".format(e.description),
             status_code=400,
         )
+    except werkzeug.exceptions.UnsupportedMediaType as e:
+        raise BadRequestError(
+            message="UnsupportedMediaType error: {}".format(e.description),
+            status_code=415,
+        )
 
     if notification_type == EMAIL_TYPE:
         form = validate(request_json, post_email_request)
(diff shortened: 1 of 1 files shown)


app/v2/notifications/post_notifications.py
```diff
@@ -161,6 +161,11 @@ def post_bulk():
         request_json = request.get_json()
     except werkzeug.exceptions.BadRequest as e:
         raise BadRequestError(message=f"Error decoding arguments: {e.description}", status_code=400)
+    except werkzeug.exceptions.UnsupportedMediaType as e:
+        raise BadRequestError(
+            message="UnsupportedMediaType error: {}".format(e.description),
+            status_code=415,
+        )
```
sastels, line 168: I had thought we should add some details here, but TBH it's probably better to just pass the error on appropriately as you do here until we hear otherwise from our users.

sastels: actually in my local test I get the error message
```
"UnsupportedMediaType error: Did not attempt to load JSON data because the request Content-Type was not 'application/json'."
```
which does totally explain the problem! 👍

sastels (approved):
tested locally!

PR #1984, opened 2023-09-20
increase concurrency on staging sms celery pods

sastels:
# Summary | Résumé

Bump up the concurrency on the sms-send celery pods from 4 to 6 to try to get them to do more work.

Note that these pods / script are currently only in staging.

### DIFF
--- a/scripts/run_celery_send_sms.sh
+++ b/scripts/run_celery_send_sms.sh
@@ -4,7 +4,7 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


ben851 (approved):
As mentioned in slack, we should eventually split the CELERY_CONCURRENCY param value between sms and emails, but this is good for the test for now

sastels:
or tbh maybe we just decide on good concurrency numbers and get get rid of the param 🤷

PR #2001, opened 2023-10-11
Fix script to send to std err

jzbahrai:
# Summary | Résumé

Update the celery exit script to send to stderr


ben851 (approved):

sastels:
Should we also change the  "Run script" echo in line 70?

PR #2002, opened 2023-10-12
Change the output of echo to /proc/1/fd/1

jzbahrai:
# Summary | Résumé

Instead of writing to stdout we will try the above. This is because of the following assumption:
1. celery runs on one thread
2. this script runs on another thread

the script doesnt write to the same stdout that gets picked by the cloudwatch agent. (again this is a working theory)
by making the above change, we should be writing to the same output which will then translate to logs on cloudwatch

To be tested on staging 🤞🏽


ben851 (approved):
LGTM

PR #2011, opened 2023-10-31
SMS soak test

sastels:
# Summary | Résumé

Add flag to soak test SMS rather than emails.

- Add a `SMS_TEMPLATE_ID=` to your `scripts/soak_test/.env` file. This template should have no variables.
- run `locust -f ./soak_test_send_notification.py --sms`

### DIFF
--- a/scripts/load_test/README.md
+++ b/scripts/load_test/README.md
@@ -1,4 +1,4 @@
-# Soak test
+# Load test
 
 ## Goals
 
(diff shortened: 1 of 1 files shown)


ben851 (approved):

scripts/soak_test/soak_test_send_notification.py
```diff
@@ -0,0 +1,38 @@
+import os
```
jimleroyer, line 1: We didn't want this to sit in a subfolder of tests-stress?

sastels: 🤔 We could think of this as a very low level stress test... generally I've been using high priority templates for these soak tests and low priority ones for the larger stress tests so it would require either rewriting the code a bit or changing the .env file when switching between them. 🤷

PR #2013, opened 2023-11-02
Adding cwagent wait script to all celery launchers

ben851:
# Summary | Résumé

There were a few celery run scripts that weren't checking that cwagent was ready before launching.

Deploy to staging
Ensure the waiting for cwagent line is in celery logs
Smoke test staging


sastels (approved):
good catch

PR #2014, opened 2023-11-06
add local celery beat script

sastels:
# Summary | Résumé

add a local celery-beat script that doesn't contain the CWAgent stuff (this causes an error when running locally)

```
make run-celery-beat-local
```


jimleroyer (approved):
LGTM

PR #2022, opened 2023-11-15
Fix Shellcheck errors by doing as little work as possible

CalvinRodo:
# Summary | Résumé

These fixes a bunch of shellcheck errors and ignores a few I didn't test, actually I didn't test any of this, most of the fixes were pretty straightforward ones, there was one I fixed with co-pilot that should be tested and the others I didn't fix I just assumed they worked as is since they were in the existing codebase.

- **Fixes**: https://www.shellcheck.net/wiki/SC1091 Couldn't find the file that was being referenced (`environment_test.sh`) in `scripts/run_single_test.sh` so I just the source to /dev/null
- **Fixes**: https://www.shellcheck.net/wiki/SC2028 Replaces some echo's where a control character was passed "\n" with a printf so it should work the same and not complain
- **Fixes the read -p error**: https://www.shellcheck.net/wiki/SC3045 by asking co-pilot to fix it, I can't guarantee this one will work, **YOU SHOULD TEST THIS FILE: scripts/run_celery_purge.sh**
- **Ignores**: https://www.shellcheck.net/wiki/SC2009 I'm assuming it worked and I didn't want to find the pgrep way of doing it, also it's not posix
- **Ignores**: https://www.shellcheck.net/wiki/SC2219 I'm assuming it worked and I didn't want to test if the following works
```bash 
((wait_time++)) || true
```
- **Fixes**: https://www.shellcheck.net/wiki/SC2105 By removing the break from the case, they don't need it.
- **Fixes**: https://www.shellcheck.net/wiki/SC3046 By replacing `source` with `.`
- **Fixes**: https://www.shellcheck.net/wiki/SC2068 by wrapping array Expansion with double quotes
- **Fixes**: https://www.shellcheck.net/wiki/SC2086 By wrapping the variable in double quotes

### DIFF
--- a/scripts/run_celery.sh
+++ b/scripts/run_celery.sh
@@ -27,4 +27,4 @@ fi
 
 echo "Start celery, concurrency: ${CELERY_CONCURRENCY-4}"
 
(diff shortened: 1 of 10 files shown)


jimleroyer (approved):
🙇 🖖

PR #2024, opened 2023-11-16
Count by sms messages not sms parts for daily limit

whabanks:
# Summary | Résumé
This PR changes the way we count SMS messages toward a services daily limit, we now count in a 1:1 manner instead of counting fragments toward their daily limit, provided the message does not exceed 4 fragments. 1 SMS message = 1 count toward the daily limit.

Previously we counted individual fragments toward a service's daily limit in Redis. This resulted in a situation where, when a service with 50 daily SMS messages remaining, sends a bulk job of 25 SMS messages with a template size totally 3 SMS parts was counting 75 messages counted toward their daily limit, preventing them from performing the send.

Related work: https://github.com/cds-snc/notification-utils/pull/256

# Testing locally
Checkout the branch: [align-csv-sms-counting-locally-test](https://github.com/cds-snc/notification-api/tree/align-csv-sms-counting-locally-test)

1. Send a one-off SMS message with a template totaling 2 or more fragments
2. Check the redis key: `sms-<service_id>-{date_today}-count` and make sure it was only incremented by 1 even though there were 2 SMS parts

## Multi-Fragment sms sends do not prevent sending
With the aforementioned scenario in mind:
1. Perform a bulk send of SMS messages with multiple sms parts that, previously, would go over the service's current daily sms limit
2. Check the Redis key `sms-<service_id>-{date_today}-count` to ensure that the daily count was incremented by the size of the job rather than the fragments in each message.

**Another example for clarity**:
In this case the Redis key should increment by 167 **_not_** 501 like it would previously if we were counting by fragment.
- daily sms limit - 500
- SMS template  - 3 fragments
- Bulk send size - 167

### DIFF
--- a/app/sms_fragment_utils.py
+++ b/app/sms_fragment_utils.py
@@ -13,11 +13,11 @@ def fetch_todays_requested_sms_count(service_id: UUID) -> int:
         return fetch_todays_total_sms_count(service_id)
 
     cache_key = sms_daily_count_cache_key(service_id)
-    fragment_count = redis_store.get(cache_key)
-    if fragment_count is None:
-        fragment_count = fetch_todays_total_sms_count(service_id)
-        redis_store.set(cache_key, fragment_count, ex=int(timedelta(hours=2).total_seconds()))
-    return int(fragment_count)
+    sms_count = redis_store.get(cache_key)
+    if sms_count is None:
+        sms_count = fetch_todays_total_sms_count(service_id)
+        redis_store.set(cache_key, sms_count, ex=int(timedelta(hours=2).total_seconds()))
+    return int(sms_count)
 
 
 def increment_todays_requested_sms_count(service_id: UUID, increment_by: int):
--- a/tests/app/v2/notifications/test_post_notifications.py
+++ b/tests/app/v2/notifications/test_post_notifications.py
@@ -1797,12 +1797,13 @@ class TestEmailsAndLimitsForSMSFragments:
             "template_id": str(template.id),
             "personalisation": {" Name": "Jo"},
         }
-        for x in range(5):
+        for x in range(7):
             create_sample_notification(notify_db, notify_db_session, service=service)
 
         __send_sms()  # send 1 sms (3 fragments) should be at 80%
         assert send_warning_email.called
 
+        __send_sms()
         response = __send_sms()  # send one more, puts us at 11/10 fragments
         assert response.status_code == 429  # Ensure send is blocked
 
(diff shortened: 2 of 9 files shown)


tests/app/v2/notifications/test_post_notifications.py
```diff
@@ -1895,48 +1896,6 @@ def __send_sms(number_to_send=1):
         response = __send_sms(1)  # attempt to send over limit (11 with max 10)1
         assert response.status_code == 400  # Ensure send is blocked - not sure why we send a 400 here and a 429 everywhere else
 
-    def test_API_BULK_cant_hop_over_limit_2_fragment(self, notify_api, client, notify_db, notify_db_session, mocker):
```
whabanks, line 1898: There are several other tests involving fragment hopping. Considering the [move away from / removal of sms parts](https://gcdigital.slack.com/archives/CV38DBNVA/p1698960023450489) for counting do we need to keep these tests?

app/sms_fragment_utils.py
```diff
@@ -13,11 +13,11 @@ def fetch_todays_requested_sms_count(service_id: UUID) -> int:
...
 
     cache_key = sms_daily_count_cache_key(service_id)
-    fragment_count = redis_store.get(cache_key)
-    if fragment_count is None:
-        fragment_count = fetch_todays_total_sms_count(service_id)
-        redis_store.set(cache_key, fragment_count, ex=int(timedelta(hours=2).total_seconds()))
-    return int(fragment_count)
+    sms_count = redis_store.get(cache_key)
+    if sms_count is None:
+        sms_count = fetch_todays_total_sms_count(service_id)
```
andrewleith, line 18: I think you need to update the code in `fetch_todays_total_sms_count()`.. it is still returning sms parts isn't it?

andrewleith (commented):
Should [this line change ](https://github.com/cds-snc/notification-api/pull/2024/files#diff-cc69ccd3b3b0ba4e44f73c182412190d6af4e52c1bc0680fdc745a60c2eced22R300)too?

app/sms_fragment_utils.py
whabanks: That line does need to be changed. Nice catch on `fetch_todays_total_sms_count()` I'll update both places.

tests/app/v2/notifications/test_post_notifications.py
andrewleith: Yeah agreed we won't need these hop over tests anymore since it won't work the same way.  For the tests that hop over by 1 though, are they still relevant? Or maybe that functionality is covered elsewhere?

whabanks: I considered keeping the hop by 1 tests. I chose not to because it seemed like we were testing the scenario where a service had 1 fragment left and tried to send a message that had 2 fragments in it. Since it is a 1:1 count now, we won't run into that situation anymore. I think [test_post_sms_not_enough_messages_left](https://github.com/cds-snc/notification-api/blob/f6747d29b52c42a0f90b49ec7a3caf2623dc16fe/tests/app/v2/notifications/test_post_notifications.py#L1474) is adequate.

This does raise an area that I did not test thoroughly though - Where a service has 1 message left and they want to send a job with 2 or more. I will test this more and add tests if necessary but we should be covered.

whabanks: I did some boundary tests trying to go over the SMS limit with jobs. I wasn't able to break anything locally.

andrewleith (approved):
:shipit:

PR #2025, opened 2023-11-16
10m -> 20m perf test

sastels:
# Summary | Résumé

perf test running longer, but still hasn't gotten to levelling off.
image

check perf test after it runs


andrewleith (approved):
:shipit:

PR #2034, opened 2023-11-22
just test send rates

sastels:
# Summary | Résumé

Update README with new instructions for running celery locally.


sastels:
this... wasn't what I was trying to do...

PR #2035, opened 2023-11-22
Update local celery script

sastels:
# Summary | Résumé

~Update README with local celery info~
Locally, have the celery worker also run beat. Update the documentation accordingly.

- run the api
- run all celery with`make run-celery-local`
- see the beat logs coming through
- verify that sending still works


andrewleith (approved):
LGTM 🎉

README.md
```diff
@@ -109,13 +109,9 @@ file. Copy that file to `.env` and customize it to your needs.
...
 ##  To run the queues
-```
-scripts/run_celery_local.sh
-```
+`make run-celery-local`
 
-```
-scripts/run_celery_beat.sh
-```
+`make run-celery-beat-local`
```
jimleroyer, line 114: It's great to mention the make target instead. As these are one more level of abstraction, we could even just have a one target named `make run-celeries-local`, as it doesn't have to be a one to one mapping with the actual scripts but rather intended usage (run celery setup locally).

jimleroyer (approved):
LGTM, merci!

I left a comment on potential improvement (on top of the improvement in this PR) but it's outside the scope of this change.

README.md
sastels: hrm, running both the beat and celery workers with one make command? I didn't think that would work, but admittedly didn't try! I shall see!

sastels: oh wow, we can just add "--beat" to the worker script. That was easy.

sastels:
Clearing approvals since I changed the scripts.

jimleroyer (approved):

PR #2038, opened 2023-11-24
Smoke Test Loop Script

ben851:
# Summary | Résumé

Here's a quick script that will run the smoke tests in a loop for a given number of times (or indefinitely)

make smoke-loop -m 2

### DIFF
--- a/tests_smoke/smoke_loop.py
+++ b/tests_smoke/smoke_loop.py
@@ -1,0 +1,100 @@
+import subprocess
+import os
+import time
+from argparse import ArgumentParser
+from smoke.common import Attachment_type, Config, Notification_type  # type: ignore
+from smoke.test_admin_csv import test_admin_csv  # type: ignore
+from smoke.test_admin_one_off import test_admin_one_off  # type: ignore
+from smoke.test_api_bulk import test_api_bulk  # type: ignore
+from smoke.test_api_one_off import test_api_one_off  # type: ignore
+
+runCount = 0
+maxRuns = 0
+
+class bcolors:
+    HEADER = '\033[95m'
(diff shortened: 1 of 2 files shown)


tests_smoke/smoke_loop.py
```diff
@@ -0,0 +1,97 @@
...
+        print(f"Sourcing environment variables from external shell script {args.source}")
+        subprocess.run(["source", args.source], executable='/bin/bash')
+
+    print(bcolors.ENDC)
+
+
+init()
+
+
+while True:
```
jzbahrai, line 65: I would put both the init() and this part under the main function, that would be the pythonic way to call this:

```
def main():
    print("Hello World!")

if __name__ == "__main__":
    main()
```

tests_smoke/smoke_loop.py
```diff
@@ -0,0 +1,97 @@
...
+
+    result = None
+    seconds = None
+    startTime = None
```
jzbahrai, line 69: you don't need to initiatize both startTime and endTime as you are reinitializing underneath

tests_smoke/smoke_loop.py
```diff
@@ -0,0 +1,97 @@
...
+    BOLD = '\033[1m'
+    UNDERLINE = '\033[4m'
+
+
+def init():
+
+    global maxRuns
+
+    parser = ArgumentParser()
+    parser.add_argument("-m", "--max-runs", dest="maxruns",
```
jzbahrai, line 34: you could add this documentation in either the Makefile and on top of this script so folks have an example of how to run it

jzbahrai (approved):

PR #2056, opened 2023-12-06
Delete cache key so we resend warning email

jzbahrai:
# Summary | Résumé

When we changed email limits, we were not resetting the cache key that indicated an email to warn the end user had been sent. This PR does that 
Screenshot 2023-12-05 at 11 23 05 AM
Screenshot 2023-12-05 at 11 22 57 AM


andrewleith (commented):
Looks good.  Is there a card related to this? 

Should we add a test that does something like:
- Puts a service near and then at their daily limit
- Asserts email is sent 
- Updates a service's limit
- Puts the service near and then at their daily limit
- Asserts email is sent again

jzbahrai:
@andrewleith I edited an existing test.. I dont think we need to check the warning email is sending when we are at the limit.. that is checked in a different part of the codebase (and we have tests for that). We want to check that the cache key is reset (the test I updated). Is that okay?

jzbahrai:
The PR is here: https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/991

andrewleith:
> @andrewleith I edited an existing test.. I dont think we need to check the warning email is sending when we are at the limit.. that is checked in a different part of the codebase (and we have tests for that). We want to check that the cache key is reset (the test I updated). Is that okay?

Yep for sure, good point.

andrewleith (approved):
LGTM!

PR #2061, opened 2023-12-10
Adjust start dates when fetching notification stats

whabanks:
# Summary | Résumé
- No longer subtracting 1 day from the retention period limit when calculating the start_date to begin pulling notification counts from.
- Changed the stats_for_today query to start pulling notifications starting from midnight UTC of the current day instead of `midnight today - retention period days` which was resulting in duplication of all notifications within the retention period minus the current day's notification count.

# Testing
### Example affected services
- [f316aa6b-38dd-4d9e-83f8-b6ee3e1d428f](https://staging.notification.cdssandbox.xyz/services/f316aa6b-38dd-4d9e-83f8-b6ee3e1d428f)
- [44c1a3f5-6a11-4952-b6f4-94dd46d009f2](https://staging.notification.cdssandbox.xyz/services/44c1a3f5-6a11-4952-b6f4-94dd46d009f2)

1. Hook your local environment up to a snapshot of the Staging DB. 
2. Check that the notification counts are reported corrected on the reports page. The smoke test service is a good example.

-OR-

1. Push this right into staging and check the counts there, reverting if adjustment needed.


app/dao/fact_notification_status_dao.py
```diff
@@ -260,7 +260,7 @@ def fetch_notification_status_for_service_for_today_and_7_previous_days(service_
             *([func.count().label("count")]),
         )
         .filter(
-            Notification.created_at >= start_date,
+            Notification.created_at >= get_midnight(datetime.utcnow()),
```
whabanks, line 263: This line was the core of the issue that inflated the notification counts. We weren't pulling just today's notifications in this query, but all notifications from starting from `today - days of  retention` to `today`.

app/dao/fact_notification_status_dao.py
```diff
@@ -239,7 +239,7 @@ def fetch_notification_status_for_service_for_day(bst_day, service_id):
 
 
 def fetch_notification_status_for_service_for_today_and_7_previous_days(service_id, by_template=False, limit_days=7):
-    start_date = utc_midnight_n_days_ago(limit_days - 1)
+    start_date = utc_midnight_n_days_ago(limit_days)
```
whabanks, line 242: I was hesitant to remove the `- 1` here as we added it in [this PR](https://github.com/cds-snc/notification-api/pull/1976/files#diff-4c1d9f51a839bbba6edbea4f69046a7410906281bc359526415b8fbd898fd673R242) for an edge case.

app/dao/fact_notification_status_dao.py
```diff
@@ -239,7 +239,7 @@ def fetch_notification_status_for_service_for_day(bst_day, service_id):
...
 
     stats_for_7_days = db.session.query(
```
whabanks, line 244: An idea that crossed my mind to make this code a bit more testable is pulling the `stats_for_7_days` and `stats_for_today` out into their own methods. We could then reuse them in other contexts in the future and get a bit more granular and focused with our testing considering we have had repeated issues in this area of the code.

jzbahrai: I like that idea, if you want to do it in this PR feel free

jzbahrai (approved):

andrewleith (approved):
Good work, this all makes good sense to me.  Thanks for being thorough here and getting to the bottom of this.

PR #2071, opened 2024-01-04
Fix ses-process-results error handling

andrewleith:
# Summary | Résumé

Initialize variables at the start so that error handling is simplified and existence of variables doesn't need to be checked before logging.


ben851 (approved):
I like variable init's

PR #2074, opened 2024-01-09
Match the 5AM->5AM time window used to populate ft_notifications

whabanks:
Reverts cds-snc/notification-api#2070 We pushed this in during code freeze to sanity test in preparation for code unfreeze. This PR is considered the final PR for this work.

# Summary | Résumé
When [the nightly task that populates ft_notification_status](https://github.com/cds-snc/notification-api/blob/main/app/celery/reporting_tasks.py#L60) runs, the timeframe it uses for fetching notifications to put into the table is 5AM to 5AM UTC, NOT midnight to midnight UTC. We can see this timeframe in the logs:

image

Since we used a midnight to midnight UTC timeframe to fetch today's notifications, notifications between `00:00` and `05:00` UTC were being counted twice

This PR adjusts the timeframe used to fetch the current day's notifications to match the timeframe used when populating the `ft_notification_status` table. Removing the 5 hour overlap where notifications were being counted twice. 

The date objects used are EST/EDT timezone aware, and will "self-adjust" when `timedelta` math is applied to them, this should handle blips that can occur when date calculations that put a date past an EST <-> EDT transition period are performed and make date math in this area of the code a bit less cumbersome to work with.

### Side note:
[`freezegun` seems to have limitations when working with time zone aware dates.](https://github.com/cds-snc/notification-api/pull/2069/files#diff-7d0715b38b4cffc6f79bf2a272e846ae4e8f4435673879ad43d3330b7ecca743R1141)

### DIFF
--- a/app/dao/date_util.py
+++ b/app/dao/date_util.py
@@ -83,6 +83,22 @@ def get_midnight(datetime: datetime) -> datetime:
     return datetime.replace(hour=0, minute=0, second=0, microsecond=0)
 
 
+def tz_aware_utc_now() -> datetime:
+    """
+    Returns a localized, EST/EDT timezone aware, UTC now datetime.
+    Call dst() on the returned object to determine daylight savings status.
+    """
+    return pytz.utc.localize(datetime.utcnow())
+
+
+def tz_aware_midnight_n_days_ago(days_ago: int = 1) -> datetime:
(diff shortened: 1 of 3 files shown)


jzbahrai (approved):
lgtm, will test on staging

PR #2075, opened 2024-01-09
Better nightly performance test

sastels:
# Summary | Résumé

Add an email send rate  test
- runs Tuesday through Saturday after midnight UTC
- send approx 100K emails over 10 minutes. This should be enough to scale to system to sustained max capacity.

Running locally, we get and hold a sustained 5500ish as we found during load tests
image

No warnings / errors either in CloudWatch 🎉 

Run manually locally / evaluate results after merging and running overnight.


tests-perf/locust/locust-notifications.py
```diff
@@ -26,15 +26,15 @@
...
+    wait_time = constant_pacing(60)  # 60 seconds between each task
     host = os.getenv("PERF_TEST_DOMAIN", "https://api.staging.notification.cdssandbox.xyz")
 
     def __init__(self, *args, **kwargs):
         super(NotifyApiUser, self).__init__(*args, **kwargs)
 
         self.headers = {"Authorization": os.getenv("PERF_TEST_AUTH_HEADER")}
         self.email = os.getenv("PERF_TEST_EMAIL", "success@simulator.amazonses.com")
-        self.phone_number = os.getenv("PERF_TEST_PHONE_NUMBER", "16132532222")
+        self.phone_number = os.getenv("PERF_TEST_PHONE_NUMBER", "16135550123")
```
sastels, line 37: we're not using this right now, but I changed to our internal test number just in case we do in the future.

sastels:
Just noticed that there's a perf test card that says to expand to an hour long test, let me do that first...

jimleroyer:
I spoke with Steve and we agreed to split that test in two so that we have one speicialized for hitting the API and one is  stressing the sending pipelines.

jzbahrai (dismissed):

bin/execute_and_publish_performance_test.sh
```diff
@@ -1,13 +1,24 @@
...
 aws s3 cp $perf_test_csv_directory_path/ "s3://$perf_test_aws_s3_bucket" --recursive || exit 1
 
+# Sleep 15 minutes to allow the system to stabilize
+sleep 900
+
+# Run email send rate performance test
+# This configuration should send 10K emails / minute for 10 minutes for 100K emails total.
+# We run this test on Tuesday through Friday (just after midnight UTC) only.
+if [ "$(date +%u)" -ge 2 ] && [ "$(date +%u)" -le 5 ]; then
+    locust --headless --host https://api.staging.notification.cdssandbox.xyz --locustfile tests-perf/locust/send_rate_email.py --users 5 --run-time 10m --spawn-rate 1
```
jimleroyer, line 20: Why is the locust test named send_rate_email? is this to send to a specific rate? or to hit the rate limit?

jimleroyer: We are not copying results into s3 for these?

sastels: We could but the main point in this test is just to see how high we get the send rate, which isn't going to be in the locust results.

sastels: We habve the blazer query but would be easy to add it to QuickSight too

sastels: Just saw your first question! I named it that since all it's testing is the max send rate we can achieve. Ie it does bulk sends to get the system running at max capacity. After the test we can look at a query / dashboard to see what send rate was reached.

jzbahrai (approved):

jimleroyer (approved):
LGTM! Thanks Steve!

PR #2079, opened 2024-01-16
Add a retry mechanism when fetching email attachements

whabanks:
# Summary | Résumé
Adds retry capabilities when fetching email attachments during the `send_to_providers` process. Uses [`urllib3` retry functionality](https://urllib3.readthedocs.io/en/stable/user-guide.html#retrying-requests).


app/delivery/send_to_providers.py
```diff
@@ -231,27 +233,30 @@ def send_email_to_provider(notification: Notification):
...
-                        filename = personalisation_data[key]["document"].get("filename")
-                        mime_type = personalisation_data[key]["document"].get("mime_type")
-                        attachments.append(
-                            {
-                                "name": filename,
-                                "data": buffer,
-                                "mime_type": mime_type,
-                            }
-                        )
+                    logging.getLogger("urllib3").setLevel(logging.DEBUG)
```
jzbahrai, line 244: why is this logging level debug? can it be warning?

jzbahrai (commented):
thanks for adding this! I assumed we had a retry functionality piece already, but turns out we only have those for async celery tasks. would be good to take what you have done and make it a decorator at some point.

whabanks:
> thanks for adding this! I assumed we had a retry functionality piece already, but turns out we only have those for async celery tasks. would be good to take what you have done and make it a decorator at some point.

Love the decorator idea!

app/delivery/send_to_providers.py
whabanks: Good catch, this should be removed. I was using it to debug which calls `urllib3` was making internally for mocking in tests and forgot to remove it.

jzbahrai (approved):

PR #2081, opened 2024-01-17
Add API Validation

jzbahrai:
# Summary | Résumé

We wanted to do three things:
1. Check the notify prefix is sent through the api key
2. Check the service key is part of the secret
3. Return the result from the secret key

# Testing:
Added unit tests and tested locally

# Considerations:
Anyone who isnt sending in a whole key, will have to update their api key

### DIFF
--- a/app/api_key/rest.py
+++ b/app/api_key/rest.py
@@ -121,8 +121,7 @@ def revoke_api_keys():
 
     # Step 1
     try:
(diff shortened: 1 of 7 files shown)


andrewleith (approved):
Looks great and works well!  One little comment about a test name with a typo in it.

PR #2093, opened 2024-01-24
Clean up celery config for beat schedule

jzbahrai:
# Summary | Résumé

The celery beat schedule needed to be fixed for UTC instead of EST. This PR does that.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -438,87 +438,54 @@ class Config(object):
         # app/celery/nightly_tasks.py
         "timeout-sending-notifications": {
             "task": "timeout-sending-notifications",
-            "schedule": crontab(hour=0, minute=5),
+            "schedule": crontab(hour=5, minute=5),  # 00:05 EST in UTC
             "options": {"queue": QueueNames.PERIODIC},
         },
         "create-nightly-billing": {
             "task": "create-nightly-billing",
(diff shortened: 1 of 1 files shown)


app/config.py
```diff
@@ -438,87 +438,54 @@ class Config(object):
         # app/celery/nightly_tasks.py
         "timeout-sending-notifications": {
             "task": "timeout-sending-notifications",
-            "schedule": crontab(hour=0, minute=5),
+            "schedule": crontab(hour=5, minute=5),  # 00:05 EST in UTC
```
whabanks, line 441: When the clock rolls back 1 hour on May 10th could this end up scheduling the task for 11:05PM EST the day before?
Otherwise looks good!

whabanks (changes requested):

app/config.py
jzbahrai: yes. I believe that is exactly what will happen

whabanks: Are we ok with that or should we use something like `pytz` to check the current DST status and adjust for consistent scheduling?

jzbahrai: I am okay with us not having a consistent schedule on one day, I cant think of downsides here

whabanks: Sounds good to me then!

whabanks (approved):

PR #2094, opened 2024-01-26
Prevent naming variations of .env files to be submited to git

jimleroyer:
# Summary | Résumé

Prevent naming variations of .env files to be submited to git.

This also removes an older encrypted secret file that should not be used anymore, i.e. `.env.enc.aws`.

Create a file named `.env.staging` locally and check if it can get submitted with git.


whabanks (approved):

sastels (approved):
:shipit:

PR #2097, opened 2024-02-01
Fix time range for for daily stats

andrewleith:
# Summary | Résumé

Update the time range used by daily stats in the `fetch_notification_status_for_service_for_today_and_7_previous_days` function to calculate over the time range since the previous midnight UTC.  This is to align with how our daily limits work.

## Test instructions | Instructions pour tester la modification

- Sends before midnight UTC should not be part of the total contained in the `Daily Usage` section

## Release Instructions | Instructions pour le déploiement

- None.


app/dao/fact_notification_status_dao.py
```diff
@@ -313,6 +319,86 @@ def fetch_notification_status_for_service_for_today_and_7_previous_days(service_
     ).all()
 
 
+def fetch_daily_notification_status_for_service(service_id, by_template=False):
```
jzbahrai, line 322: this is a new function? I don't see where it is called in this PR?

app/dao/fact_notification_status_dao.py
```diff
@@ -313,6 +319,86 @@ def fetch_notification_status_for_service_for_today_and_7_previous_days(service_
...
+        )
+    )
+
+    all_stats_table = stats_for_today.subquery()
+
+    query = db.session.query(
+        *(
+            [
+                Template.name.label("template_name"),
+                Template.is_precompiled_letter,
```
jzbahrai, line 356: I am proabaly missing this but why do we need the line is_precompiled_letter?

app/dao/fact_notification_status_dao.py
andrewleith: Ah my bad, at first I was going to make a new method so I wouldnt affect other things, but I thought better of that.  Let me clean that up.

jzbahrai (approved):

whabanks (approved):
LGTM! Let's get this in staging and see how it goes, fingers crossed!

app/dao/fact_notification_status_dao.py
andrewleith: This is existing code - not sure.

PR #2117, opened 2024-02-21
Revert "Task: Filter Heartbeats"

whabanks:
Reverts cds-snc/notification-api#2108

The language toggle on the index page was causing major slow downs and in some cases full timeouts.


jimleroyer (approved):
LGTM

PR #2127, opened 2024-03-04
Fix/restrict job query size

sastels:
# Summary | Résumé

Stop `dao_get_jobs_older_than_data_retention` from crashing.

We now have almost 500K unarchived jobs that are more than a week old. This is crashing the query so that none are getting archived and the list grows. This PR sets the limit to the query to 20K, which is how many jobs / day we had in the past without any OOM issues.

* https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/307

- change [this line](https://github.com/cds-snc/notification-api/blob/260722e4ffdcf9b2f865583161dd6642e9f36fad/app/config.py#L486) to `"schedule": 60,` (to run the task every minute)
- change [this line](https://github.com/cds-snc/notification-api/blob/260722e4ffdcf9b2f865583161dd6642e9f36fad/app/dao/jobs_dao.py#L153) to `end_date = today + timedelta(days=7)` (to remove all jobs, not just old ones)
- change the [query limit](https://github.com/cds-snc/notification-api/pull/2127/files#diff-dd4f118014701d2a8a706498e3c2d318a7aa191065f0a2e3d61aa5633c24c3a8R53) from `20000` to `2`
- purge your queues with `make run-celery-purge`
- run celery and api with `make run-celery-local` and `make run`
- POST several (5 or so) jobs job to your local /bulk endpoint. 
- After a minute you should see something like (intermingled with other lines):
```
[2024-03-04 18:44:48,955: INFO/MainProcess] Task remove_sms_email_jobs[a813a22c-94df-4dbd-9bef-7129278672a3] received
...
[2024-03-04 18:44:48,974: INFO/ForkPoolWorker-2] Archiving 2 jobs.
...
[2024-03-04 18:44:50,110: INFO/ForkPoolWorker-2] Archiving 2 jobs.
...
[2024-03-04 18:44:51,072: INFO/ForkPoolWorker-2] Archiving 1 jobs.
```

- Should see lines in the logs similar to `Archiving 20000 jobs.`
- Should see the number of unarchived jobs in the database drop significantly after the first time this code runs

### DIFF
--- a/app/dao/jobs_dao.py
+++ b/app/dao/jobs_dao.py
@@ -129,7 +129,7 @@ def dao_update_job(job):
     db.session.commit()
 
 
-def dao_get_jobs_older_than_data_retention(notification_types):
+def dao_get_jobs_older_than_data_retention(notification_types, limit=None):
     flexible_data_retention = ServiceDataRetention.query.filter(
         ServiceDataRetention.notification_type.in_(notification_types)
     ).all()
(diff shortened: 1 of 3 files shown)


app/dao/jobs_dao.py
```diff
@@ -162,8 +163,10 @@ def dao_get_jobs_older_than_data_retention(notification_types):
                 Job.service_id.notin_(services_with_data_retention),
             )
             .order_by(desc(Job.created_at))
-            .all()
         )
+        if limit:
+            query = query.limit(limit)
```
jimleroyer, line 168: So the limit would apply on 2 different queries and if the distribution of the data is split evenly across flexible and non-flexible data retention, we'd get 40K jobs for a 20K jobs. We can assume that this is not the case and it will run fine, as we would have less services with flexible data retention.. but we could easily prevent that if that comes up at that point with:
```
query = query.limit(limit - len(jobs))
```
(untested code; might be wrong!)

app/dao/jobs_dao.py
```diff
@@ -129,16 +129,15 @@ def dao_update_job(job):
     db.session.commit()
 
 
-def dao_get_jobs_older_than_data_retention(notification_types):
+def dao_get_jobs_older_than_data_retention(notification_types, limit=None):
```
whabanks, line 132: Do we need to change [`remove_transformed_dvla_files`](https://github.com/cds-snc/notification-api/blob/main/app/celery/nightly_tasks.py#L196)? Are we even running this [task](https://github.com/cds-snc/notification-api/blob/main/app/config.py#L479)?

sastels: That's old letter code that apparently [we're still running](https://github.com/cds-snc/notification-api/blob/bfadea3bb7ba83cf57d3d206e907b01bfa9239ef/app/config.py#L479-L482) but should just be doing nothing

app/dao/jobs_dao.py
sastels: ah, very good idea 👍

tests/app/dao/test_jobs_dao.py
```diff
@@ -348,6 +348,32 @@ def test_should_get_jobs_seven_days_old_by_scheduled_for_date(sample_service):
...
+
+    create_job(sample_template, created_at=eight_days_ago)
+    create_job(sample_template, created_at=eight_days_ago)
+    create_job(sample_template, created_at=eight_days_ago)
+
+    jobs = dao_get_jobs_older_than_data_retention(notification_types=[sample_template.template_type])
+
+    assert len(jobs) == 3
+
+
```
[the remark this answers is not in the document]
sastels: Test updated!

jimleroyer (approved):
LGTM!

PR #2134, opened 2024-03-07
K8s rollout workflow

ben851:
# Summary | Résumé

This will move the Kubernetes rollout in the docker build GitHub Action to the new workflow in notifications-manifest.

This is required in order to support having a private EKS cluster.

## Related Issues | Cartes liées

* [https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/1
](https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/293)

# Test instructions | Instructions pour tester la modification

This can't be tested until it's merged due to issues with the AWS Auth. 

I have removed the docker build/push steps and ensured that the call from this workflow to the manifests workflow works. 

I will be standing by to revert this in the event it doesn't work.

# Release Instructions | Instructions pour le déploiement

Release, and ensure that the kubernetes rollout job in notification-manifests is run.

# Reviewer checklist | Liste de vérification du réviseur

- [ ] This PR does not break existing functionality.
- [ ] This PR does not violate GCNotify's privacy policies.
- [ ] This PR does not raise new security concerns. Refer to our GC Notify Risk Register document on our Google drive.
- [ ] This PR does not significantly alter performance.
- [ ] Additional required documentation resulting of these changes is covered (such as the README, setup instructions, a related ADR or the technical documentation).

> ⚠ If boxes cannot be checked off before merging the PR, they should be moved to the "Release Instructions" section with appropriate steps required to verify before release. For example, changes to celery code may require tests on staging to verify that performance has not been affected.

PR #2136, opened 2024-03-07
Revert "K8s rollout workflow"

ben851:
Reverts cds-snc/notification-api#2134

Restores the workflow since the sha doesn't seem to get correctly populated in the other job.


sastels (approved):
Getting close!

PR #2151, opened 2024-03-28
Reverting #2099

jzbahrai:
# Summary | Résumé

Reverting this: https://github.com/cds-snc/notification-api/pull/2099/files


jimleroyer:
This can be closed as the root of the incident that occurred on 2024-03-28 was determined to be other than the api key changes.

PR #2152, opened 2024-04-05
Use pinpoint for designated templates

sastels:
# Summary | Résumé

We add two main features here:
  - Use an AWS Pinpoint pool to send messages from specified templates (other templates not affected).
  - add a task to process pinpoint delivery receipts

This code can be merged in and will sit unused unless
- the required env vars are set
- an sms is sent using one of the designated templates.

In other cases the normal sns code will be used to send the text.

* https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/286

- set your `AWS_PINPOINT_SC_POOL_ID` env variable to the pool id in staging `AWS_PINPOINT_POOL_ID = "pool-b20333ce1e4e49309ba1db3bf94a3f57"`
- set the env var `AWS_PINPOINT_SC_TEMPLATE_IDS` to a list of your sms template ids. Just one is ok, ie `AWS_PINPOINT_SC_TEMPLATE_IDS = 000000-1111-2222-3333-444444444` 
- send an sms using a template in the list. You should see a line such as `AWS Pinpoint request finished in ...` in the celery logs
- send an sms using a different template. Instead of the Pinpoint line you should see `AWS SNS request finished in ...`
- note that both sends could be from the same number.

None.


app/celery/process_pinpoint_receipts_tasks.py
```diff
@@ -0,0 +1,147 @@
...
+#         "messageId": "221bc70c-7ee6-4987-b1ba-9684ba25be20",
+#         "messageRequestTimestamp": 1712944267685,
+#         "messageEncoding": "GSM",
+#         "messageType": "TRANSACTIONAL",
+#         "messageStatus": "DELIVERED",
+#         "messageStatusDescription": "Message has been accepted by phone",
+#         "totalMessageParts": 1,
+#         "totalMessagePrice": 0.00581,
+#         "totalCarrierFee": 0.006
+#     }
```
jimleroyer, line 42: Nice documentation to have in the code!

.env.example
```diff
@@ -19,3 +19,6 @@ AWS_PINPOINT_REGION=us-west-2
 AWS_EMF_ENVIRONMENT=local
 
 CONTACT_FORM_EMAIL_ADDRESS = ""
+
+AWS_PINPOINT_POOL_ID=
+AWS_PINPOINT_TEMPLATE_IDS=
```
jimleroyer, line 24: Are these two would be meant for the short code specifically? Because on long term, we will want to have a several pools with a 1:M mapping to templates, we might want to specify for now this is only for short code in the environment name. It might be better too in case that this stays this way for longer than we expect and cannot complete the short code integration work as we plan it at the moment.

.env.example
```diff
@@ -19,3 +19,6 @@ AWS_PINPOINT_REGION=us-west-2
 AWS_EMF_ENVIRONMENT=local
 
 CONTACT_FORM_EMAIL_ADDRESS = ""
+
+AWS_PINPOINT_POOL_ID=
```
jimleroyer, line 23:
```suggestion
AWS_PINPOINT_SC_POOL_ID=
```

app/clients/sms/aws_pinpoint.py
```diff
@@ -0,0 +1,61 @@
...
+        # super(SmsClient, self).__init__(*args, **kwargs)
+        self.current_app = current_app
+        self.name = "pinpoint"
+        self.statsd_client = statsd_client
+
+    def get_name(self):
+        return self.name
+
+    def send_sms(self, to, content, reference, multi=True, sender=None):
+        pool_id = self.current_app.config["AWS_PINPOINT_POOL_ID"]
```
jimleroyer, line 27:
```suggestion
        pool_id = self.current_app.config["AWS_PINPOINT_SC_POOL_ID"]
```

app/config.py
```diff
@@ -266,6 +266,9 @@ class Config(object):
     AWS_SES_ACCESS_KEY = os.getenv("AWS_SES_ACCESS_KEY")
     AWS_SES_SECRET_KEY = os.getenv("AWS_SES_SECRET_KEY")
     AWS_PINPOINT_REGION = os.getenv("AWS_PINPOINT_REGION", "us-west-2")
+    AWS_PINPOINT_POOL_ID = os.getenv("AWS_PINPOINT_POOL_ID", None)
```
jimleroyer, line 269:
```suggestion
    AWS_PINPOINT_SC_POOL_ID = os.getenv("AWS_PINPOINT_SC_POOL_ID", None)
```

app/aws/mocks.py
```diff
@@ -192,6 +192,57 @@ def sns_failed_callback(provider_response, reference=None, timestamp="2016-06-28
...
+        "messageEncoding": "GSM",
+        "messageType": "TRANSACTIONAL",
+        "messageStatus": "DELIVERED",
+        "messageStatusDescription": "Message has been accepted by phone",
+        "totalMessageParts": 1,
+        "totalMessagePrice": 0.00581,
+        "totalCarrierFee": 0.006,
+    }
+
+    return _pinpoint_callback(body)
```
jimleroyer, line 219: I like these mocks and it's a good doc to know what is returned by the callback.

app/aws/mocks.py
```diff
@@ -192,6 +192,57 @@ def sns_failed_callback(provider_response, reference=None, timestamp="2016-06-28
...
+        "eventTimestamp": timestamp,
+        "isFinal": True,
+        "originationPhoneNumber": "+13655550100",
+        "destinationPhoneNumber": destination,
+        "isoCountryCode": "CA",
+        "messageId": reference,
+        "messageRequestTimestamp": timestamp,
+        "messageEncoding": "GSM",
+        "messageType": "TRANSACTIONAL",
+        "messageStatus": "CARRIER_UNREACHABLE",
```
jimleroyer, line 236: I expect the returned status to be the same as AWS SNS? Maybe we could output these as a log for our testing so we can get the unique values and if that differ from the current implementation?

app/aws/mocks.py
```diff
@@ -192,6 +192,57 @@ def sns_failed_callback(provider_response, reference=None, timestamp="2016-06-28
...
+        "messageStatusDescription": provider_response,
+        "totalMessageParts": 1,
+        "totalMessagePrice": 0.00581,
+        "totalCarrierFee": 0.006,
```
jimleroyer, line 240: Let's put this in our database later! 🙏

app/celery/process_pinpoint_receipts_tasks.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+@notify_celery.task(bind=True, name="process-pinpoint-result", max_retries=5, default_retry_delay=300)
+@statsd(namespace="tasks")
+def process_pinpoint_results(self, response):
```
jimleroyer, line 47: Was this part of the resurrected code too? Not to be done right now, but as we want to move away from Celery processing the callbacks, I was wondering if that would not be a good opportunity to have pinpoint results processing all in lambda sooner with this work. What do you think?

app/celery/process_pinpoint_receipts_tasks.py
```diff
@@ -0,0 +1,145 @@
...
+        statsd_client.incr(f"callback.sns.{notification_status}")  # TODO: do we want a Pinpoint metric here?
+
+        if notification.sent_at:
+            statsd_client.timing_with_dates(
+                "callback.sns.elapsed-time", datetime.utcnow(), notification.sent_at
+            )  # TODO: do we want a Pinpoint metric here?
+
+        _check_and_queue_callback_task(notification)
+
+        return True
```
jimleroyer, line 108: Why do we return true here, but simple return (no value) for others? This was also picked up by our scanner over here:
https://github.com/cds-snc/notification-api/security/code-scanning/804

app/celery/process_pinpoint_receipts_tasks.py
```diff
@@ -0,0 +1,145 @@
...
+
+    except Retry:
+        raise
+
+    except Exception as e:
+        current_app.logger.exception(f"Error processing Pinpoint results: {str(e)}")
+        self.retry(queue=QueueNames.RETRY)
+
+
+def determine_pinpoint_status(sns_status, provider_response):
```
jimleroyer, line 118: Can we add typing info to the function's signature please? 🙏

app/celery/process_pinpoint_receipts_tasks.py
```diff
@@ -0,0 +1,145 @@
...
+        "Phone is on a blocked list": NOTIFICATION_TECHNICAL_FAILURE,
+        "Phone is currently unreachable/unavailable": NOTIFICATION_PERMANENT_FAILURE,
+        "Phone number is opted out": NOTIFICATION_PERMANENT_FAILURE,
+        "This delivery would exceed max price": NOTIFICATION_TECHNICAL_FAILURE,
+        "Unknown error attempting to reach phone": NOTIFICATION_TECHNICAL_FAILURE,
+    }
+
+    status = reasons.get(provider_response)  # could be None
+    if not status:
+        # TODO: Pattern matching in Python 3.10 should simplify this overall function logic.
```
jimleroyer, line 141: This was copied over older code I assume? we can try this pattern matching now? 👀

app/celery/process_pinpoint_receipts_tasks.py
```diff
@@ -0,0 +1,145 @@
...
+                    f"Pinpoint delivery failed: notification id {notification.id} and reference {reference} has error found. "
+                    f"Provider response: {provider_response}"
+                )
+            )
+        else:
+            current_app.logger.info(
+                f"Pinpoint callback return status of {notification_status} for notification: {notification.id}"
+            )
+
+        statsd_client.incr(f"callback.sns.{notification_status}")  # TODO: do we want a Pinpoint metric here?
```
jimleroyer, line 99: I would say yes! it might come up very useful to have our own pinpoint metric, the downside though is we'd have to update our dashboards with duplicate widgets for the ones tracking the SNS one.

app/celery/process_pinpoint_receipts_tasks.py
```diff
@@ -0,0 +1,145 @@
...
+
+        if notification.sent_at:
+            statsd_client.timing_with_dates(
+                "callback.sns.elapsed-time", datetime.utcnow(), notification.sent_at
+            )  # TODO: do we want a Pinpoint metric here?
```
jimleroyer, line 104: Also same here! I would think yes.

app/clients/sms/aws_pinpoint.py
```diff
@@ -0,0 +1,61 @@
...
+        messageType = "TRANSACTIONAL"
+        matched = False
+
+        for match in phonenumbers.PhoneNumberMatcher(to, "US"):
```
jimleroyer, line 31: The `US` parameter is to say the phone number has an American phone number format?

app/clients/sms/aws_pinpoint.py
```diff
@@ -0,0 +1,61 @@
...
+            matched = True
+            to = phonenumbers.format_number(match.number, phonenumbers.PhoneNumberFormat.E164)
+            destinationNumber = to
+
+            try:
+                start_time = monotonic()
+                response = self._client.send_text_message(
+                    DestinationPhoneNumber=destinationNumber,
+                    OriginationIdentity=pool_id,
```
jimleroyer, line 40: 🥳

app/clients/sms/aws_pinpoint.py
```diff
@@ -0,0 +1,61 @@
...
+                    MessageBody=content,
+                    MessageType=messageType,
+                    ConfigurationSetName=self.current_app.config["AWS_PINPOINT_CONFIGURATION_SET_NAME"],
+                )
+            except ClientError as e:
+                self.statsd_client.incr("clients.pinpoint.error")
+                raise Exception(e)
+            except Exception as e:
```
jimleroyer, line 48: I think we can combine these two if the logic is exactly the same.

```suggestion
            except (ClientError, Exception) as e:
```

There might be syntactic sugar to avoid the tuple and rather use the `|` operator nowadays, not sure..

app/delivery/send_to_providers.py
```diff
@@ -334,9 +336,14 @@ def update_notification_to_sending(notification, provider):
     dao_update_notification(notification)
 
 
-def provider_to_use(notification_type, notification_id, international=False, sender=None):
+def provider_to_use(notification_type, notification_id, international=False, sender=None, template_id=None):
+    if notification_type == SMS_TYPE and template_id is not None and str(template_id) in Config.AWS_PINPOINT_TEMPLATE_IDS:
```
jimleroyer, line 340:
```suggestion
    # Temporary redirect setup for template IDs that are meant for the short code usage.
    if notification_type == SMS_TYPE and template_id is not None and str(template_id) in Config.AWS_PINPOINT_TEMPLATE_IDS:
```

app/aws/mocks.py
sastels: 💯

app/aws/mocks.py
sastels: Currently if we do not recognize the status we get back [we log a warning](https://github.com/cds-snc/notification-api/pull/2152/files#diff-b77addf5c57c375a9d50ec45b68ae498dd4b7fecee7f4f4fec96a2049ee42733R56).

We should definitely keep an eye on these. Possibly we should add an alert that would warn us about these in Slack.

app/celery/process_pinpoint_receipts_tasks.py
sastels: yes it is mostly the old code (and also pretty much a copy of the current sns/sms receipt processing).

I thought about moving it all to the new lambda, but figured that for a first step we just get this all in and working and then we could think about refactoring.

Admittedly, the pinpoint receipts would be a good first target if, for example, we restrict the shortcode to GCNotify initially.

app/celery/process_pinpoint_receipts_tasks.py
sastels: Yeah, this is a good question. It's what the other tasks do, and all I could think of was that the tests were returning `None` if they failed and this `True` would indicate a success. My googling wasn't able to confirm or dispute this theory however :/

app/celery/process_pinpoint_receipts_tasks.py
sastels: Done! plus a docstring. Unfortunately I couldn't figure out a way to restrict the return values to `NOTIFICATION_DELIVERED`, `NOTIFICATION_TECHNICAL_FAILURE`, etc. Typing really didn't want variables involved.

app/celery/process_pinpoint_receipts_tasks.py
jimleroyer: Hmm I wonder if this function was called elsewhere maybe and hence expected to return something, maybe the older CLI commands? 🤔

app/celery/process_pinpoint_receipts_tasks.py
jimleroyer: Yeah we'd have to turn these into `Enum` I guess. `str` should be good enough for now. Thanks!

app/celery/process_pinpoint_receipts_tasks.py
sastels: ha! I was excluding our tests when looking to see how it's called. Our tests use the return value to determine if the function successfully processed the receipt (`True`) or if there was a problem (`None`).

I'll add a `return` at the end of the function to quiet down the scanner.

app/celery/process_pinpoint_receipts_tasks.py
sastels: I did some matching but I think it's cleaner refactored into an if/elif block. 🤔

app/celery/process_pinpoint_receipts_tasks.py
```diff
@@ -0,0 +1,150 @@
...
+        case response_lower if "unknown error" in response_lower:
+            return NOTIFICATION_TECHNICAL_FAILURE
+        case response_lower if "exceed max price" in response_lower:
+            return NOTIFICATION_TECHNICAL_FAILURE
+        case "phone carrier is currently unreachable/unavailable":
+            return NOTIFICATION_TEMPORARY_FAILURE
+        case "phone is currently unreachable/unavailable":
+            return NOTIFICATION_PERMANENT_FAILURE
+
+    return None
```
sastels, line 150: I think this is the cleanest way to use matching, but I think it's even better with if/elifs:
```
    if "blocked" in response_lower:
        return NOTIFICATION_TECHNICAL_FAILURE
    elif "invalid" in response_lower:
        return NOTIFICATION_TECHNICAL_FAILURE
    elif "is opted out" in response_lower:
        return NOTIFICATION_PERMANENT_FAILURE
    elif "unknown error" in response_lower:
        return NOTIFICATION_TECHNICAL_FAILURE
    elif "exceed max price" in response_lower:
        return NOTIFICATION_TECHNICAL_FAILURE
    elif "phone carrier is currently unreachable/unavailable" in response_lower:
        return NOTIFICATION_TEMPORARY_FAILURE
    elif "phone is currently unreachable/unavailable" in response_lower:
        return NOTIFICATION_PERMANENT_FAILURE
    else:
        return None
```

app/celery/process_pinpoint_receipts_tasks.py
```diff
@@ -0,0 +1,145 @@
...
+        if notification.sent_at:
+            statsd_client.timing_with_dates(
+                "callback.sns.elapsed-time", datetime.utcnow(), notification.sent_at
+            )  # TODO: do we want a Pinpoint metric here?
+
+        _check_and_queue_callback_task(notification)
+
+        return True
+
+    except Retry:
```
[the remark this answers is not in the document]
sastels: That's basically my understanding - if a task is retried then a `Retry` exception is raised, and we want to guard the subsequent "something went wrong, let's manually retry it" code in this case.

We could add some sort of `if e != Retry` to the `except Exception as e:` block instead, but I think the way we already have it is more pythonic. Or were you thinking of a different approach to refactoring?

app/celery/process_pinpoint_receipts_tasks.py
sastels: I shall add, though it's not clear to me that we have the sns ones in alarms or dashboards... at least, I've found `callback_ses_elapsed-time` in our SLO dashboard, but `callback_sns_elapsed-time` does not appear to be used anywhere... Possibly we should have a "callbacks" dashboard. with all the details.

app/clients/sms/aws_pinpoint.py
sastels: It's used as the default region, ie if you don't start your number with a region code then it assumes it's US / CA.

app/clients/sms/aws_pinpoint.py
sastels: 🤔 a `ClientError` is an `Exception` so I think we can just get rid of the `ClientError` bit alltogether!

app/celery/process_pinpoint_receipts_tasks.py
jimleroyer: Yeah the match case isn't the most terse I have seen in a language unfortunately. I hope Python improves on that. I wouldn't mind the `if...elif...else`. :\

app/celery/process_pinpoint_receipts_tasks.py
jimleroyer: I would expect the tests to be able to determine the end by verifying if the `_check_and_queue_callback_task` function was called but the code wanted it simpler I guess. 

Do you still want to remove that `True` here? 👀

app/celery/process_pinpoint_receipts_tasks.py
```diff
@@ -0,0 +1,150 @@
...
+        raise
+
+    except Exception as e:
+        current_app.logger.exception(f"Error processing Pinpoint results: {str(e)}")
+        self.retry(queue=QueueNames.RETRY)
+
+    return
```
jimleroyer, line 116: Not sure we need this `return` statement here.

app/celery/process_pinpoint_receipts_tasks.py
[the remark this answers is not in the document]
jimleroyer: I didn't have anything specific in mind unfortunately and it seems it is the best way we have!

app/celery/process_pinpoint_receipts_tasks.py
sastels: yeah... having code in the app just for test verification is ugly and also could lead to tests passing when it doesn't really work (like if there was a `return True` where there shouldn't be). I changed the tests to just checking that callback task as you suggested.

app/celery/process_pinpoint_receipts_tasks.py
sastels: 🔥

jimleroyer (approved):
:shipit: 🥳

PR #2163, opened 2024-04-25
fix/use a service's fixed phone number if set

sastels:
# Summary | Résumé

Have a case where a [service with a dedicated number is not using the number for bulk sends](https://cds-snc.freshdesk.com/a/tickets/17255) though it is for individual sends.

Found that the `reply_to_text` is not being set for these notifications. Admin one-offs still are working fine, but the other 3 ways to send are affected.

I think there's an issue where foreign keys set in other tables but referencing the `services` table do not appear in the services table when we cache it. In this case, though, we can just set the reply_to_text / sender_id when the job or notification is created.

- Set a dedicated phone number for a service (locally you can just choose one from staging and add it to your local setup as an inbound number, then assign it to your service)
- send yourself some sms. All should come from the dedicated number.
- go to another service that doesn't have a set number
- send yourself sms
- all should come from a number different from the dedicated number (ie one of our numbers in ca-central-1)

Ask [user](https://cds-snc.freshdesk.com/a/tickets/17255) to test.

### DIFF
--- a/app/celery/tasks.py
+++ b/app/celery/tasks.py
@@ -8,7 +8,10 @@ from uuid import UUID, uuid4
 from flask import current_app
 from itsdangerous import BadSignature
 from more_itertools import chunked
-from notifications_utils.recipients import RecipientCSV
+from notifications_utils.recipients import (
+    RecipientCSV,
+    try_validate_and_format_phone_number,
+)
 from notifications_utils.statsd_decorators import statsd
 from notifications_utils.template import SMSMessageTemplate, WithSubjectTemplate
 from notifications_utils.timezones import convert_utc_to_local_timezone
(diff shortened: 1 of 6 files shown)


app/celery/tasks.py
```diff
@@ -245,7 +248,9 @@ def save_smss(self, service_id: Optional[str], signed_notifications: List[Signed
 
         reply_to_text = ""  # type: ignore
         if sender_id:
-            reply_to_text = dao_get_service_sms_senders_by_id(service_id, sender_id).sms_sender
+            reply_to_text = try_validate_and_format_phone_number(
+                dao_get_service_sms_senders_by_id(service_id, sender_id).sms_sender
+            )
```
sastels, line 253: The old code is wrong - later on we ignore the reply_to_text if it doesn't start with a "+".

app/celery/tasks.py
```diff
@@ -243,13 +246,18 @@ def save_smss(self, service_id: Optional[str], signed_notifications: List[Signed
         sender_id = _notification.get("sender_id")  # type: ignore
         notification_id = _notification.get("id", create_uuid())
 
-        reply_to_text = ""  # type: ignore
-        if sender_id:
-            reply_to_text = dao_get_service_sms_senders_by_id(service_id, sender_id).sms_sender
-        elif template.service:
-            reply_to_text = template.get_reply_to_text()
+        if "reply_to_text" in _notification and _notification["reply_to_text"]:
+            reply_to_text = _notification["reply_to_text"]
```
sastels, line 250: The code allows us to pass it in, but we were explicitly ignoring the passed in value (?)

whabanks (approved):
LGTM - Tested locally per instructions and both one offs and bulk sends are using the dedicated number as expected! 🚀

jimleroyer (commented):
Can we add some tests with this new scenario where the sender IDs would be null and needs to be applied?

jimleroyer (approved):
LGTM!

sastels:
Will merge and test tomorrow!

PR #2173, opened 2024-05-08
Pinpoint for everyone

sastels:
# Summary | Résumé

Use Pinpoint to send the non-shortcode SMS using the default pool

* https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/286

- set your `AWS_PINPOINT_SC_POOL_ID` and  `AWS_PINPOINT_DEFAULT_POOL_ID` environment variables to the pool ids in staging 
```
AWS_PINPOINT_DEFAULT_POOL_ID = "pool-8885654b47d6466f9aaee14c62494ff2"
```

- set the env var `AWS_PINPOINT_SC_TEMPLATE_IDS` to a list of some sms template ids. Just one is ok
```
AWS_PINPOINT_SC_TEMPLATE_IDS = 000000-1111-2222-3333-444444444
```
- send an sms using a template in the list. You should see a line such as `AWS Pinpoint request finished in ...` in the celery logs and the text should be sent from one of the numbers in the shortcode pool
- send an sms using a different template. You should see a line such as `AWS Pinpoint request finished in ...` in the celery logs and the text should be sent from one of the numbers in the default pool
- Set `AWS_PINPOINT_SC_POOL_ID` and `AWS_PINPOINT_DEFAULT_POOL_ID` to empty (or don't set) and send an SMS with a template in and out of the shortcode list. Both should be sent with SNS, ie you should see in the logs `AWS SNS request finished in ...`

None.

Note in particular that if the pool_id variables are not sent then we default to just using SNS.


tests/app/clients/test_aws_pinpoint.py
```diff
@@ -0,0 +1,73 @@
+import pytest
+
+from app import aws_pinpoint_client
+from tests.conftest import set_config_values
+
+
+@pytest.mark.serial
```
sastels, line 7: There's some sort of race condition in our tests that I couldn't figure out :/ If these two tests are not run serially then something resets the config variables that they change.

app/clients/sms/aws_sns.py
```diff
@@ -66,12 +65,9 @@ def send_sms(self, to, content, reference, multi=True, sender=None):
             try:
                 start_time = monotonic()
                 response = client.publish(PhoneNumber=to, Message=content, MessageAttributes=attributes)
-            except botocore.exceptions.ClientError as e:
-                self.statsd_client.incr("clients.sns.error")
-                raise str(e)
             except Exception as e:
                 self.statsd_client.incr("clients.sns.error")
-                raise str(e)
```
sastels, line 74: This stuff was redundant and also the raise didn't work! You can't raise a string, just an `Exception`.

jimleroyer: pylint for the win! I feel bad we had that for that long in the code. The whole error management does look weird in retrospective, thank you for refactoring.

app/clients/sms/aws_sns.py
```diff
@@ -66,12 +65,9 @@ def send_sms(self, to, content, reference, multi=True, sender=None):
...
+                raise Exception(e)
```
jimleroyer, line 70: Most likely it would be enough to re-raise the exception as-is as we do not add any new type, context or message to it? That should be less obfuscating as well instead of wrapping it up into a generic `Exception` type.
```suggestion
                raise e
```

app/clients/sms/aws_pinpoint.py
```diff
@@ -14,19 +14,22 @@ class AwsPinpointClient(SmsClient):
...
-    def send_sms(self, to, content, reference, multi=True, sender=None):
-        pool_id = self.current_app.config["AWS_PINPOINT_SC_POOL_ID"]
+    def send_sms(self, to, content, reference, multi=True, sender=None, template_id=None):
         messageType = "TRANSACTIONAL"
         matched = False
 
+        if template_id is not None and str(template_id) in self.current_app.config["AWS_PINPOINT_SC_TEMPLATE_IDS"]:
+            pool_id = self.current_app.config["AWS_PINPOINT_SC_POOL_ID"]
+        else:
+            pool_id = self.current_app.config["AWS_PINPOINT_DEFAULT_POOL_ID"]
```
jimleroyer, line 31: What happens if the pool_id is `None`? Is a random `OriginationIdentity` would be used then?

app/config.py
```diff
@@ -267,6 +267,7 @@ class Config(object):
     AWS_SES_SECRET_KEY = os.getenv("AWS_SES_SECRET_KEY")
     AWS_PINPOINT_REGION = os.getenv("AWS_PINPOINT_REGION", "us-west-2")
     AWS_PINPOINT_SC_POOL_ID = os.getenv("AWS_PINPOINT_SC_POOL_ID", None)
+    AWS_PINPOINT_DEFAULT_POOL_ID = os.getenv("AWS_PINPOINT_DEFAULT_POOL_ID", None)
```
jimleroyer, line 270: I wonder if we really want `None` here as a fallback default, or rather prevent the server from booting up to prevent a misconfiguration. It depends how `OriginationIdentity` set to `None` in the `send_text_message` call behaves.

app/delivery/send_to_providers.py
```diff
@@ -336,16 +339,55 @@ def update_notification_to_sending(notification, provider):
...
+def provider_to_use(
+    notification_type: str,
+    notification_id: UUID,
+    to: Union[str, None] = None,
+    international: bool = False,
+    sender: Union[str, None] = None,
+) -> Any:
+    """
+    Get the provider to use for sending the notification.
+    SMS that are being sent with a dedicated number or to a US number should not use Pinpoint.
```
jimleroyer, line 351: Thanks for the documentation and typing annotations.

app/config.py
jimleroyer: Ok reading the code further, it seems that a `None` value would route the SNS implementation to get used instead. I got this right?

tests/app/clients/test_aws_pinpoint.py
jimleroyer: It would be a great topic for a code review session. It seems we have a misunderstanding on our some of our test concurrency is setup and fails when they should be successful. We can put our mind to it and try to brainstorm on this.

tests/app/clients/test_aws_pinpoint.py
```diff
@@ -0,0 +1,73 @@
...
+    mocker.patch.object(aws_pinpoint_client, "_client", create=True)
+    mocker.patch.object(aws_pinpoint_client, "statsd_client", create=True)
+
+    to = ""
+    content = reference = "foo"
+
+    with pytest.raises(ValueError) as excinfo:
+        aws_pinpoint_client.send_sms(to, content, reference)
+
+    assert "No valid numbers found for SMS delivery" in str(excinfo.value)
```
jimleroyer, line 73: Should we have a warning alert in TF specifically for this one? 🤔 I guess it might fall into celery errors but these might be generic enough that we don't always check on these?

tests/app/dao/test_provider_details_dao.py
```diff
@@ -241,9 +241,14 @@ def test_get_sms_provider_with_equal_priority_returns_provider(
 
 
 def test_get_current_sms_provider_returns_active_only(restore_provider_details):
+    # Note that we currently have two acive sms providers, sns and pinpoint
```
jimleroyer, line 244:
```suggestion
    # Note that we currently have two active sms providers: sns and pinpoint.
```

app/delivery/send_to_providers.py
```diff
@@ -336,16 +339,55 @@ def update_notification_to_sending(notification, provider):
...
 
 
-def provider_to_use(notification_type, notification_id, international=False, sender=None, template_id=None):
-    # Temporary redirect setup for template IDs that are meant for the short code usage.
-    if notification_type == SMS_TYPE and template_id is not None and str(template_id) in Config.AWS_PINPOINT_SC_TEMPLATE_IDS:
-        return clients.get_client_by_name_and_type("pinpoint", SMS_TYPE)
+def provider_to_use(
+    notification_type: str,
+    notification_id: UUID,
+    to: Union[str, None] = None,
```
jimleroyer, line 345: A more up to date syntax or type might be (after a proper `import typing.Optional`):
```suggestion
    to: Optional[str] = None,
```
Unsure for Python 3.10 but I think you can even go with:

```
    to: str | None = None,
```

But I personally prefer the `Optional` wrapper type in this case instead of the syntactic sugar version, might be preference, but definitely favored over `Union` with a `None` type though.

app/delivery/send_to_providers.py
```diff
@@ -336,16 +339,55 @@ def update_notification_to_sending(notification, provider):
...
+    international: bool = False,
+    sender: Union[str, None] = None,
```
jimleroyer, line 347: Same here with `Optional` or `|` for None.

tests/app/delivery/test_send_to_providers.py
```diff
@@ -51,6 +51,53 @@
...
+            {
+                "AWS_PINPOINT_SC_POOL_ID": "sc_pool_id",
+                "AWS_PINPOINT_DEFAULT_POOL_ID": "default_pool_id",
+            },
+        ):
+            provider = send_to_providers.provider_to_use("sms", "1234", "+17065551234")
+        assert provider.name == "sns"
+
+    @pytest.mark.parametrize("sc_pool_id, default_pool_id", [(None, "default_pool_id"), ("sc_pool_id", None)])
+    def test_should_use_sns_if_pinpoint_not_configured(self, restore_provider_details, notify_api, sc_pool_id, default_pool_id):
```
jimleroyer, line 89: Love it

tests/app/delivery/test_send_to_providers.py
```diff
@@ -51,6 +51,53 @@
 from tests.conftest import set_config_values
 
 
+class TestProviderToUse:
```
jimleroyer, line 54: Nice set of tests!

app/clients/sms/aws_sns.py
sastels: Ha, I didn't find it with pylint but by accidentally triggering the error locally (I was trying to send to a USA number without setting the `AWS_US_TOLL_FREE_NUMBER` variable)

app/config.py
sastels: yeah exactly. A `None` value will break the Pinpoint send, so right now we route to SNS. In the future we may want to throw an error instead (ie when we insist on Pinooint for everything). Right now it allows production to ignore pinpoint entirely since we don't have it set there yet.

app/delivery/send_to_providers.py
sastels: I  figured I had added enough new logic to this function that some documentation was useful 😬

app/clients/sms/aws_pinpoint.py
sastels: if the  `OriginationIdentity` is `None` in the boto call then the call returns an error
```
Invalid type for parameter OriginationIdentity, value: None, type: <class 'NoneType'>, valid types: <class 'str'>
```
But as you mention below, we catch that case in our code and fall back to SNS.

tests/app/clients/test_aws_pinpoint.py
sastels: possibly, though this is a bit of a strange test - testing that if the notification has an invalid phone number in it then we get a log. Agree though that this is a strange thing that we'd like surfaced immediately though. Basically it means that we screwed up the phone number and AWS doesn't recognize it. So, for example, we start sending texts to a new country and our number formatter messes up. Yeah, I'll add a warning for these 👍

app/clients/sms/aws_sns.py
jimleroyer: Okay. I thought it might be pylint because the latter did raise it as a potential problem when I looked at this code locally in vscode.

tests/app/clients/test_aws_pinpoint.py
jimleroyer: I gave some more thoughts to this and one thing I wish we could do better  is OOP in our code base. We barely have much and this could be our focus on code reviews session. We rely a lot on environment variables which is not a good practice in general, as the code relies on environment rather than configuration of the logic to run. Encapsulating the pool config in an object would make this easier probably to run instead of relying on environments. Do you know what I mean here?

jimleroyer (approved):
LGTM! can't wait to have more of the pool config going forward!

tests/app/clients/test_aws_pinpoint.py
sastels: We do pull the env vars out of the env and put into the `Config` object, but we should make sure that we always read that object and never the env vars directly (which I think we're pretty good at?). I think that there's magic going on behind the scenes though (with flask, for example), where it takes vars directly from the env :/

In this case, though, in the test I made serial the app wasn't using the values of `Config` that I added in the test setup 😞 Which - how can that happen?

PR #2178, opened 2024-05-23
Fix fake callback

sastels:
# Summary | Résumé

The `INTERNAL_TEST_NUMBER` (which is used by the smoke test) was causing an error after we switched on Pinpoint: the fake callback was being processed by the sns callback task rather than the pinpoint callback task. This PR ensures that the correct task (ie based on the provider being used) is called.

* https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/360

To test locally:

Set env vars (from staging):
```
AWS_PINPOINT_DEFAULT_POOL_ID = "pool-8885654b47d6466f9aaee14c62494ff2"
```

Send an SMS to 16135550123. It should immediately register as sent. and there should be no celery errors.

None.


app/celery/research_mode_tasks.py
```diff
@@ -64,6 +74,22 @@ def aws_sns_callback(notification_id, to):
...
+    if to.strip().endswith(perm_fail):
+        return pinpoint_failed_callback(
+            "Phone is currently unreachable/unavailable", notification_id, destination=to, timestamp=timestamp
+        )
+    elif to.strip().endswith(temp_fail):
+        return pinpoint_failed_callback(
+            "Phone carrier is currently unreachable/unavailable", notification_id, destination=to, timestamp=timestamp
+        )
+    else:
+        return pinpoint_delivered_callback(notification_id, destination=to, timestamp=timestamp)
```
jimleroyer, line 90: It looks like only the last line is different, could we merge the logic on top into one common function, which would differ on that last function call?

tests/app/celery/test_research_mode_tasks.py
```diff
@@ -50,6 +52,30 @@ def test_make_sns_success_callback(notify_api, mocker, phone_number, sns_callbac
...
+    some_ref = str(uuid.uuid4())
+    now = datetime.now()
+    timestamp = now.strftime("%Y-%m-%d %H:%M:%S.%f")[:-3]
+
+    send_sms_response("pinpoint", phone_number, some_ref)
+
+    mock_task.apply_async.assert_called_once_with(ANY, queue=QueueNames.RESEARCH_MODE)
+    message_celery = mock_task.apply_async.call_args[0][0][0]
+    pinpoint_callback_args.update({"reference": some_ref, "destination": phone_number, "timestamp": timestamp})
+    assert message_celery == pinpoint_callback(**pinpoint_callback_args)
```
jimleroyer, line 76: 🥳

app/celery/research_mode_tasks.py
sastels: 🤔 instead of
```
def aws_pinpoint_callback(notification_id, to):
    now = datetime.now()
    timestamp = now.strftime("%Y-%m-%d %H:%M:%S.%f")[:-3]

    if to.strip().endswith(perm_fail):
        return pinpoint_failed_callback(
            "Phone is currently unreachable/unavailable", notification_id, destination=to, timestamp=timestamp
        )
    elif to.strip().endswith(temp_fail):
        return pinpoint_failed_callback(
            "Phone carrier is currently unreachable/unavailable", notification_id, destination=to, timestamp=timestamp
        )
    else:
        return pinpoint_delivered_callback(notification_id, destination=to, timestamp=timestamp)
```
What about 
```
def aws_pinpoint_callback(notification_id, to):
    now = datetime.now()
    timestamp = now.strftime("%Y-%m-%d %H:%M:%S.%f")[:-3]
    using_test_perm_fail_number = to.strip().endswith(perm_fail)
    using_test_temp_fail_number = to.strip().endswith(temp_fail)

    if using_test_perm_fail_number or using_test_temp_fail_number:
        return pinpoint_failed_callback(
            "Phone is currently unreachable/unavailable" if using_test_perm_fail_number else "Phone carrier is currently unreachable/unavailable", 
            notification_id, 
            destination=to, 
            timestamp=timestamp
        )
    else:
        return pinpoint_delivered_callback(notification_id, destination=to, timestamp=timestamp)
```

jimleroyer: Let's ship it as is for now. I think what I was thinking of is a proper abstraction that would send the callback based on the implementation, which would simplify the call sites code, but would likely repeat in the abstractions (potentially objects and with some template patterns).

jimleroyer (approved):
:shipit: sir

PR #2185, opened 2024-05-31
Task/remove letters

jzbahrai:
# Summary | Résumé

Remove letter code


app/letters/rest.py
```diff
@@ -10,6 +8,4 @@
 
 @letter_job.route("/letters/returned", methods=["POST"])
 def create_process_returned_letters_job():
-    references = validate(request.get_json(), letter_references)
-
-    return jsonify(references=references["references"]), 200
+    pass
```
andrewleith, line 11: Why not get rid of all of this, would that cause other issues?

app/letters/utils.py
andrewleith, line 1: Can we not delete this entire file?

app/service/rest.py
```diff
@@ -793,8 +773,7 @@ def create_one_off_notification(service_id):
 
 @service_blueprint.route("/<uuid:service_id>/send-pdf-letter", methods=["POST"])
 def create_pdf_letter(service_id):
-    resp = send_pdf_letter_notification(service_id, request.get_json())
-    return jsonify(resp), 201
+    pass
```
andrewleith, line 776: Are these routes reachable by end-users? If so, what kind of response is received after this change?

app/letters/utils.py
jzbahrai: we could! Lets do this incrementally though. I removed a bunch, will keep removing more cause its a lot of finding is something is called elsewhere or not.

app/service/rest.py
jzbahrai: I think they are currently in admin, though not visible. will remove them all in another pr

andrewleith (approved):
LGTM 🪓

PR #2191, opened 2024-06-05
use empty string for empty pool id

sastels:
# Summary | Résumé

We have
```
os.getenv("AWS_PINPOINT_DEFAULT_POOL_ID", None)
```
However if in our env var file we have
```
AWS_PINPOINT_DEFAULT_POOL_ID=
```
Then the pool id will be set to the empty string which will cause an error later on. Here we use the empty string instead of `None` as a default value.

* https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/360

Set
```
AWS_PINPOINT_SC_TEMPLATE_IDS = <one of your templates>
AWS_PINPOINT_DEFAULT_POOL_ID=
```
Send an sms with the template listed and one with a different template. All SMS should be delivered with no errors. the one from the sc template should be sent with Pinpoint.

Now set
```
AWS_PINPOINT_SC_POOL_ID=
```
and send with the sc template. The sms should send with no errors from SNS.

None.


app/delivery/send_to_providers.py
```diff
@@ -379,10 +379,9 @@ def provider_to_use(
     do_not_use_pinpoint = (
         has_dedicated_number
         or sending_to_us_number
-        or current_app.config["AWS_PINPOINT_SC_POOL_ID"] is None
-        or (current_app.config["AWS_PINPOINT_DEFAULT_POOL_ID"] is None and not using_sc_pool_template)
+        or not current_app.config["AWS_PINPOINT_SC_POOL_ID"]
```
ben851, line 382: This is my lack of experience with Python - so forgive me - just want to confirm my understanding. This will render as false if AWS_PINPOINT_SC_POOL_ID is an empty string in the config? I would have thought it would only do that if null (which I guess is what you thought originally too hence the "none" from before?)

sastels: yeah exactly, this is the "pythonic" way to check for an empty string
```
>>> not ""
True
>>> not "hello"
False
>>> "" is None
False
```

ben851 (approved):
LGTM

PR #2193, opened 2024-06-11
Add Template_Categories table

whabanks:
# Summary | Résumé
This PR implements the backend work for Template Categories per the [Short code ADR](https://github.com/cds-snc/notification-adr/pull/69/files#diff-828715828f8c151d04893b37ee67484650d9b914d2e1b0abc815422e8b617d74).

## Change Details
- Add the `template_categories` table to the DB in a migration
- Add the `TemplateCategory` model and associations with `Template` and `TemplateHistory`
- Implement dao CRUD methods to interact with the table
- New Admin API endpoints under `/template-category` 
    - Create Category - `/template-category`
    - Get all Categories - `/template-category`
    - Get Category - `/template-category/<template_category_id>`
    - Get Category by template id - `/template-category/by-template-id/<template-id>`
    - Update Category - `/template-category/<template_category_id>`
    - Delete Category - `/template-category/<template_category_id>?cascade=[True | False]`
- New Admin API endpoint to associate a template with a category: `/service/<id>/template/<template_id>/category/<template_category_id>`

- https://github.com/cds-snc/notification-adr/pull/69
- https://github.com/cds-snc/notification-admin/pull/1869

If you're using Postman, [you can import this request collection](https://github.com/user-attachments/files/15975957/Test.Template.Categories.postman_collection.json) to hit each new endpoint. I've included example payloads for create and update

Default generic Template Category id's you can use in the requests
```
    DEFAULT_TEMPLATE_CATEGORY_LOW = "0dda24c2-982a-4f44-9749-0e38b2607e89"
    DEFAULT_TEMPLATE_CATEGORY_MEDIUM = "f75d6706-21b7-437e-b93a-2c0ab771e28e"
    DEFAULT_TEMPLATE_CATEGORY_HIGH = "c4f87d7c-a55b-4c0f-91fe-e56c65bb1871"
```

- [x]  Basic Create and Update
    - [x] Create - `/template-category`
    - [x] Update - `/template-category/<template_category_id>`

## Getting a Category
- [x] Get a single category - `/template-category/<template_category_id>`
- [x] Associate template with a Category - `/service/<service_id>/template/<template_id>/category/<template_category_id>`
    -  Associate a template with one of the generic categories listed above
- [x] Get category by template id - `/template-category/by-template-id/<template-id>`
    - Get the category by template id with the template you used in the above test

## Getting all Categories
You'll need to create some new categories to fully test the gets with filtering.
* One `hidden` category - associate it with an SMS template
* One  `not hidden` category - associate it with an email template

- [x] Get all categories - `/template-category`
- [x] Get all categories filter by hidden `/template-category?hidden=[True | False | None]`
- [x] Get all categories filter by template_type`/template-category?template_type=[sms | email]`
- [x] Get all categories filter by hidden & template type `/template-category?hidden=[True | False | None]&template_type=[sms | email]`

## Deleting Categories
- [x] **Delete category - No cascade - Category not associated with any templates** 
    1. First, add a new category
    2. Delete the category with `/template-category/template_category_id?cascade=False`
    4. Note that the delete was successful
- [x] **Delete Category - No cascade - Category associated with a template**
    1. Get the `category_id` from the earlier test where you associated it with a template
    2. Delete the category with `/template-category/<template_category_id>?cascade=False` making sure `cascade=False` or omitted entirely
    4. Note that you are prevented from deleting the category due to the association with a template
- [x] **Delete Category - Cascade - Category associated with a template**
    - Set `cascade=True` in the query string and try deleting the same category again
    - Note that the category was deleted, and the template was associated with the Generic Default Low category


app/models.py
```diff
@@ -1198,6 +1217,14 @@ def get_link(self):
             _external=True,
         )
 
+    @property
+    def template_process_type(self):
+        if self.template_type == SMS_TYPE and self.template_categories.sms_process_type:
```
jzbahrai, line 1222: I think this needs to be self.category.sms_process_type IF self.process_type is empty? 
Can we add some tests to this

migrations/versions/0454_add_template_category.py
```diff
@@ -0,0 +1,76 @@
...
+    )
+    op.create_foreign_key(
+        "fk_template_template_categories",
+        "templates",
+        "template_categories",
+        ["template_category_id"],
+        ["id"]
+    )
+
+    # Insert the generic Low priority (bulk) category
```
jzbahrai, line 63: lets add this! but not just got bulk, we want to do "default high", med, low. Also lets store the IDs in the config as they are magic values (or if not confg, then defs on top of this file as globals)

app/models.py
whabanks: Yea you're right, I had the logic correct the first time and then changed it 🤦

app/dao/template_categories_dao.py
```diff
@@ -0,0 +1,30 @@
...
+def dao_update_template_category(template_category: TemplateCategory):
+    db.session.add(template_category)
+
+
+def dao_get_template_category_by_id(template_category_id):
+    return TemplateCategory.query.filter_by(id=template_category_id).one()
+
+
+def dao_get_template_category_by_template_id(template_id):
+    return TemplateCategory.query.join(TemplateCategory.templates).filter_by(id=template_id).one()
```
jzbahrai, line 26: I think you should do this the other way around.. cause the join on template <> template categories will be more exp then getting the Template filtered by id. Also you can add TemplateCategory (not just id) to the Template model, so Template.TemplateCategory should give you the TemplateCategory for that. (use sqlalchemy Relationship)

app/dao/template_categories_dao.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def dao_get_all_template_categories():
```
jzbahrai, line 29: lets write this with filters - so if we only want to return sms/ email/ hidden

app/dao/templates_dao.py
```diff
@@ -78,6 +78,40 @@ def dao_update_template_reply_to(template_id, reply_to):
     return template
 
 
+@transactional
+def dao_update_template_category(template_id, category_id):
+    Template.query.filter_by(id=template_id).update(
```
jzbahrai, line 83: I didn't even think so updating TemplateHistory! food catch

app/models.py
```diff
@@ -1179,6 +1213,7 @@ class Template(TemplateBase):
 
     service = db.relationship("Service", backref="templates")
     version = db.Column(db.Integer, default=0, nullable=False)
+    category = db.relationship("TemplateCategory", backref="templates")
```
jzbahrai, line 1216: I would change this to template_category. Ignore my relationship comment above, this should work!

app/models.py
```diff
@@ -1198,6 +1233,14 @@ def get_link(self):
             _external=True,
         )
 
+    @property
+    def template_process_type(self):
+        if self.template_type == SMS_TYPE:
```
jzbahrai, line 1238: can we add a doc string explaining why we do this. By default we use process_type in TemplateCategory - unless overridden through the admins for a particular template ( we then use the process_type on the Template table)

app/dao/template_categories_dao.py
whabanks: For the SMS/Email filter, are we filtering by template in this case? e.g. `Select all template categories used by at least 1 sms template`

jzbahrai: I am unsure? I was going to say select all SMS template categories / or select template categories where hidden is not true. something along those lines

whabanks: That's what I had in mind as well, I've implemented filtering so that we can select categories by any combination of `template_type` and `hidden` e.g:

Categories associated with sms templates that are hidden / not hidden
Categories associated with email templates that are hidden / not hidden

I added a lazy join on the category<>template so we can also easily get a list of the templates filtered this way as well if we need to.

whabanks:
Question: Since [`Template.process_type`](https://github.com/cds-snc/notification-api/blob/33006d7279122dbaa839ea353bad26d1b748f5df/app/models.py#L1129) acts as and override to the process_type from the category should we changes this column to be nullable?

app/dao/template_categories_dao.py
```diff
@@ -0,0 +1,91 @@
...
+
+from flask import current_app
+
+from app import db
+from app.dao.dao_utils import transactional
+from app.models import Template, TemplateCategory
+
+
+@transactional
+def dao_create_template_category(template_category: TemplateCategory):
```
jzbahrai, line 11: do we need validation here? If the fields that are required like name/ etc are empty - then we should bubble the error up? or we could do in the API

whabanks: This is covered by the [schema class](https://github.com/cds-snc/notification-api/pull/2193/files#diff-743c8db0af830ae358f828cadf2faa7caaa075029cdce9bd4650ed51ca80038f) when we [call `load` on the request json](https://github.com/cds-snc/notification-api/pull/2193/files#diff-0ad60abb8684d413a0a5c260f18655d434359f538388c5d8e0f5affc3db8624cR28) it will raise a `ValidationError` when the values are invalid.

app/dao/template_categories_dao.py
```diff
@@ -0,0 +1,87 @@
...
+    templates = Template.query.filter_by(template_category_id=template_category_id).all()
+
+    if not templates or cascade:
+        # When there are templates and we are cascading, we set the category to a default
+        # that matches the template's previous category's priority
+        if cascade:
+            for template in templates:
+                # Get the a default category that matches the previous priority of the template, based on template type
+                default_category_id = _get_default_category_id(
+                    template_category.sms_process_type
```
jzbahrai, line 70: so I am hesitant about this line, as we are going to have template.process_type be empty as a default..so there won't be a previous template. Can we just set the category to DefaultLow?

app/dao/templates_dao.py
```diff
@@ -78,6 +78,71 @@ def dao_update_template_reply_to(template_id, reply_to):
...
+            "process_type": template.process_type,
+            "service_letter_contact_id": template.service_letter_contact_id,
+        }
+    )
+    db.session.add(history)
+    return template
+
+
+@transactional
+def dao_update_template_category(template_id, category_id):
```
jzbahrai, line 113: can we say template_category_id here? Just cause I thought we were updating TemplateCategory here

tests/app/conftest.py
```diff
@@ -241,6 +335,7 @@ def create_sample_template(
     subject_line="Subject",
     user=None,
     service=None,
+    category=None,
```
jzbahrai, line 338: Do we need template_category_id here instead of category?

tests/app/dao/test_template_categories_dao.py
```diff
@@ -0,0 +1,368 @@
...
+        "hidden": False,
+    }
+
+    template_category = TemplateCategory(**data)
+    dao_create_template_category(template_category)
+
+    assert dao_get_template_category_by_id(template_category.id) == template_category
+
+
+@pytest.mark.parametrize(
```
whabanks, line 127: Originally I wanted to cut down the size of these parameters by creating the `categories_to_insert` in the test body and modifying the `hidden` field per test. When I did it this way, then between the tests the categories were not being removed from the test DB and the unique constraint on the `name_en` and `name_fr` was causing the tests to fail. 

I can leave it as is or do something like set the name to a `str(uuid)` to get around the unique constraint and shrink the size of this behemoth.

andrewleith (commented):
Looks good to me in general, I'll wait for Jumana for the sign off.  I just approved your WAF update to fix the waffles issue - if you merge that one your tests should pass here I hope!

tests/app/conftest.py
whabanks: Just setting the `category_id` should work fine. What we could also do is if the category is `None` we can set the `category_id` to one of the default categories.

jzbahrai (commented):
Excellent Job! I liked the way the delete worked and thinking of the cascade usecases. 

1 requirement:
When doing the delete - we update the Templates table. When I checked the DB - the updated_at didn't change. We should fix that


1 suggestion (please make a ticket for this based on what you think)
- Add a created_at/ updated_at on the TemplateCategory table (this should also update with the API accordingly)

migrations/versions/0454_add_template_category.py
```diff
@@ -27,23 +27,25 @@ def upgrade():
         sa.Column("sms_process_type", sa.String(length=255), nullable=False),
         sa.Column("email_process_type", sa.String(length=255), nullable=False),
         sa.Column("hidden", sa.Boolean(), nullable=False),
+        sa.Column("created_at", sa.DateTime(), nullable=False),
```
jzbahrai, line 30: these can be default to datetime now

jzbahrai (approved):

PR #2233, opened 2024-07-25
Fix TC deletion bug

whabanks:
# Summary | Résumé
This PR fixes an issue where a Template Category could not be deleted following deletion of the templates it was associated with. 

When a user "deletes" a template the template is marked as `archived` rather than deleted and removed from the templates table, preventing deletion.

`dao_delete_template_category_by_id` now checks if the category is ONLY associated with archived templates. If that is the case, we set the archived template's `template_category_id` to `null` and proceed with deleting the template category. 

A side effect of this is that the `templates_history` table will contain rows with references to categories that no longer exist and cannot ever be cross referenced as we do not have a `template_category_history` table.

This issue was uncovered during the template categories bug bash and is documented below
https://docs.google.com/document/d/1swqV2sKYBZB9UHKjXsspPSp6SEJSBWVxmbAcUKFu5Ts/edit#task=OuE8AO3ICha-9aYF

1. Create a new template category
2. Associate the category with 1 template
3. Delete the template
4. Delete the template category
5. The delete should succeed
6. Verify in the DB that the archived template's `template_category_id` is set to `null`

### DIFF
--- a/app/dao/template_categories_dao.py
+++ b/app/dao/template_categories_dao.py
@@ -59,23 +59,37 @@ def dao_delete_template_category_by_id(template_category_id, cascade=False):
     template_category = dao_get_template_category_by_id(template_category_id)
     templates = Template.query.filter_by(template_category_id=template_category_id).all()
 
-    if templates and not cascade:
-        raise InvalidRequest(
-            "Cannot delete categories associated with templates. Dissociate the category from templates first.", 400
-        )
+    if templates:
+        if not cascade:
+            # When a template is "deleted" it is marked as archived, thus the association with the catgory
+            # remains. If the category is only associated with archived templates, we can delete it after
+            # setting the template's category to None. The template's original category is preserved in the
+            # template_history table.
+            if all(template.archived for template in templates):
+                for template in templates:
+                    template.template_category_id = None
+                    template.updated_at = datetime.utcnow()
+                    db.session.add(template)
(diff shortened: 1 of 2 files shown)


sastels:
Should we also null out the `template_category_id` in the affected rows of `templates_history`?

whabanks:
> Should we also null out the `template_category_id` in the affected rows of `templates_history`? Seems bad to leave the database in a state where looking up things might fail.

Agreed, we can null it as you suggested or if we want to preserve the priority / `process_type` in history we could:

2. Set `template_category_id` in history to one of the generic categories that maps to the `process_type` of the deleted category
3. Null the `template_category_id` in the history table but set the `process_type` to the one corresponding to the deleted category. (This might be misleading as `process_type` is intended as an override now)

jzbahrai (approved):

jzbahrai (changes requested):
I am wondering if we should have an archived flag.. instead. You can't delete a category - it just goes into archived on the template category table. I am not a fan of losing history

whabanks:
> I am wondering if we should have an archived flag.. instead. You can't delete a category - it just goes into archived on the template category table. I am not a fan of losing history

Neat idea, it would require some more front end work but I can certainly get the back end ready with this PR.

jzbahrai:
We are going to shut this PR, to be revisited when we add the archiving functionality

PR #2255, opened 2024-08-14
We want to save the reply_to_email_addresses in the cache

jzbahrai:
# Summary | Résumé

We are not currently storing the reply_to_email_addresses which is a field on the Service model, to show up when we retrieve the Service from the cache. This meant that the reply_to_address was empty for the end user which is incorrect

Two parts:
1. The service gets added to the cache when the admin gets loaded - we needed to change the model Schema so we send the reply_to_email_addresses to the admin
2. We need to change the model to_json, so we might appropriately set the reply_to_email_addresses as a ServiceReplytoEmail model

# Testing:
[ ] Set the reply_to_email_address
[ ] Stay on the main branch: Send an email using the csv method  - it will send an email WITHOUT the reply-to set
[ ] Switch to this branch: Send an email using the csv method - it will send an email WITH the reply-to set

Example:
Screenshot 2024-08-14 at 4 24 24 PM

* https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/1633

- [x] This PR does not break existing functionality.
- [x] This PR does not violate GCNotify's privacy policies.
- [x] This PR does not raise new security concerns. Refer to our GC Notify Risk Register document on our Google drive.
- [x] This PR does not significantly alter performance.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -608,8 +608,16 @@ class Service(BaseModel, Versioned):
         fields.pop("letter_contact_block", None)
         fields.pop("email_branding", None)
         fields["sms_daily_limit"] = fields.get("sms_daily_limit", 100)
+        reply_to_addresses = fields.get("reply_to_email_addresses", None)
+        fields.pop("reply_to_email_addresses", None)
+        current_service = cls(**fields)
+        # If reply_to_addresses were in the JSON, add them to the service
+        if reply_to_addresses:
+            current_service.reply_to_email_addresses = [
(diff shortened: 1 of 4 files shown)


whabanks (approved):
LGTM! Tested without this branch and noted no `reply_to_text` in the notifications table when sending via CSV. Switched to this branch, cleared my service cache, re-sent, and noted the `reply_to_text` was present in the DB.

Nice work tracking this one down!

PR #2269, opened 2024-08-22
Skip subsegment creation if segment is missing in current thread context

jimleroyer:
# Summary | Résumé

There is a possible situation when a Celery periodic task starts itself and has no segment in the current thread context. This will lead to errors as a subsegment will be created without a parent segment, which is required.

We can opt to skip the subsegment creation as this would normally be applied in a context of a parent task that would call down the current task. If there is no current segment, skip the creation of the subsegment and let the prerun signal handler to create the segment later down the road.

The same applies on closing the subsegment on the `after_task_publish` signal, i.e. check if a current subsegment exists before closing and if not, log a warning along with the task details to know on which one that happened, which should be a periodic tasks.

Also, I added missing parameters on the x-ray signal handler functions and log these to inspect what really is in there and which options we could open up with these.

* https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/400

Deploy to staging and

1. Make sure all the pods redeploy normally.
2. Query CloudWatch with the following query to make sure that there are no errors anymore:
```
| filter kubernetes.container_name like /^celery/
| filter @message like /Failed to create a X-Ray subsegment on task publish/
| sort @timestamp desc
| limit 20
```
3. Target a task ID from one of the beat workers using this query...
```
| filter kubernetes.container_name like /^celery/
| filter @message like /Scheduler: Sending due task beat-inbox/
| sort @timestamp desc
| limit 20
```
...and make sure that the `before_task_publish` signals do not execute for the periodic tasks, and that the `task_prerun` signals execute properly for the periodic tasks.
4. Look at the CloudWatch X-Ray trace map in the AWS console and verify that the Celery tasks are present for both regular non-periodic tasks and periodic ones.

Make sure that all pods deploy correctly. There is initialization code that did prevent the Celery and API pods to successfully launch in previous X-Ray SDK extension fixes.

### DIFF
--- a/app/aws/xray_celery_handlers.py
+++ b/app/aws/xray_celery_handlers.py
@@ -17,8 +17,10 @@ logger = logging.getLogger("celery_aws_xray_sdk_extension")
 CELERY_NAMESPACE = "celery"
 
 
-def xray_before_task_publish(sender=None, headers=None, **kwargs):
-    logger.info(f"xray-celery: before publish: sender: {sender} headers: {headers}")
+def xray_before_task_publish(
+    sender=None, headers=None, exchange=None, routing_key=None, properties=None, declare=None, retry_policy=None, **kwargs
+):
+    logger.info(f"xray-celery: before publish: sender={sender}, headers={headers}, kwargs={kwargs}")
     headers = headers if headers else {}
     task_id = headers.get("id")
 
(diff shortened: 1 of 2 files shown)


P0NDER0SA (approved):
Ok - let's give it a try

PR #2275, opened 2024-08-28
Add api to suspend a callback

jzbahrai:
# Summary | Résumé

We want an API that will set the suspension to false for a given callback

* https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/1648

You can test it with the admin PR 

None.

### DIFF
--- a/app/dao/service_callback_api_dao.py
+++ b/app/dao/service_callback_api_dao.py
@@ -73,6 +73,11 @@ def reset_service_callback_api(service_callback_api, updated_by_id, url=None, be
     db.session.add(service_callback_api)
 
 
+def get_service_callback_api_with_service_id(service_id):
+    # There is ONLY one callback configured per service
+    return ServiceCallbackApi.query.filter_by(service_id=service_id).all()
+
+
 def get_service_callback_api(service_callback_api_id, service_id):
     return ServiceCallbackApi.query.filter_by(id=service_callback_api_id, service_id=service_id).first()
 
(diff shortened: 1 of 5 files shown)


app/dao/service_callback_api_dao.py
```diff
@@ -92,8 +97,8 @@ def delete_service_callback_api(service_callback_api):
 
 @transactional
 @version_class(ServiceCallbackApi)
-def suspend_service_callback_api(service_callback_api, updated_by_id):
-    service_callback_api.is_suspended = True
+def suspend_unsuspend_service_callback_api(service_callback_api, updated_by_id, suspend=False):
+    service_callback_api.is_suspended = suspend
```
whabanks, line 101: Since we have a `service_callback_api_history` table, when we un-suspend callbacks we should also `null` the `suspended_at` date column. If the date is present while `is_suspended` is False it could be confusing.

jzbahrai: hmm, I think a False with the datetime would show when the callback went live again? What do you think?

whabanks: Yea it would show that and will be useful later down the road for automation. The `updated_at` in the history table also shows that.

image

We should document somewhere, that `suspended_at` can also represent a time of resumption.

whabanks (approved):

app/dao/service_callback_api_dao.py
jzbahrai: added in a comment on models.py

PR #2293, opened 2024-09-19
Fix Notification History to have the full day of data

jzbahrai:
# Summary | Résumé

We were deleting data from the notifications table for partial days. We want to delete the data for the entire day. This PR changes our datetime handling to fix that


andrewleith (approved):
LGTM - did we want to add any additional logging?

PR #2298, opened 2024-09-23
Freshdesk: fix template link

jzbahrai:
# Summary | Résumé

Fix the template link as we were sending https:// twice
Screenshot 2024-09-23 at 3 58 17 PM


whabanks (approved):
Nice simple catch LGTM

PR #2302, opened 2024-09-25
Update the fetch for notifications to take the correct datetimes into account

jzbahrai:
# Summary | Résumé
- We want query the facts table for 6 days prev to the current day
- We want to query notifications for todays data. due to timezones/ and when the job runs to populate facts, we will take the max existing date in facts from the last 6 days and use that as a starting point to query notifications for
- We added a filter for notification_type as there is a part in the frontend that fetches this query only for a singular notification type

* https://github.com/cds-snc/notification-planning/issues/1651

## Tests:
Updated and added unit tests

### DIFF
--- a/app/dao/fact_notification_status_dao.py
+++ b/app/dao/fact_notification_status_dao.py
@@ -1,4 +1,4 @@
-from datetime import datetime, time, timedelta
+from datetime import datetime, time, timedelta, timezone
 
 from flask import current_app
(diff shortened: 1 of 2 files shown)


andrewleith (approved):
This looks good to me.  I like the way you've refactored things, it is much easier to follow 🏅 

I committed 2 extra tests to help cover more scenarios!

PR #2320, opened 2024-10-08
Use dryrun for sms to INTERNAL_TEST_NUMBER

sastels:
# Summary | Résumé

Instead of faking a boto call for SMS to our `INTERNAL_TEST_NUMBER` we can do a real boto call with the parameter `DryRun` set to `True`. This should give us better timing estimates and still not cost anything (Dryrun does not actually send the message or charge us).

## Related Issues | Cartes liées

* https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/426

# Test instructions | Instructions pour tester la modification

Send to +16135550123. See in the logs `Dry run enabled for SMS to +16135550123`...
 
# Release Instructions | Instructions pour le déploiement

None.

# Reviewer checklist | Liste de vérification du réviseur

- [ ] This PR does not break existing functionality.
- [ ] This PR does not violate GCNotify's privacy policies.
- [ ] This PR does not raise new security concerns. Refer to our GC Notify Risk Register document on our Google drive.
- [ ] This PR does not significantly alter performance.
- [ ] Additional required documentation resulting of these changes is covered (such as the README, setup instructions, a related ADR or the technical documentation).

> ⚠ If boxes cannot be checked off before merging the PR, they should be moved to the "Release Instructions" section with appropriate steps required to verify before release. For example, changes to celery code may require tests on staging to verify that performance has not been affected.

PR #2322, opened 2024-10-09
Revert "Use dryrun for sms to INTERNAL_TEST_NUMBER"

sastels:
Reverts cds-snc/notification-api#2320

broke the smoke test because there's no fake delivery receipt anymore :/ I'll add it back...


sastels:
fixed in #2323

PR #2324, opened 2024-10-10
fix: use CDS Trivy vulnerability database

patheard:
# Summary
Update the Docker scan actions to use a self-hosted Trivy vulnerability database. This is being done to address the rate limiting of the publicly hosted database.

# Related
- https://github.com/cds-snc/platform-core-services/issues/597


andrewleith (approved):
Thank you Pat!! ❤️

PR #2329, opened 2024-10-21
Add email/sms annual limit columns to services table

whabanks:
# Summary | Résumé

This PR adds two new columns to track a service's annual limits:
- `email_annual_limit`
- `sms_annual_limit`

_TODO: Fill in test instructions for the reviewer._

None.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -519,6 +519,8 @@ class Config(object):
         "job": "{}-dvla-file-per-job".format(os.getenv("NOTIFY_ENVIRONMENT", "development")),
         "notification": "{}-dvla-letter-api-files".format(os.getenv("NOTIFY_ENVIRONMENT", "development")),
     }
+    SERVICE_ANNUAL_EMAIL_LIMIT = env.int("SERVICE_ANNUAL_EMAIL_LIMIT", 10_000_000)
+    SERVICE_ANNUAL_SMS_LIMIT = env.int("SERVICE_ANNUAL_SMS_LIMIT", 25_000)
 
     FREE_SMS_TIER_FRAGMENT_COUNT = 250000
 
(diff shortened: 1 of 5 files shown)


app/config.py
```diff
@@ -519,6 +519,8 @@ class Config(object):
         "job": "{}-dvla-file-per-job".format(os.getenv("NOTIFY_ENVIRONMENT", "development")),
         "notification": "{}-dvla-letter-api-files".format(os.getenv("NOTIFY_ENVIRONMENT", "development")),
     }
+    SERVICE_ANNUAL_EMAIL_LIMIT = env.int("SERVICE_ANNUAL_EMAIL_LIMIT", 10_000_000)
+    SERVICE_ANNUAL_SMS_LIMIT = env.int("SERVICE_ANNUAL_SMS_LIMIT", 25_000)
```
andrewleith, line 523: Where do we define the daily limits? I expected to see them somewhere here too...

app/models.py
```diff
@@ -65,6 +65,8 @@
 DELIVERY_STATUS_CALLBACK_TYPE = "delivery_status"
 COMPLAINT_CALLBACK_TYPE = "complaint"
 SERVICE_CALLBACK_TYPES = [DELIVERY_STATUS_CALLBACK_TYPE, COMPLAINT_CALLBACK_TYPE]
+DEFAULT_SMS_ANNUAL_LIMIT = 25000
+DEFAULT_EMAIL_ANNUAL_LIMIT = 10000000
```
andrewleith, line 69: Do we need these here AND in `config.py`?

andrewleith (commented):
Looks good, just a question on where we are storing the defaults!

app/models.py
whabanks: Sadly yes, as far as I know we aren't able to access the Flask app's environment from `models.py`.

```python
email_annual_limit = db.Column(db.BigInteger, nullable=False, default=current_app.config["SERVICE_ANNUAL_EMAIL_LIMIT"])
```
Results in:
```python
File "/home/vscode/.cache/pypoetry/virtualenvs/notification-api-PZspUUhw-py3.10/lib/python3.10/site-packages/werkzeug/local.py", line 519, in _get_current_object
    raise RuntimeError(unbound_message) from None
RuntimeError: Working outside of application context.
```

Technically we could do: 
```python
from app.config import Config
config = Config()
email_annual_limit = db.Column(db.BigInteger, nullable=False, default=config.SERVICE_ANNUAL_EMAIL_LIMIT)
```

But only for keys that aren't overridden in different environments. It could be useful to separate constant-like configs into their own class so we can reference them outside of the app context.

app/config.py
whabanks: We hard-coded defaults in the [front end config](https://github.com/cds-snc/notification-admin/blob/84e6be282c84c1b4d00972042f05d29432e13185/app/config.py#L92) for display in `service-settings`. I also expected to see this configured in API but did not.

andrewleith (approved):
LGTM

PR #2347, opened 2024-11-08
fix test

jzbahrai:
# Summary | Résumé

Test was not returning a result in order (it isn't required to). For testing purposes made the result a set.


andrewleith (approved):
✅

PR #2363, opened 2024-11-18
Fix/tests using called_once_with

sastels:
# Summary | Résumé

There's no such thing as `mock.called_once_with()` - pytest thinks this is a new mocked function and so lines such as
```
assert mock.called_once_with("sadness")
```
will always pass. 😢 

So there are a few tests that have been passing despite sometimes actually not testing things correctly.

There is a `mock.assert_called_once_with()` that we can use instead though. This PR mostly changes to those.

Discovered because Python 3.12 doesn't allow this craziness.

none

None.


tests/app/test_email_limit_utils.py
```diff
@@ -30,17 +31,9 @@ def test_fetch_todays_requested_email_count(self, client, mocker, sample_service
...
         increment_todays_email_count(sample_service.id, increment_by)
 
-        assert mocked_incrby.called_once_with(cache_key, increment_by)
-        if redis_value is None:
-            assert mocked_set.called_once_with(
-                cache_key,
-                db_value,
-            )
-        else:
-            mocked_set.assert_not_called()
```
sastels, line 46: We're mocking the code that involves the set (`fetch_todays_email_count()`) so the set isn't actually called. That's  ok since this code is tested above, so it's cleaner to mock it out in this test (we just can't test the `set` though)

tests/app/test_sms_fragment_utils.py
```diff
@@ -33,17 +34,9 @@ def test_fetch_todays_requested_sms_count(client, mocker, sample_service, redis_
...
     increment_todays_requested_sms_count(sample_service.id, increment_by)
 
-    assert mocked_incrby.called_once_with(cache_key, increment_by)
-    if redis_value is None:
-        assert mocked_set.called_once_with(
-            cache_key,
-            db_value,
-        )
-    else:
-        mocked_set.assert_not_called()
```
sastels, line 49: same comment as above.

tests/app/test_email_limit_utils.py
```diff
@@ -19,28 +19,21 @@ def test_fetch_todays_requested_email_count(self, client, mocker, sample_service
...
-
-        assert mocked_incrby.called_once_with(cache_key, increment_by)
-        if redis_value is None:
-            assert mocked_set.called_once_with(
-                cache_key,
-                db_value,
-            )
-        else:
-            mocked_set.assert_not_called()
+        with set_config(client.application, "REDIS_ENABLED", True):
```
sastels, line 37: If we don't set `REDIS_ENABLED` then `increment_todays_email_count()` immediately returns `None`, and it turns out that `REDIS_ENABLED` is false in CI by default.

jzbahrai (approved):
thanksfor finding and fixing this!

sastels:
oh dear I found more, we also have use lines like `assert mock.not_called()` instead of `mock.assert_not_called()` :/ Let me add these....

PR #2369, opened 2024-11-27
Updated migration

jzbahrai:
# Summary | Résumé

We needed some more french specific words for the template + for the numbers (due to french formatting). This PR adds that

### DIFF
--- a/migrations/versions/0468_edit_limit_emails.py
+++ b/migrations/versions/0468_edit_limit_emails.py
@@ -1,0 +1,220 @@
+"""empty message
+
+Revision ID: 0468_edit_limit_emails
+Revises: 0467_add_limit_emails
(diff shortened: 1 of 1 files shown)


migrations/versions/0468_edit_limit_emails.py
```diff
@@ -0,0 +1,220 @@
...
+    quarterly_annual_limit_template_id,
+]
+
+annual_limit_reached = "\n".join(
+    [
+        "(la version française suit)",
+        "",
+        "Hello ((name)),",
+        "",
+        "0 ((message_type)) remaining until April  ((fiscal_end)).",
```
whabanks, line 36: Should we have `message_type_[en|fr]` so we can differentiate between: `email(s)`:`courriel(s)`?

PR #2387, opened 2024-12-10
Use official 3.10 New Relic later and Debian slim

sastels:
# Summary | Résumé

Having issues with the official New Relic layers after 3.9 and alpine. Builds fine but cannot register with AWS Lambda.

Here we
- switch to the recommended Debian distribution (well, Debian-slim).
- download the official New Relic layer for 3.10 rather than using the one we made ourselves. (the download script has been changed to reflect this)

Surprisingly this debian based image is 20% smaller than the current alpine based image. 

We will test this in staging and if all is good we will revert it and see if the 3.12 upgrade works with Debian.

## Related Issues | Cartes liées

* https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/465

# Test instructions | Instructions pour tester la modification

This has been built and deployed to dev and can be tested there.
- go to the [dev lambda function](https://ca-central-1.console.aws.amazon.com/lambda/home?region=ca-central-1#/functions/api-lambda?tab=testing)
- under the "test" tab choose the template "API Gateway Http API"
- click the "test" button.
- the "executing function" box should eventually turn green. The details say a 404 was returned. 
- To get the status out of api, change the "rawPath" on line 4 to "/" and the "method" on line 48 to "GET". Testing this should result in a 200 and the api status line.
- run the test a bunch (10-20 times).
- after a few minutes your invocations should show up in the [New Relic dashboard](https://one.newrelic.com/nr1-core/lambda/overview/MjY5MTk3NHxJTkZSQXxOQXwyMzIzMjU3MTc4NTM2MzgwMzI4?account=2691974&duration=1800000&filters=%28domain%20%3D%20%27INFRA%27%20AND%20type%20%3D%20%27AWSLAMBDAFUNCTION%27%29&state=c3492efc-5a7e-1f4a-9150-7399c671d2b7)

# Release Instructions | Instructions pour le déploiement

None.

# Reviewer checklist | Liste de vérification du réviseur

- [x] This PR does not break existing functionality.
- [x] This PR does not violate GCNotify's privacy policies.
- [x] This PR does not raise new security concerns. Refer to our GC Notify Risk Register document on our Google drive.
- [x] This PR does not significantly alter performance.
- [ ] Additional required documentation resulting of these changes is covered (such as the README, setup instructions, a related ADR or the technical documentation).

[...]

PR #2394, opened 2024-12-13
Revert "Use official 3.10 New Relic later and Debian slim" if needed!!

sastels:
Reverts cds-snc/notification-api#2387 in case we need to.


jzbahrai (approved):

sastels:
don't need anymore

PR #2401, opened 2025-01-01
Task/fix stats

jzbahrai:
# Summary | Résumé

Switching off stats due to incident:
https://gcdigital.slack.com/archives/CNWA63606/p1735758715690619

PR #2402, opened 2025-01-02
Revert "Task/fix stats"

jzbahrai:
Reverts cds-snc/notification-api#2401

I believe this PR: https://github.com/cds-snc/notification-admin/pull/2029
will fix the issue that was being caused by sending the wrong year.


smcmurtry (approved):
👍

PR #2403, opened 2025-01-03
Task/fix schedule

jzbahrai:
# Summary | Résumé

This task requires a process_day to run - we should have a default when one is not provided.


whabanks (approved):
LGTM!

PR #2419, opened 2025-01-20
tweak INTERNAL_TEST_NUMBER sending

sastels:
# Summary | Résumé

When we send to the INTERNAL_TEST_NUMBER we were queueing the fake callback before setting the reference in the notification. This could lead to the callback failing on its first attempt.

Here we set the reference before creating the fake callback.

Test on staging - make sure SMS to this number get marked delivered.

None.

### DIFF
--- a/app/delivery/send_to_providers.py
+++ b/app/delivery/send_to_providers.py
@@ -20,7 +20,13 @@ from unidecode import unidecode
 from urllib3 import PoolManager
 from urllib3.util import Retry
 
-from app import bounce_rate_client, clients, document_download_client, statsd_client
+from app import (
+    bounce_rate_client,
+    clients,
+    create_uuid,
+    document_download_client,
+    statsd_client,
+)
 from app.celery.research_mode_tasks import send_email_response, send_sms_response
 from app.clients.sms import SmsSendingVehicles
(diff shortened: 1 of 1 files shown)


tests/app/delivery/test_send_to_providers.py
```diff
@@ -490,29 +488,19 @@ def test_should_call_send_sms_response_task_if_research_mode(
...
     assert persisted_notification.to == sample_notification.to
     assert persisted_notification.template_id == sample_notification.template_id
     assert persisted_notification.status == "sent"
     assert persisted_notification.sent_at <= datetime.utcnow()
     assert persisted_notification.sent_by == "sns"
-    assert persisted_notification.reference == reference
     assert not persisted_notification.personalisation
 
 
-def test_should_not_have_sent_status_if_fake_callback_function_fails(sample_notification, mocker):
```
sastels, line 505: This test was for the corner case where a user would send with a TEST key and somehow our fake callback code would fail (so the sms would stay CREATED and get resent). We're explicitly marking it SENT first now to make the callbacks more likely to pass the first time so this test no longer works.

ben851 (approved):
I think so?

PR #2430, opened 2025-01-28
Added error handling for pinpoint errors

jzbahrai:
# Summary | Résumé

If Pinpoint is throwing an error due to a ConflictError or ValidationError - we do not want to retry the notification. We will store the error so it might be propagated to the frontend in the future.

https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/1760


whabanks (approved):
Raise the ConflictException in the pinpoint client, catch it in the provider task, log the exception, set notification to permanent failure so we don't retry and send the callback for the notification status.
LGTM!!

PR #2443, opened 2025-02-05
Add `feedback_reason` column to the notification table

whabanks:
# Summary | Résumé

This PR adds a new column, `feedback_reason`, to the notifications table. It will be leveraged to track more granular reasons why a notification fails. It also adds a new notification_status `pinpoint-failure`.

E.g. boto validates that the requested resource on AWS can handle the request being made through it. Since we disabled international sending in Pinpoint boto will throw an exception. 

1. Tests pass
2. `flask db upgrade` and `flask db downgrade` work locally

None.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1548,6 +1548,7 @@ NOTIFICATION_FAILED = "failed"
 NOTIFICATION_TECHNICAL_FAILURE = "technical-failure"
 NOTIFICATION_TEMPORARY_FAILURE = "temporary-failure"
 NOTIFICATION_PERMANENT_FAILURE = "permanent-failure"
+NOTIFICATION_PINPOINT_FAILURE = "pinpoint-failure"
 NOTIFICATION_PENDING_VIRUS_CHECK = "pending-virus-check"
 NOTIFICATION_VALIDATION_FAILED = "validation-failed"
 NOTIFICATION_VIRUS_SCAN_FAILED = "virus-scan-failed"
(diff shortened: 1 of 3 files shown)


app/models.py
```diff
@@ -1548,6 +1548,7 @@ def check_code(self, cde):
 NOTIFICATION_TECHNICAL_FAILURE = "technical-failure"
 NOTIFICATION_TEMPORARY_FAILURE = "temporary-failure"
 NOTIFICATION_PERMANENT_FAILURE = "permanent-failure"
+NOTIFICATION_PINPOINT_FAILURE = "pinpoint-failure"
```
jimleroyer, line 1551: Let's name the status to untie from a specific technology. We could go back to AWS SES or we could support Twilio as well and other providers. Is there a way to name this without depending on the implementation (and extend the model on an abstract name)?

jzbahrai: Currently all the failures are SES related. This is now Piinpoint related and we wanted to store that and make it obv. I think as we grow and add more providers/ clients/ - we should add failures according to those implementations so we might know exactly where the failure came from and not have to traceback through logs

jzbahrai (approved):

app/models.py
whabanks: I agree with not tying statuses to functionality or vendor specific terminology. That said, properly achieving vendor agnostic terminology for statuses, while still improving observability would take some effort. 

Since we're handling a conflict between a Request - `send to intl. #`, and the configuration of the target provider resource - `pinpoint not configured for intl. sending`, we _could_ use something more generic like:
- `PROVIDER_FAILURE`
- `PROVIDER_CONFLICT_FAILURE`
- `PROVIDER_VALIDATION_FAILURE`

Then capture the vendor specific info + failure reason in the `feedback_reason`. To Jumana's point - the current solution does provide added observability while requiring minimal changes to existing infrastructure for the time being.

jimleroyer: Hey I missed these replies. I am not big on tying our implementations with vendor specific names, especially when changing that to be generic would have been minimal effort IMO. We are starting a technical debt here that could have been avoided easily. This breaks abstractions and good design in general. I would bring this to the code review sessions.

PR #2453, opened 2025-02-12
Remove unused celery queue

smcmurtry:
# Summary | Résumé

Both the queue name and the variable it is assigned to only appear in this file.


sastels (approved):
🔥

PR #2454, opened 2025-02-12
Remove letter queues

smcmurtry:
# Summary | Résumé

This PR removes the following celery queues: `bulk-tasks`, `priority-tasks`, and `normal-tasks`. These have been replaced with sms and email specific queues and they were only being used for letter sending, so they should be safe to remove.


app/notifications/process_notifications.py
```diff
@@ -267,14 +267,14 @@ def send_notification_to_queue(notification, research_mode, queue=None):
         if notification.sends_with_custom_number():
             deliver_task = provider_tasks.deliver_throttled_sms
             queue = QueueNames.SEND_THROTTLED_SMS
-        if not queue or queue == QueueNames.NORMAL:
+        if not queue:
```
smcmurtry, line 270: This looks like a fallback that was used when the sms/email specific queues were released. It is no longer possible to have `QueueNames.NORMAL` as far as I can tell.

jimleroyer: Let's test the 💩 out of this before releasing with some bug bash?

jimleroyer (approved):
Thanks for the clean up! let's test this out extensively before the release. 🙏

app/notifications/process_notifications.py
smcmurtry: Sounds good! I will wait until Monday to merge this.

smcmurtry:
Testing will be time-consuming - I'll re-visit this if/when removing the letter code is prioritized

PR #2460, opened 2025-02-14
Make poetry venv path static

whabanks:
This PR adds quality of life changes to the devcontainer around poetry virtual environments, ensuring that:
1. The Python interpreter in Vscode is detected and set automatically between container builds
2. New shells created in Vscode post-build always activate the poetry virtual environment, ensuring that use of the `poetry run ...` prefix is never necessary

**Changes to dev container configs:**
- Disabled Poetry's auto venv creation
- Manually create the venv to allow setting a static venv path `~/.venv/workspace`
- Set vscode default interpreter path to the static venv
- Ensure all shells started post container creation auto-activate the venv

**Some background**
By default Poetry automatically creates virtual environments and names them based on a hash value. Although Vscode is smart enough to find these venvs, it's not smart enough to automatically choose them.

image

[Vscode internally caches the interpreter path once you choose it](https://github.com/microsoft/vscode-python/wiki/Setting-descriptions#experience), however this cached setting seems to persist across all running instances of vscode. i.e If you set your interpreter path while working in API, then switch over to Admin, rebuild the container and go to select the interpreter the cached path for API is still present:

image

This means each time select an interpreter you're effectively screwing up the setting for the next container you build. By setting each project's venv path to `~/.venv/workspace`, we can get around this as vscode will cache it and automatically pick up the venv regardless of the project you're working in.

1. In vscode - `ctrl + shift + p` and select `Python: clear workspace interpreter setting`
2. Rebuild the dev container 
3. Once the container is built, you may need to select the interpreter again
4. Open a new terminal / shell
5. Note that the virtual environment is automatically activated

## Ensure venv path is maintained across separate container builds
1. Check out [this admin PR](https://github.com/cds-snc/notification-admin/pull/2076)
2. Rebuild the container **without cache**
3. Note that the interpreter is automatically selected
4. Open a terminal, note that the correct venv is automatically activated

At this point, you should no longer have to manually activate the interpreter when switching between the Admin and API containers.

## Things to be aware of until this fix is propagated to all our containers
- Proper automatic venv selection will only work between containers with this fix.
- Building a container that doesn't have this fix applied to it will still prompt you to select the venv, thus overwriting the internally cached path. This will require you to select the interpreter again when building a "fixed" container after building a container without the fix.


jimleroyer (approved):
Nice improvement 👍

PR #2464, opened 2025-02-18
Add integrity error handling for email branding and template categories

whabanks:
# Summary | Résumé

This PR adds `IntegrityError` handling when creating new email branding or template categories. Attempting to save either using already taken names will result in a proper error message.

* https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/1774

## New Email Branding
1. Login as platform admin
2. In the admin panel select email branding and copy a name from the list to you clipboard
3. Click New branding
4. Fill out the form, pasting the name you copied in step 2
5. Click Save
6. Note the error message 

## New Template Category (EN & FR)
1. Login as platform admin
2. In the admin panel select `Template Categories` and copy a name from the list to you clipboard
3. Click `New category`
4. Fill out the form, pasting the name you copied in step 2 into the English name field
5. Click Save
6. Note the error message 
7. Switch the language to french and repeat the above steps, filling out the French name field instead

None.


app/email_branding/rest.py
```diff
@@ -18,6 +19,18 @@
 register_errors(email_branding_blueprint)
 
 
+@email_branding_blueprint.errorhandler(IntegrityError)
+def handle_integrity_error(exc):
+    """
+    Handle integrity errors caused by the unique constraint on uq_email_branding_name
+    """
+    if "uq_email_branding_name" in str(exc):
+        return jsonify(result="error", message="Email branding name already exists"), 400
```
jzbahrai, line 28: Given this is api, this returns in eng. I think its fine as its the admin page, but if there is a way to add to the admin page so its fr as well, that would be idea.

jzbahrai (commented):
tested based on the above scenarios! lgtm

app/email_branding/rest.py
whabanks: We can pass the english string from the API to admin and translate it in admin, yes.

jzbahrai:
Actually I am seeing this in the API logs:
```
  File "/workspace/app/email_branding/rest.py", line 59, in create_email_branding
    dao_create_email_branding(email_branding)
  File "/workspace/app/dao/dao_utils.py", line 18, in commit_or_rollback
    current_app.logger.error(e)
Message: IntegrityError('(psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "uq_email_branding_name"\nDETAIL:  Key (name)=(test) already exists.\n')
Arguments: ()
```
this will throw an error in cloudwatch, we can catch this as this stage

whabanks:
> Actually I am seeing this in the API logs:
> 
> ```
>   File "/workspace/app/email_branding/rest.py", line 59, in create_email_branding
>     dao_create_email_branding(email_branding)
>   File "/workspace/app/dao/dao_utils.py", line 18, in commit_or_rollback
>     current_app.logger.error(e)
> Message: IntegrityError('(psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "uq_email_branding_name"\nDETAIL:  Key (name)=(test) already exists.\n')
> Arguments: ()
> ```
> 
> this will throw an error in cloudwatch, we can catch this as this stage

I've updated the endpoints to catch and raise specific exceptions and created an [Admin PR](https://github.com/cds-snc/notification-admin/pull/2094) to display the errors in EN and FR.

andrewleith (changes requested):
Looks good but i am also seeing an exception being unhandled in the console for some reason...

app/email_branding/rest.py
```diff
@@ -38,24 +39,33 @@ def create_email_branding():
...
-
     email_branding = EmailBranding(**data)
+
     if "text" not in data.keys():
         email_branding.text = email_branding.name
 
-    dao_create_email_branding(email_branding)
+    try:
+        dao_create_email_branding(email_branding)
+    except IntegrityError as e:
```
andrewleith, line 49: Just a note that this constraint means we could have brandings called `Test` and `test` and they wouldn't conflict :/

whabanks:
> Looks good but i am also seeing an exception being unhandled in the console for some reason...

Yea it is still being logged. It's coming from our [`transactional` decorator](https://github.com/cds-snc/notification-api/blob/502aba813daea64b8c9a93be07bc62611dad947a/app/dao/dao_utils.py#L18). If we want to eliminate logging of UniqueViolations and related exceptions we could either:

1. Not log it in the decorator and instead make the caller responsible for choosing to log the message or not. Not a fan of this one as we'd potentially miss out on logging if the implementor forgets to add logging.
2. Build more comprehensive exception handling in for the decorator to raise a more specific exception, such as the ones [found here](https://github.com/cds-snc/notification-api/pull/2464/files#diff-463068b4d0340d64d95bb9d1e816b1d4da643fd30341c637f2745fff3609559bR86). Something like this?

```
def transactional(func):
    @wraps(func)
    def commit_or_rollback(*args, **kwargs):
        from flask import current_app

        try:
            res = func(*args, **kwargs)
            db.session.commit()
            return res
        except Exception as e:
            db.session.rollback()
            specify_and_raise_exception(str(e))

    return commit_or_rollback

def specify_and_raise_exception(e):
    if "uq_email_branding_name" in e:
        raise CannotSaveDuplicateEmailBrandingError()
    elif "template_categories_name_en_key" in e:
        raise CannotSaveDuplicateTemplateCategoryError()
    else:
        current_app.logger.error(e)
        raise e
```

andrewleith:
> Yea it is still being logged. It's coming from our [`transactional` decorator](https://github.com/cds-snc/notification-api/blob/502aba813daea64b8c9a93be07bc62611dad947a/app/dao/dao_utils.py#L18). If we want to eliminate logging of UniqueViolations and related exceptions we could either:
> 
> 1. Not log it in the decorator and instead make the caller responsible for choosing to log the message or not. Not a fan of this one as we'd potentially miss out on logs if the implementor forgets to add them.
> 2. Build more comprehensive exception handling in for the decorator to raise a more specific exception, such as the ones > ...

As long as it doesn't set off alarms, its probably fine!

app/email_branding/rest.py
whabanks: We could potentially solve this by changing the name column to a case insensitive data type: [`citext`](https://www.postgresql.org/docs/current/citext.html). This way we don't need to perform the validation in API and add another query to the process.

andrewleith (approved):
LGTM - as I mentioned, as long as the logged error doesnt cause us alarms, i think its fine for it to be there.  This is admin side anyway.

PR #2479, opened 2025-03-14
Add created_by_id columns to email branding and template categories tables

whabanks:
# Summary | Résumé
This PR is the first in a series of PRs to add `created_by_id` columns to many of our entities in an effort to ensure that our Cypress tests can clean up after themselves and lead to greater stability of our Cypress tests. 

In the past we had issues where Cypress tests could time out and fail. In one particular instance the Cypress service had ~1k templates, causing the templates page to take too long to load before the test would time out. 

Additionally, knowing who created or updated what entity, and when, improves our audit-ability. 

Added / ensured that the following columns are present on the `template_categories` and `email_branding` tables:
- `created_at`
- `created_by_id`
- `updated_at`
- `updated_by_id`

**Slip ins:**
- Updated `.git-blame-ignore-revs`, fixing an bad commit SHA and adding the latest Ruff commit SHA.

* https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/1789

CI is passing 
Refer to the test instructions in the [corresponding Admin PR](https://github.com/cds-snc/notification-admin/pull/2114)

None.

### DIFF
--- a/app/dao/services_dao.py
+++ b/app/dao/services_dao.py
@@ -42,6 +42,7 @@ from app.models import (
     ServicePermission,
     ServiceSmsSender,
     Template,
+    TemplateCategory,
     TemplateHistory,
     TemplateRedacted,
     User,
@@ -403,6 +404,8 @@ def delete_service_and_all_associated_db_objects(service):
     list(map(db.session.delete, verify_codes))
     db.session.commit()
     users = [x for x in service.users]
+    for user in users:
+        _delete_commit(TemplateCategory.query.filter_by(created_by_id=user.id))
     map(service.users.remove, users)
     [service.users.remove(x) for x in users]
     _delete_commit(Service.get_history_model().query.filter_by(id=service.id))
--- a/tests/app/email_branding/test_rest.py
+++ b/tests/app/email_branding/test_rest.py
@@ -4,9 +4,20 @@ from app.models import BRANDING_ORG_NEW, EmailBranding
 from tests.app.db import create_email_branding
 
 
-def test_get_email_branding_options(admin_request, notify_db, notify_db_session, sample_organisation):
-    email_branding1 = EmailBranding(colour="#FFFFFF", logo="/path/image.png", name="Org1", organisation_id=sample_organisation.id)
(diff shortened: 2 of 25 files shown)


app/dao/services_dao.py
```diff
@@ -403,6 +404,8 @@ def _delete_commit(query):
     list(map(db.session.delete, verify_codes))
     db.session.commit()
     users = [x for x in service.users]
+    for user in users:
+        _delete_commit(TemplateCategory.query.filter_by(created_by_id=user.id))
```
andrewleith, line 408: What happens if a TC created by this user is still in use?

tests/app/email_branding/test_rest.py
```diff
@@ -230,18 +277,46 @@ def test_post_update_email_branding_returns_400_if_name_is_duplicate(admin_reque
...
+        (
+            {
+                "logo": "images/text_x3.png",
+            }
+        ),
+        (
+            {
+                "logo": "images/text_x3.png",
+            }
+        ),
```
andrewleith, line 305: Are these just formatting changes?  Why are they triggering now? 🤔

andrewleith (commented):
Just wanted to make a note that in addition to making it possible to automate test clean-up, this also gives us better auditing info on these entities.  So this is useful in two different ways. Until now, we couldn't tell who created these entities, which is a bit of an oversight on our part.

Looks good in general - nicely done!  This isn't easy to do!  A couple of questions/comments below.

app/dao/services_dao.py
whabanks: As it stands currently if the category that a user created is in use by any templates that user / user's service does not own then this delete will fail with a foreign key constraint violation:`fk_template_template_categories`. 

So to delete a user we'd need to delete their TC which doesn't make sense if it's in use by other templates. If we remove the FK constraint between the user <-> TC we end up with TC's referencing users that may not exist. Technically that should be fine as we typically just archive a user rather than delete.

andrewleith (approved):
LGTM!

PR #2485, opened 2025-03-25
Add a `get_service_reports` endpoint

smcmurtry:
# Summary | Résumé

This PR:
- adds a new endpoint `get_service_reports` that returns all reports for a service that have been requested within `limit_days` (so we can avoid displaying a bunch of expired reports from a year ago on the page)
- adds a new dao function for this: `get_reports_for_service`
- adds unit tests for the above

* https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/1805

- tests pass
- code looks good

### DIFF
--- a/app/dao/reports_dao.py
+++ b/app/dao/reports_dao.py
@@ -1,3 +1,6 @@
+import datetime
+from typing import List
+
 from app import db
 from app.dao.dao_utils import transactional
 from app.models import Report
@@ -16,3 +19,21 @@ def create_report(report: Report) -> Report:
     db.session.commit()
 
     return report
+
+
+def get_reports_for_service(service_id: str, limit_days: int) -> List[Report]:
+    """
+    Get all reports for a service generated within the specified number of days.
+    Args:
+        service_id: The UUID of the service
+        limit_days: Number of days to look back (default: 30)
+    Returns:
+        List of Report objects sorted by requested_at (newest first)
+    """
+    query = Report.query.filter_by(service_id=service_id)
+
+    if limit_days is not None:
+        date_threshold = datetime.datetime.utcnow() - datetime.timedelta(days=limit_days)
+        query = query.filter(Report.requested_at >= date_threshold)
+
+    return query.order_by(Report.requested_at.desc()).all()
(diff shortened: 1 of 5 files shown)


app/dao/reports_dao.py
```diff
@@ -16,3 +19,21 @@ def create_report(report: Report) -> Report:
...
+    Args:
+        service_id: The UUID of the service
+        limit_days: Number of days to look back (default: 30)
+    Returns:
+        List of Report objects sorted by requested_at (newest first)
+    """
+    query = Report.query.filter_by(service_id=service_id)
+
+    if limit_days is not None:
+        date_threshold = datetime.datetime.utcnow() - datetime.timedelta(days=limit_days)
```
andrewleith, line 36: Pylance is telling me:

> The method "utcnow" in class "datetime" is deprecated
> Use timezone-aware objects to represent datetimes in UTC; e.g. by calling .now(datetime.timezone.utc)

Same thing in the tests

andrewleith (approved):
LGTM just a small issue with a deprecated datetime method.

How does expiry work here?  I dont see any business code to set it based on request_date - will that come later?

smcmurtry:
> How does expiry work here? I dont see any business code to set it based on request_date - will that come later?

@andrewleith Yeah that's going to come later. There will be a celery function that picks up requested reports, generates them, and updates the entry in the db.

app/dao/reports_dao.py
smcmurtry: It looks like it is going to be hard to switch to timezone-aware datetime objects because it's not backwards compatible with the timezone-naive datetimes we use everywhere else in the app. If we try to compare a tz-aware to a tz-naive datetime, that will throw a `TypeError` and could break things. Here's some sample code that shows the issue:
Screenshot 2025-03-25 at 11 37 18 AM

andrewleith: Wow it got deprecated and they recommend using something that doesn't even return the same type? Why, python, why?

PR #2486, opened 2025-03-27
Add a celery task to generate the reports

smcmurtry:
# Summary | Résumé

Add a celery task to generate the reports. This PR does the following:
- adds a new scheduled celery task `run-generate-reports` that runs every minute
- adds a new celery task `generate-report` that is called by the scheduled task
- adds a new function `create_report_in_s3` that gets the report data from the db and uploads the file to s3
- adds some new functions in app/report/utils.py that are used to generate the csv file

What this PR doesn't do - I'll do these in follow up PRs:
- doesn't get more than the first page of report data (only the first 5000 rows)
- doesn't do a multipart s3 upload - this will be needed for large reports
- doesn't generate a pre-signed s3 url
- doesn't translate the csv file into french

* https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/1807

1. Add `REPORTS_BUCKET_NAME=notification-canada-ca-staging-reports` to your local .env file
1.  Check out this branch and run locally.
2. run notification-admin locally as well
3. send a few notifications from your service so that your report will not be empty
4. Go to `http://localhost:6012/services/your-service-id/reports`
5. generate a report
6. wait 2 or 3 minutes and refresh the page
7. verify that the report status has been updated to "ready"
8. log in to aws staging and locate the new reports s3 bucket by searching for "reports" in the s3 bucket list
10. download your report from the following location: `notification-canada-ca-staging-reports/service-<your-serivice-id>/<your-report-id>.csv`
11. verify that the contents of the csv file are correct

### DIFF
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -400,4 +400,4 @@ def run_generate_reports():
 
     for report in requested_reports:
         current_app.logger.info(f"calling generate_report for Report ID {report.id}")
-        generate_report.apply_async([report.id], queue=QueueNames.BULK_DATABASE)
+        generate_report.apply_async([report.id], queue=QueueNames.REPORTING)
--- a/app/celery/scheduled_tasks.py
+++ b/app/celery/scheduled_tasks.py
@@ -393,42 +393,38 @@ def beat_inbox_sms_priority():
         receipt_id_sms, list_of_sms_notifications = sms_priority.poll()
 
 
-@notify_celery.task(name="generate-reports")
+@notify_celery.task(name="run-generate-reports")
 @statsd(namespace="tasks")
-def generate_reports():
+def run_generate_reports():
+    current_app.logger.error("starting run-generate-reports")
+
     # query for reports that have been requested but haven't been generated yet
     requested_reports = Report.query.filter(Report.status.in_([ReportStatus.REQUESTED.value])).order_by(Report.requested_at).all()
 
-    try:
-        for report in requested_reports:
-            # mark the report as generating
-            report.status = ReportStatus.GENERATING.value
-            update_report(report)
-
-            generate_report(report)
-
-            # mark the report as ready
-            report.status = ReportStatus.READY.value
-            update_report(report)
-
-            current_app.logger.info(f"Report ID {str(report.id)} has been generated")
-
-    except SQLAlchemyError:
-        current_app.logger.exception("Failed to generate reports complete")
-        raise
+    for report in requested_reports:
+        current_app.logger.info(f"calling generate_report for Report ID {report.id}")
+        generate_report(report)
 
 
 def generate_report(report: Report):
-    current_app.logger.exception(f"Generating report for Report ID {report.id}")
+    current_app.logger.info(f"Generating report for Report ID {report.id}")
+    test_data = b"test data"
     try:
-        # save a fake file to s3
-        test_data = "test data"  # probably needs to be encoded differently
+        # mark the report as generating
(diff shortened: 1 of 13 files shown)


app/config.py
```diff
@@ -380,6 +381,11 @@ class Config(object):
             "schedule": crontab(),
             "options": {"queue": QueueNames.PERIODIC},
         },
+        "run-generate-reports": {
+            "task": "run-generate-reports",
+            "schedule": crontab(),
+            "options": {"queue": QueueNames.PERIODIC},
```
jimleroyer, line 387: This is a task that can take quite some time if I am not mistaken? I wonder if we should locate it in a queue and tasks of its own, as this might slow down other periodic tasks when a report is generated. We have 4 workers per pod, and the primary pods where that would run can scale to 30, if I am not wrong.. but the time to scale up could slow down the other tasks. I would like to speak about this with you all and @ben851 .

smcmurtry: Yes the task could take up to 20 minutes for our biggest services. Good idea with the dedicated queue! Makes sense to me.

app/celery/scheduled_tasks.py
```diff
@@ -385,3 +388,16 @@ def beat_inbox_sms_priority():
...
+@statsd(namespace="tasks")
+def run_generate_reports():
+    current_app.logger.error("starting run-generate-reports")
+
+    # query for reports that have been requested but haven't been generated yet
+    requested_reports = Report.query.filter(Report.status.in_([ReportStatus.REQUESTED.value])).order_by(Report.requested_at).all()
+
+    for report in requested_reports:
+        current_app.logger.info(f"calling generate_report for Report ID {report.id}")
+        generate_report.apply_async([report.id], queue=QueueNames.BULK_DATABASE)
```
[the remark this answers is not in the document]
jimleroyer: Alright, then we could isolate that `run-generate-reports` into its own. I see there is a `reporting-tasks` queue that we could reuse.

jzbahrai:
I see the reports in s3.

Thing I couldnt' test - the report's page doesn't show "ready" - though the s3 bucket has the report
Csv gets deleted after a certain retention date

app/celery/scheduled_tasks.py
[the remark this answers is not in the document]
smcmurtry: Okay I moved both new tasks to the `reporting-tasks` queue

app/celery/tasks.py
```diff
@@ -874,3 +883,48 @@ def seed_bounce_rate_in_redis(service_id: str, interval: int = 24):
...
+        current_app.logger.info(f"Report ID {str(report.id)} has been generated")
+    except Exception as e:
+        current_app.logger.exception(f"Failed to generate report for Report ID {report.id}: {str(e)}")
+        report.status = ReportStatus.ERROR.value
+        update_report(report)
+        raise
+
+
+def create_report_in_s3(report: Report) -> str:
+    """Creates a report in S3 and returns the URL"""
```
jzbahrai, line 916: maybe we should add some logging here to keep track of how long it is taking to generate this report?

jzbahrai (approved):

app/celery/tasks.py
smcmurtry: I'm seeing logs with timing for the `generate-report` task that calls this function - so that will give us the time to generate each report and the report id. I think the `@statsd` decorator is doing that for us. 
Screenshot 2025-04-01 at 1 24 00 PM

PR #2502, opened 2025-04-11
Switch to using postgres `copy` and stream csv data to s3

smcmurtry:
# Summary | Résumé

This PR:
- uses the "copy" feature in postgresql https://www.postgresql.org/docs/current/sql-copy.html 
- reduces memory usage in the celery worker by streaming the csv data to s3, rather than trying to store the whole csv file in memory
- adds tests for this

I'd like to test this on staging with our perf test service to see how it does with a large file.

- check out the branch locally
- generate a report
- verify that the report looks correct

### DIFF
--- a/app/aws/s3.py
+++ b/app/aws/s3.py
@@ -5,6 +5,7 @@ from typing import List
 import botocore
 import pytz
 from boto3 import client, resource
+from boto3.s3.transfer import TransferConfig
 from flask import current_app
 from notifications_utils.s3 import s3upload as utils_s3upload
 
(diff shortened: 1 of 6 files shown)


app/aws/s3.py
```diff
@@ -189,3 +191,33 @@ def generate_presigned_url(bucket_name: str, object_key: str, expiration: int =
...
+def stream_to_s3(bucket_name, object_key, copy_command, cursor):
+    """
+    Stream data from PostgreSQL COPY command directly to S3.
+
+    :param bucket_name: S3 bucket name
+    :param object_key: S3 object key
+    :param copy_command: PostgreSQL COPY command
+    :param cursor: Database cursor
+    """
+    s3_client = client("s3", current_app.config["AWS_REGION"])
```
jzbahrai, line 205: we really should make this a class and have all these functions be under it, instead of instantiating an s3 connection for each func. (not for this PR but in the future)

app/aws/s3.py
```diff
@@ -189,3 +191,33 @@ def generate_presigned_url(bucket_name: str, object_key: str, expiration: int =
...
+    config = TransferConfig(multipart_threshold=1024 * 25, max_concurrency=10)
```
jzbahrai, line 206: can we pass these two numbers in as variables in the func?
is it worth putting this in config

app/report/utils.py
```diff
@@ -44,3 +51,97 @@ def get_csv_file_data(serialized_notifications: list[Any], lang="en") -> bytes:
...
+    j = aliased(Job)
+    u = aliased(User)
+
+    # Build the query using SQLAlchemy
+    return (
+        db.session.query(
+            n.to.label("Recipient"),
+            t.name.label("Template"),
+            n.notification_type.label("Type"),
+            func.coalesce(u.name, "").label("Sent by"),
```
jzbahrai, line 80: when i ran this locally both the sent_by and sent_by_email were empty in the file

smcmurtry: I believe that is supposed to happen if the sends were via the api. Did you send via api calls?

app/aws/s3.py
smcmurtry: I've moved them to be args to that function. We could move them to config if it turns out that we want different values for different environments. Currently these are just guesses at values to test out.

jzbahrai (approved):

PR #2515, opened 2025-04-29
Removed the heartbeat specific filters from API

jimleroyer:
# Summary | Résumé

Removing specific filters for heartbeat notifications. This won't be needed as we moved the heartbeat notifications into a dedicated service and after we rework the existing heartbeat in the database so they are associated with the new service.

Related task:
https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/560

Test the dashboards that were affected with the filters, make sure these before and after the changes report the same numbers for previous periods of time.


tests/app/v2/notifications/test_get_bulk_jobs.py
```diff
@@ -73,10 +73,10 @@ def test_get_bulk_jobs_returns_only_jobs_for_authenticated_service(client, sampl
     assert str(other_job.id) not in returned_ids
 
 
-def test_get_bulk_jobs_supports_cursor_pagination(client, sample_template):
+def test_get_bulk_jobs_supports_cursor_pagination(client, sample_template, monkeypatch):
     first_job = create_job(sample_template, created_at=datetime(2026, 8, 18, 12, 0))
     second_job = create_job(sample_template, created_at=datetime(2026, 8, 18, 11, 0))
-    client.application.config["API_PAGE_SIZE"] = 1
+    monkeypatch.setitem(client.application.config, "API_PAGE_SIZE", 1)
```
jimleroyer, line 79: @copilot Why would this test fail without the monkeypatch? It worked before and it's not related to the changes.

jimleroyer: @copilot But why would the application.config not be sufficient? Especially as it worked before?

ben851: @copilot This test is not a new test, so it should have worked before. My copilot session in VS Code said the following:
```
It likely started failing because merging latest changed the test execution order, exposing an existing order-dependent test.

notify_api is session-scoped, so client.application.config is shared across the suite. The bulk-jobs cursor test sets API_PAGE_SIZE = 1; any v2 notification-list tests run afterward will be limited to one notification. Your heartbeat changes did not alter that endpoint’s behavior, but the merge can easily have moved or added enough tests to make this ordering happen consistently.

The monkeypatch change makes that test restore the config automatically, removing the order dependency.
```

Does this make more sense?

ben851: @copilot Does recent the parallel tests changes might have caused this (it was all sequential before), done in github actions that you can find in the .github/workflows folder? as it might have thrown around the execution order. If not, which changes?

jimleroyer (commented):
LGTM - I can't approve though because I opened that PR months ago and @ben851 took over. Feel free @ben851 to approve yourself and merge.

ben851 (approved):
As per Jimmy, the monkeypatch is good to go and this can be merged.

PR #2533, opened 2025-05-16
[WIP] Fix "add user to service" bug

smcmurtry:
# Summary | Résumé

When adding a user to a service, the code is doing two operations that can conflict:

1. First, it adds the user to the service with `service.users.append(user)`
2. Then, it tries to get the created `ServiceUser` record with `dao_get_service_user(user.id, service.id)`
3. The error occurs because when calling `dao_get_service_user`. SQLAlchemy tries to auto-flush the session to ensure all pending changes are committed before executing the query. This causes the uniqueness constraint violation - it's trying to insert the record, but the uniqueness constraint check sees the same user-service combination is already about to be inserted.

The error specifically mentions: [duplicate key value violates unique constraint `"uix_user_to_service"` for the combination of user_id and service_id.

The Fix
I implemented a solution that:

1. First checks if the user is already associated with the service
2. If not, it creates the association and the ServiceUser object explicitly
3. If the user is already associated, it just updates the permissions


jimleroyer:
I am not sure I follow that:
> The error occurs because when calling dao_get_service_user. SQLAlchemy tries to auto-flush the session to ensure all pending changes are committed before executing the query. This causes the uniqueness constraint violation - it's trying to insert the record, but the uniqueness constraint check sees the same user-service combination is already about to be inserted.

Flushing the session would finalize the insert for the service to user associations and the next query is a select, but you imply that the insert would happen twice? That would go against the default expected behavior of sqlalchemy too, to have everything working in a single transaction with the ongoing db session.

PR #2536, opened 2025-05-20
Fix/handle unique violation

smcmurtry:
# Summary | Résumé

This PR:
- handles the `UniqueViolation` that we were seeing get raised from the dao fn
- Return a [409 error](https://stackoverflow.com/questions/3825990/http-response-code-for-post-when-resource-already-exists) when the user already belongs to the service. This is so we don't trigger our alarms
- tests coming soon

Related admin PR: https://github.com/cds-snc/notification-admin/pull/2181

* https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/1914

We can't reproduce the bug, so I'm not sure how to test this.

### DIFF
--- a/app/service/rest.py
+++ b/app/service/rest.py
@@ -11,6 +11,7 @@ from notifications_utils.clients.redis import (
     over_email_daily_limit_cache_key,
     over_sms_daily_limit_cache_key,
 )
+from psycopg2.errors import UniqueViolation
 from sqlalchemy import func
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm.exc import NoResultFound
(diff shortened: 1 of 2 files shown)


jzbahrai (approved):

app/service/rest.py
```diff
@@ -466,16 +467,23 @@ def add_user_to_service(service_id, user_id):
...
     permissions = [Permission(service_id=service_id, user_id=user_id, permission=p["permission"]) for p in data["permissions"]]
     folder_permissions = data.get("folder_permissions", [])
 
-    dao_add_user_to_service(service, user, permissions, folder_permissions)
+    try:
+        dao_add_user_to_service(service, user, permissions, folder_permissions)
+    except UniqueViolation:
+        message = f"UniqueViolation: User id: {user_id} already part of service id: {service_id}"
+        current_app.logger.info(message)
+        raise UserAlreadyInServiceError(status_code=409, message=message)
```
jimleroyer, line 485: I like business specific exception to raise up to business code higher in the stack, good job!

smcmurtry: I gave that a try but it resulted in a circular import error. So I went back to the easy solution. But thanks for the suggestion, I agree that would be best practice.

PR #2543, opened 2025-05-29
Revert "Add code to test for an user being added to a service they are already in"

jzbahrai:
Reverts cds-snc/notification-api#2540


jimleroyer:
We will keep this revert PR opened during the weekend, if anything happens.

whabanks (approved):

PR #2549, opened 2025-06-05
Fix/make timing query faster

jzbahrai:
# Summary | Résumé

The query to get the maximum date from the ft_notification_status was inefficient, so we made that efficient.
And added an index for a fast check:

```
 updated_at          | timestamp without time zone |           |          | 
 billable_units      | integer                     |           | not null | 
Indexes:
    "ft_notification_status_pkey" PRIMARY KEY, btree (bst_date, template_id, service_id, job_id, notification_type, key_type, notification_status)
    "ix_ft_notification_service_bst" btree (service_id, bst_date)
    ```
 
 I tested both the upgrade and the downgrade


whabanks (approved):
LGTM, +1 for that composite key.

PR #2577, opened 2025-06-30
Tag generated reports with `earliest_created_at` metadata

whabanks:
# Summary | Résumé

This PR updates the report generation process for jobs to make it simpler for the front end to validate whether a given report contains notification data that is older than the service's retention period.
- When a job report is generated, the `earliest_created_at` date is pulled from the result set
- When the streamed to S3, the file is tagged with new metadata `earliest_created_at`

1. CI is passing
2. Once merged, it can be properly tested on the admin PR.

None.

### DIFF
--- a/app/aws/s3.py
+++ b/app/aws/s3.py
@@ -196,7 +196,13 @@ def generate_presigned_url(bucket_name: str, object_key: str, expiration: int = 
 
 
 def stream_to_s3(
-    bucket_name, object_key, copy_command, cursor, multipart_threshold=MULTIPART_THRESHOLD, max_concurrency=MAX_CONCURRENCY
+    bucket_name,
+    object_key,
+    copy_command,
+    cursor,
+    metadata=None,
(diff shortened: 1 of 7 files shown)


smcmurtry:
> This PR updates the report generation process for jobs to make it simpler for the front end to validate whether a given report contains notification data that is older than the service's retention period.

The report will never contain notification data older than the service's retention period since these notifications have been removed from the `notifications` table. It is just empty when we are past the retention period.

smcmurtry (changes requested):
If the goal here is to eventually return this data to admin when it calls the api,
I think we should just use `job.archived` from the db since that is the source of truth that other parts of the codebase already use.

It is set like this:
https://github.com/cds-snc/notification-api/blob/main/app/celery/nightly_tasks.py#L70

whabanks:
Closing as this our policy was reworded to allow for data outside of a services retention period to exist temporarily in generated reports.

PR #2589, opened 2025-07-21
Add fido2_key_id column to users table

whabanks:
# Summary | Résumé
This column will be used to determine which fido2 security key is currently in use for 2FA, and also to be able to display the key name in the user's profile in the new security section.

# Test instructions | Instructions pour tester la modification
1. `flask db upgrade` and `flask db downgrade` are working as expected.
2. CI is passing

# Release Instructions | Instructions pour le déploiement

None.

# Reviewer checklist | Liste de vérification du réviseur

- [ ] This PR does not break existing functionality.
- [ ] This PR does not violate GCNotify's privacy policies.
- [ ] This PR does not raise new security concerns. Refer to our GC Notify Risk Register document on our Google drive.
- [ ] This PR does not significantly alter performance.
- [ ] Additional required documentation resulting of these changes is covered (such as the README, setup instructions, a related ADR or the technical documentation).

> ⚠ If boxes cannot be checked off before merging the PR, they should be moved to the "Release Instructions" section with appropriate steps required to verify before release. For example, changes to celery code may require tests on staging to verify that performance has not been affected.

PR #2590, opened 2025-07-22
Revert "Add fido2_key_id column to users table"

whabanks:
Reverts cds-snc/notification-api#2589

We decided against only allowing a specific key for login. Instead we will continue to allow you to log in using any of your currently existing security keys.


whabanks:
Closed because I forgot we can't revert migrations 👎

PR #2604, opened 2025-08-14
Revert "re-add the migration but do not use the process_type (#2603)"

smcmurtry:
# Summary | Résumé

This reverts commit 08ce0aaefb16e5f62a57f56165c730d46eeae92d.

I've downgraded staging manually. This was preventing log in on staging because the `template_category_id` was not set. I have been able to fix the issue locally and I'll put up a new PR with the working migration.


ben851 (approved):
LGTM

PR #2616, opened 2025-09-04
New Relic Variables and On/Off Switch Docker implementation (STAGING)

P0NDER0SA:
# Summary | Résumé

updating docker file to choose handler based on NEW_RELIC_ENABLED var, adding all NEW_RELIC vars from param store to our actual Lambda configuration, outputting some useful logging info for the variable script

https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/656

### DIFF
--- a/.github/workflows/lambda_staging.yml
+++ b/.github/workflows/lambda_staging.yml
@@ -54,6 +54,37 @@ jobs:
             --function-name ${{ matrix.image }} \
             --image-uri $REGISTRY/${{ matrix.image }}:${GITHUB_SHA::7} > /dev/null 2>&1
 
+      - name: Sync NEW_RELIC_* variables from ENVIRONMENT_VARIABLES SSM to Lambda environment
+        run: |
+          PARAMS=$(aws ssm get-parameter \
+            --name ENVIRONMENT_VARIABLES \
+            --with-decryption \
+            --query 'Parameter.Value' \
+            --output text)
+
+          # Extract all NEW_RELIC_* variables from the parameter store content
+          NEW_RELIC_VARS=$(echo "$PARAMS" | grep '^NEW_RELIC_' || true)
+
+          # Build a jq filter to add all NEW_RELIC_* vars
+          JQ_FILTER='.'
+          while IFS= read -r line; do
+            [ -z "$line" ] && continue
+            key=$(echo "$line" | cut -d '=' -f 1)
+            value=$(echo "$line" | cut -d '=' -f 2-)
+            JQ_FILTER="$JQ_FILTER + {\"$key\": \"$value\"}"
+          done <<< "$NEW_RELIC_VARS"
+
+          EXISTING_VARS=$(aws lambda get-function-configuration \
+            --function-name ${{ matrix.image }} \
+            --query 'Environment.Variables' \
+            --output json)
+
+          echo "$EXISTING_VARS" | jq "$JQ_FILTER" > merged_env.json
+
+          aws lambda update-function-configuration \
+            --function-name ${{ matrix.image }} \
+            --environment "Variables=$(cat merged_env.json)"
+
       - name: Publish lambda version and update alias
         run: |
           aws lambda wait function-updated --function-name ${{ matrix.image }}
--- a/bin/entry.sh
+++ b/bin/entry.sh
@@ -1,7 +1,10 @@
 #!/bin/sh
 if [ -z "${AWS_LAMBDA_RUNTIME_API}" ]; then
+    echo "ENTRY.SH: Running locally"
     exec /usr/bin/aws-lambda-rie $(which python) -m awslambdaric $1
 else
-    . /sync_lambda_envs.sh # Retrieve .env from parameter store and remove currently set environement variables
+    echo "ENTRY.SH: Running in AWS Lambda"
+    . /sync_lambda_envs.sh
+    env  # Optional: dump all environment variables
     exec $(which python) -m awslambdaric $1
 fi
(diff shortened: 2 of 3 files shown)


bin/entry.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/sh
 if [ -z "${AWS_LAMBDA_RUNTIME_API}" ]; then
+    echo "ENTRY.SH: Running locally"
     exec /usr/bin/aws-lambda-rie $(which python) -m awslambdaric $1
 else
-    . /sync_lambda_envs.sh # Retrieve .env from parameter store and remove currently set environement variables
+    echo "ENTRY.SH: Running in AWS Lambda"
+    . /sync_lambda_envs.sh
+    env  # Optional: dump all environment variables
```
jimleroyer, line 8: This one might be dangerous, depending on where we run this and leak secrets.

.github/workflows/lambda_staging.yml
```diff
@@ -54,6 +54,37 @@ jobs:
...
+          EXISTING_VARS=$(aws lambda get-function-configuration \
+            --function-name ${{ matrix.image }} \
+            --query 'Environment.Variables' \
+            --output json)
+
+          echo "$EXISTING_VARS" | jq "$JQ_FILTER" > merged_env.json
+
+          aws lambda update-function-configuration \
+            --function-name ${{ matrix.image }} \
+            --environment "Variables=$(cat merged_env.json)"
```
jimleroyer, line 86: Interesting, so we are merging specific New Relic env vars with existing ones for now, right?

P0NDER0SA: yes, if we simply try and "add" one it just overwrites all of them-- we have to merge them if we want to keep the ones that were there. 

I actually think we could remove all of these from Terraform  to make this all less confusing

bin/entry.sh
P0NDER0SA: This thought occured to me as well -- but the reason for this was to actually show us in the logs what variables exist!  as it stands we have no idea what variables are exposed to the container because none of them show up in the lambda config env vars.

This gets run in the container entry script, and would show up in the lambda logs.  

I'm good to remove if you think we should, but i think it would also be helpful to have something to search for in cloudwatch when we want to verify our env vars :)

.github/workflows/lambda_staging.yml
```diff
@@ -54,6 +54,48 @@ jobs:
...
+            CONFIG_KEYS_TRUNCATED="${CONFIG_KEYS:0:$MAX_LEN}"
+            echo "CONFIG_KEYS truncated from ${#CONFIG_KEYS} to ${#CONFIG_KEYS_TRUNCATED} characters" >&2
+            CONFIG_KEYS="$CONFIG_KEYS_TRUNCATED"
+          fi
+
+          # Overwrite entire static environment: only keep these two keys
+          jq -n \
+            --arg nr "$NEW_RELIC_ENABLED" \
+            --arg keys "$CONFIG_KEYS" \
+            '{NEW_RELIC_ENABLED:$nr, CONFIG_KEYS:$keys}' > merged_env.json
```
jimleroyer, line 89: Why do we want to keep `NEW_RELIC_ENABLED`? and not have it part of CONFIG_KEYS listed like the others?

bin/entry.sh
```diff
@@ -4,9 +4,13 @@ if [ -z "${AWS_LAMBDA_RUNTIME_API}" ]; then
...
-All environment variable names (from AWS Parameter Store):
-${VAR_NAMES}"
+    # Collect environment variable names (sorted) into a JSON array for a single CloudWatch structured log line
+    VAR_NAMES_NL=$(env | cut -d '=' -f 1 | sort)
+    ENV_VAR_COUNT=$(printf '%s\n' "$VAR_NAMES_NL" | grep -c '^')
+    # Build JSON array manually (avoid requiring jq at runtime)
+    VAR_NAMES_JSON=$(printf '%s\n' "$VAR_NAMES_NL" | awk 'BEGIN { printf("["); first=1 } { gsub(/"/, "\\\""); if(!first) printf(","); printf("\"%s\"", $0); first=0 } END { printf("]") }')
+    TS=$(date -u +"%Y-%m-%dT%H:%M:%SZ" 2>/dev/null || date -u)
+    # Single structured JSON line
+    echo "{\"timestamp\":\"$TS\",\"source\":\"entry.sh\",\"mode\":\"lambda\",\"loader\":\"sync_lambda_envs.sh\",\"new_relic_enabled\":\"${NEW_RELIC_ENABLED:-unset}\",\"env_var_count\":$ENV_VAR_COUNT,\"env_var_names\":$VAR_NAMES_JSON}"
```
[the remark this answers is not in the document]
jimleroyer: Just a note, to use `jo`, we need to add it in the `Dockerfile.lambda` file:

```
RUN apt-get install -y bash git libtool  autoconf automake gcc  g++ make libffi-dev jo unzip
```

.github/workflows/lambda_staging.yml
```diff
@@ -31,11 +31,24 @@ jobs:
...
+            NEW_RELIC_ENABLED=false
+          fi
+          echo "Building image (toggle captured)" >&2
           docker build \
-          --build-arg GIT_SHA=${GITHUB_SHA::7} \
-          -t $REGISTRY/${{ matrix.image }}:${GITHUB_SHA::7} \
-          . \
-          -f ci/Dockerfile.lambda
+            --build-arg GIT_SHA=${GITHUB_SHA::7} \
+            --build-arg NEW_RELIC_ENABLED=${NEW_RELIC_ENABLED} \
```
jimleroyer, line 48: I see, I see 👀

PR #2633, opened 2025-09-16
Lambda Dockerfile NR Switch

P0NDER0SA:
# Summary | Résumé

updating docker file, entrypoint and staging workflow to hinge build and runtime on NEW_RELIC_ENABLED from param store

This will bake in the handler value and the required files for new relic (or exclude the files for new relic) at build time making the runtime very simple.

This is not the most flexible way of doing things, but if we're only adding the new relic layer files during build (as is a requirement for this) when it's enabled, it really doesn't make any sense to change the handler value/config at runtime because new relic would never work without rebuilding first

https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/656

None.

### DIFF
--- a/ci/Dockerfile.lambda
+++ b/ci/Dockerfile.lambda
@@ -49,19 +49,19 @@ RUN chmod 755 /usr/bin/aws-lambda-rie /entry.sh /sync_lambda_envs.sh
 # Decide handler based on NEW_RELIC_ENABLED and optionally extract New Relic layer
 RUN if [ "${NEW_RELIC_ENABLED}" = "true" ] || [ "${NEW_RELIC_ENABLED}" = "TRUE" ]; then \
         export APP_HANDLER=newrelic_lambda_wrapper.handler; \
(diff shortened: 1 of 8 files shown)


ci/Dockerfile.lambda
```diff
@@ -43,14 +46,22 @@ COPY bin/entry.sh /
...
+    fi; \
+    rm -f newrelic-layer.zip; \
+    echo "FINAL_APP_HANDLER=${APP_HANDLER}" > /app/handler_meta
 
-ENTRYPOINT [ "/entry.sh" ]
+# Generate a wrapper entrypoint that bakes the resolved APP_HANDLER at build time (validation not needed; derived from boolean NEW_RELIC_ENABLED)
+RUN APP_HANDLER=$(grep '^FINAL_APP_HANDLER=' /app/handler_meta | cut -d '=' -f2-) && \
+    printf '#!/bin/sh\nexec /entry.sh %s\n' "${APP_HANDLER}" > /handler_entry.sh && \
+    chmod +x /handler_entry.sh && \
+    echo "Baked APP_HANDLER=${APP_HANDLER}"
```
jimleroyer, line 65: Curious how that will turn out!

PR #2661, opened 2025-10-08
Revert "New Relic WSGI Wrapper"

jzbahrai:
Reverts cds-snc/notification-api#2659


jimleroyer (approved):
Reverting now and will reinstate later when work will resume.

PR #2843, opened 2026-04-21
Rate limiter on SMS sending / phase 1

jimleroyer:
# Summary | Résumé

Phase 1 for SMS rate limiter implementation. [ADR](https://github.com/cds-snc/ADR/blob/main/adr/notify/0029_GC_Notify_SMS_Sending_Rate_Control.md) [Issue](https://github.com/cds-snc/notification-planning-core/issues/786)

## AI disclaimer

The main code implementation was manually coded but AI assisted. The tests were generated with AI and reviewed by the developers.

### Overview

This PR introduces two complementary mechanisms to protect SMS delivery throughput: **SQS-level fair queuing** (immediate, per-service noisy-neighbour isolation) and a **global SMS parts rate limiter** (enforces a per-minute cap across all services). Together they prevent any single service from monopolising delivery capacity, whether at the queue level or the provider level.

---

#### 1. SQS fair queuing for all notification types (`app/celery/celery.py`)

`NotifyCelery.send_task` now passes `MessageGroupId` both as a **Celery task header** (`notify_message_group_id`) and as a **kombu message property** (`properties["MessageGroupId"]`).

The kombu SQS transport reads `message["properties"]["MessageGroupId"]` in `Channel._put()` before calling `boto3.send_message()`. Previously the value was only stored as a header and never reached SQS.

- **Standard queues** — AWS uses `MessageGroupId` to provide fair queuing (per-tenant noisy-neighbour protection; one slow service cannot starve others).
- **FIFO queues** — `MessageGroupId` is required by SQS to preserve per-group ordering.
- The default value is `notification.service_id`, so each service gets its own fair-share slot within any given queue.

#### 2. SMS parts rate limiter (`app/rate_limiter.py`, `FF_SMS_RATELIMIT`)

A new `RateLimiter` abstraction with two implementations:

- **`InMemoryRateLimiter`** — 60-second sliding window, suitable for a single worker (Phase 1).
- **`RedisRateLimiter`** — distributed sliding window backed by Redis, suitable for multi-worker deployments (Phase 2).

The limiter enforces `CELERY_DELIVER_SMS_RATE_LIMIT_PER_MINUTE` (parts/minute) globally and is gated behind `FF_SMS_RATELIMIT`.

#### 3. New `deliver_sms_rate_limited` task (`app/celery/provider_tasks.py`)

A new Celery task that sits between the queue and the provider:

1. Calls `get_rate_limiter().acquire_lease(parts_count)`.
2. If capacity is available → proceeds to `_deliver_sms`.
3. If the cap is hit → re-queues itself on `SEND_SMS_FAIR` with `countdown=seconds_to_wait` and the same `MessageGroupId` it arrived with (read via `self.message_group_id` from the task header), then raises `Ignore` to cleanly release the original task slot.

Re-queuing preserves `MessageGroupId` so the notification stays in the same SQS group and does not lose its fair-queue position across retries.

#### 4. Routing changes (`app/notifications/process_notifications.py`)

**`choose_queue`** determines the queue at notification creation time:

- Custom SMS sender (dedicated number) → `send-throttled-sms-tasks` _(always, even in research mode — the dedicated number must be respected)_
- Research mode / `key_type=test` → `research-mode-tasks`
- `FF_SMS_RATELIMIT=True`, normal SMS → `send-sms-fair-tasks`
- Otherwise → existing queue tiers (high/medium/low based on template category)

**`choose_deliver_task`** selects the Celery task at dispatch time:

- Custom sender → `deliver_throttled_sms`
- `FF_SMS_RATELIMIT=True` and not research mode → `deliver_sms_rate_limited`
- Otherwise → `deliver_sms`

**`db_save_and_send_notification`** and **`send_notification_to_queue`** both:

- Default `MessageGroupId` to `notification.service_id` — ensures fair queuing across all services on every queue, not just the fair queue.
- When routing to the fair queue, override `MessageGroupId` with the template's delivery queue name (e.g. `send-sms-medium-tasks`). This groups messages by their original priority class within the fair queue, so high-priority notifications from any service are not held behind bulk ones from the same service.

#### 5. New queue constant (`app/config.py`)

`SEND_SMS_FAIR = "send-sms-fair-tasks"`

---

### Feature flags

| Flag | Default | Effect |
|---|---|---|
| `FF_SMS_RATELIMIT` | `False` | Enables `deliver_sms_rate_limited` routing and the `send-sms-fair-tasks` queue |
| `FF_USE_BILLABLE_UNITS` | `True` | Uses pre-computed `billable_units` from the DB as `parts_count`; otherwise recomputes from the template |
| `SMS_RATE_LIMITER_BACKEND` | `"memory"` | `"memory"` for single-worker, `"redis"` for distributed |

---

One important caveat with the current changes is that the addition of message group ID that defaults to service ID and discriminating against services is not gated by the feature flags above. This change will go in as is but it's something we can discuss amongst reviewers, if deemed safe enough or not. 

### Routing decision tree (`FF_SMS_RATELIMIT=True`)

```
SMS notification
├── research mode OR key_type=test  →  research-mode-tasks        (deliver_sms)
├── sends_with_custom_number()      →  send-throttled-sms-tasks   (deliver_throttled_sms)
├── priority_queue provided         →  priority_queue             (deliver_sms_rate_limited)
└── default                         →  send-sms-fair-tasks        (deliver_sms_rate_limited)
                                         MessageGroupId = template's delivery queue name
```

* [SMS Rate Limiter](https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/786) / #786

1. Enable the feature flag `FF_SMS_RATELIMIT`.
1. Send tons of SMS to the API.
1. Monitor the sending rate in related AWS sending rates dashboards.

* Only enable the `FF_SMS_RATELIMIT` feature flag in staging environment; disabled in production.
* Make sure that older path (`FF_SMS_RATELIMIT` feature flag disabled) introduced no regression.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -57,7 +57,7 @@
                 "worker",
                 "--pidfile",
                 "/tmp/celery.pid",
-                "--concurrency=4",
+                "--concurrency=1",
                 "-l",
                 "DEBUG",
                 "-Q",
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -29,7 +29,8 @@
             "module": "flask",
             "env": {
                 "FLASK_APP": "application.py",
-                "FLASK_ENV": "development"
+                "FLASK_ENV": "development",
+                "FF_SMS_RATELIMIT": "true"
             },
             "args": [
                 "run",
@@ -47,6 +48,9 @@
             "module": "celery",
             "console": "integratedTerminal",
             "justMyCode": false,
+            "env": {
+                "FF_SMS_RATELIMIT": "true"
+            },
             "args": [
                 "--app",
                 "run_celery",
@@ -57,7 +61,7 @@
                 "-l",
                 "DEBUG",
                 "-Q",
-                "database-tasks,-priority-database-tasks.fifo,-normal-database-tasks,-bulk-database-tasks,job-tasks,notify-internal-tasks,periodic-tasks,priority-tasks,normal-tasks,bulk-tasks,reporting-tasks,research-mode-tasks,retry-tasks,service-callbacks-retry,send-sms-tasks,send-sms-high,send-sms-medium,send-sms-low,send-throttled-sms-tasks,send-email-high,send-email-medium,send-email-low,send-email-tasks,service-callbacks,delivery-receipts",
+                "database-tasks,-priority-database-tasks.fifo,-normal-database-tasks,-bulk-database-tasks,job-tasks,notify-internal-tasks,periodic-tasks,priority-tasks,normal-tasks,bulk-tasks,reporting-tasks,research-mode-tasks,retry-tasks,service-callbacks-retry,send-sms-tasks,send-sms-high,send-sms-medium,send-sms-low,send-throttled-sms-tasks,send-sms-fair-tasks,send-email-high,send-email-medium,send-email-low,send-email-tasks,service-callbacks,delivery-receipts",
             ]
         },
         {
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -62,6 +62,7 @@
                 "DEBUG",
                 "-Q",
                 "database-tasks,-priority-database-tasks.fifo,-normal-database-tasks,-bulk-database-tasks,job-tasks,notify-internal-tasks,periodic-tasks,priority-tasks,normal-tasks,bulk-tasks,reporting-tasks,research-mode-tasks,retry-tasks,service-callbacks-retry,send-sms-tasks,send-sms-high,send-sms-medium,send-sms-low,send-throttled-sms-tasks,send-sms-fair-tasks,send-email-high,send-email-medium,send-email-low,send-email-tasks,service-callbacks,delivery-receipts",
+                "--beat",
             ]
         },
         {
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -57,7 +57,7 @@
                 "-l",
                 "DEBUG",
                 "-Q",
-                "database-tasks,-priority-database-tasks.fifo,-normal-database-tasks,-bulk-database-tasks,job-tasks,notify-internal-tasks,periodic-tasks,priority-tasks,normal-tasks,bulk-tasks,reporting-tasks,research-mode-tasks,retry-tasks,service-callbacks-retry,send-sms-tasks,send-sms-high,send-sms-medium,send-sms-low,send-throttled-sms-tasks,send-sms-fair-tasks,send-email-high,send-email-medium,send-email-low,send-email-tasks,service-callbacks,delivery-receipts",
+                "database-tasks,-priority-database-tasks.fifo,-normal-database-tasks,-bulk-database-tasks,job-tasks,notify-internal-tasks,periodic-tasks,priority-tasks,normal-tasks,bulk-tasks,reporting-tasks,research-mode-tasks,retry-tasks,service-callbacks-retry,send-sms-tasks,send-sms-high,send-sms-medium,send-sms-low,send-throttled-sms-tasks,send-sms-fair,send-email-high,send-email-medium,send-email-low,send-email-tasks,service-callbacks,delivery-receipts",
                 "--beat",
             ]
         },
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -29,8 +29,7 @@
             "module": "flask",
             "env": {
                 "FLASK_APP": "application.py",
-                "FLASK_ENV": "development",
-                "FF_SMS_RATELIMIT": "true"
+                "FLASK_ENV": "development"
             },
             "args": [
                 "run",
@@ -48,9 +47,6 @@
             "module": "celery",
             "console": "integratedTerminal",
             "justMyCode": false,
-            "env": {
-                "FF_SMS_RATELIMIT": "true"
-            },
             "args": [
                 "--app",
                 "run_celery",
--- a/app/celery/celery.py
+++ b/app/celery/celery.py
@@ -18,6 +18,10 @@ def make_task(app):
         abstract = True
         start = None
 
+        @property
+        def message_group_id(self):
+            return self.request.get("notify_message_group_id")
+
         def on_success(self, retval, task_id, args, kwargs):
             elapsed_time = time.time() - self.start
             app.logger.info("{task_name} took {time}s".format(task_name=self.name, time="{0:.4f}".format(elapsed_time)))
@@ -32,6 +36,13 @@ def make_task(app):
 
 
 class NotifyCelery(Celery):
+    def send_task(self, name, args=None, kwargs=None, **options):
+        options["headers"] = options.get("headers") or {}
+        message_group_id = options.get("MessageGroupId")
+        if message_group_id:
+            options["headers"]["notify_message_group_id"] = message_group_id
+        return super().send_task(name, args, kwargs, **options)
+
     def init_app(self, app):
         super().__init__(
             app.import_name,
--- a/app/celery/celery.py
+++ b/app/celery/celery.py
@@ -38,9 +38,10 @@ def make_task(app):
 class NotifyCelery(Celery):
     def send_task(self, name, args=None, kwargs=None, **options):
         options["headers"] = options.get("headers") or {}
-        message_group_id = options.get("MessageGroupId")
+        message_group_id = options.pop("MessageGroupId", None)
         if message_group_id:
-            options["headers"]["notify_message_group_id"] = message_group_id
+            # Convert UUID to string for SQS compatibility
+            options["headers"]["notify_message_group_id"] = str(message_group_id)
         return super().send_task(name, args, kwargs, **options)
 
     def init_app(self, app):
--- a/app/celery/celery.py
+++ b/app/celery/celery.py
@@ -40,8 +40,15 @@ class NotifyCelery(Celery):
         options["headers"] = options.get("headers") or {}
         message_group_id = options.pop("MessageGroupId", None)
         if message_group_id:
-            # Convert UUID to string for SQS compatibility
-            options["headers"]["notify_message_group_id"] = str(message_group_id)
+            message_group_id = str(message_group_id)
+            # Store in Celery task headers so tasks can read it via self.message_group_id
+            # (used by deliver_sms_rate_limited to re-queue with the same group).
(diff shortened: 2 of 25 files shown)


.vscode/launch.json
```diff
@@ -47,17 +48,21 @@
...
+                "FF_SMS_RATELIMIT": "true"
+            },
             "args": [
                 "--app",
                 "run_celery",
                 "worker",
                 "--pidfile",
                 "/tmp/celery.pid",
-                "--concurrency=4",
+                "--concurrency=1",
```
jimleroyer, line 56: Most often, we don't need 4 celery workers when debugging (unless for a concurrency issues). While a lower number of concurrent workers generally makes it easier to debug our issues.

.vscode/launch.json
```diff
@@ -47,17 +48,21 @@
...
                 "-l",
                 "DEBUG",
                 "-Q",
-                "database-tasks,-priority-database-tasks.fifo,-normal-database-tasks,-bulk-database-tasks,job-tasks,notify-internal-tasks,periodic-tasks,priority-tasks,normal-tasks,bulk-tasks,reporting-tasks,research-mode-tasks,retry-tasks,service-callbacks-retry,send-sms-tasks,send-sms-high,send-sms-medium,send-sms-low,send-throttled-sms-tasks,send-email-high,send-email-medium,send-email-low,send-email-tasks,service-callbacks,delivery-receipts",
+                "database-tasks,-priority-database-tasks.fifo,-normal-database-tasks,-bulk-database-tasks,job-tasks,notify-internal-tasks,periodic-tasks,priority-tasks,normal-tasks,bulk-tasks,reporting-tasks,research-mode-tasks,retry-tasks,service-callbacks-retry,send-sms-tasks,send-sms-high,send-sms-medium,send-sms-low,send-throttled-sms-tasks,send-sms-fair-tasks,send-email-high,send-email-medium,send-email-low,send-email-tasks,service-callbacks,delivery-receipts",
+                "--beat",
```
jimleroyer, line 61: We were missing the beat worker and scheduling tasks with the previous config.

poetry.lock
jimleroyer, line 1: The lock file was generated 2+ weeks ago.

app/celery/celery.py
```diff
@@ -32,6 +36,21 @@ def __call__(self, *args, **kwargs):
 
 
 class NotifyCelery(Celery):
+    def send_task(self, name, args=None, kwargs=None, **options):
+        options["headers"] = options.get("headers") or {}
+        message_group_id = options.pop("MessageGroupId", None)
+        if message_group_id:
```
jimleroyer, line 42: We could gate the MessageGroupId / fair queue feature by adding the the feature flag verification on this condition. At the moment, the service ID will be passed by default as a MessageGroupId and services will be discriminated in normal queues. If we don't want this new behavior by default, we can control it here.

.vscode/launch.json
andrewleith: Could changing this to 1 hide concurrency issues though that we could otherwise find locally?

app/rate_limiter.py
```diff
@@ -0,0 +1,438 @@
...
+    import logging
+
+    global _rate_limiter_instance
+
+    logger = logging.getLogger(__name__)
+
+    # Determine which backend to use
+    backend = "memory"  # default
+
+    if use_redis is not None:
```
andrewleith, line 212: Shouldn't using redis be based on `SMS_RATE_LIMITER_BACKEND`? This logic is hard to follow.

.vscode/launch.json
```diff
@@ -47,17 +48,21 @@
             "module": "celery",
             "console": "integratedTerminal",
             "justMyCode": false,
+            "env": {
+                "FF_SMS_RATELIMIT": "true"
```
andrewleith: We don't normally set ENV vals here in launch.json.  Generally its done in `.env` and `config.py` - are we establishing a new place for these with this?

app/celery/celery.py
andrewleith: Yeah do we want to just flip this behavior on here when we merge this?  If it doesn't behave as expected, we would need to revert.  Perhaps we could gate it with the main FF or create a different FF if we want to turn this part on sooner than the rest.

app/rate_limiter.py
```diff
@@ -0,0 +1,438 @@
...
+        """
+        Get the current parts count in the active window.
+
+        Returns:
+            int: Number of parts used in the current window.
+        """
+        pass
+
+
+class InMemoryRateLimiter(RateLimiter):
```
andrewleith, line 52: So is the plan to run only a single pod with the limiter? If so, would a single pod even be capable of sending enough to warrant a limiter?

andrewleith (commented):
These changes look good in general to me. In addition to the rate limit I am happy to see the inclusion of fair queuing, it will be great not to allow big senders to monopolize!  After we sync I think we should be good to get this merged in!

app/notifications/process_notifications.py
```diff
@@ -210,18 +210,38 @@ def transform_notification(
...
+    celery_params: list = [str(notification.id)]
+    if (
+        notification.notification_type == SMS_TYPE
+        # We don't want to send test notifications to the rate limited queue.
+        and notification.queue_name != QueueNames.RESEARCH_MODE
+        and current_app.config.get("FF_SMS_RATELIMIT")
+    ):
+        queue = QueueNames.SEND_SMS_FAIR
+        # Prioritize sending per the queue names derived from the template category.
+        message_group_id = get_delivery_queue_for_template(notification.template)
```
andrewleith, line 234: Why the switch from per-service to per-priority here?

.vscode/launch.json
jimleroyer: Yes, feel free to modify to what you need when working locally. To me, `1` is easier to work with, because the celery scheduler can retry a notification that is currently being debugged and which takes several minutes to do so ( and hence creating a concurrent notification processing due to the long debugging nature of the activity).

.vscode/launch.json
jimleroyer: Not meant to be a new place to define these, but handy to not affect other local runs for other devs (meant for debugging).

app/celery/celery.py
jimleroyer: I can gate it with the main feature flag for now.

app/config.py
```diff
@@ -83,6 +83,10 @@ class QueueNames(object):
     # we have a limit to send per second and hence, needs to be throttled.
     SEND_THROTTLED_SMS = "send-throttled-sms-tasks"
 
+    # Fair queue for sending SMS, used to rate limit sending SMS from Notify to
+    # our SMS provider to avoid hitting downstream rate limits.
```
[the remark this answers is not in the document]
jimleroyer: Ok, let me review, we got 2 naming conventions in the way, but without the `-tasks` prefix, it would be better aligned with existing SMS priority queue / lanes.

app/notifications/process_notifications.py
```diff
@@ -210,18 +210,38 @@ def transform_notification(
...
+    queue: str = notification.queue_name
+    celery_params: list = [str(notification.id)]
+    if (
+        notification.notification_type == SMS_TYPE
+        # We don't want to send test notifications to the rate limited queue.
+        and notification.queue_name != QueueNames.RESEARCH_MODE
+        and current_app.config.get("FF_SMS_RATELIMIT")
+    ):
+        queue = QueueNames.SEND_SMS_FAIR
+        # Prioritize sending per the queue names derived from the template category.
```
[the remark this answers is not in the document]
jimleroyer: I will review that, not sure that applies.

app/notifications/process_notifications.py
jimleroyer: With these changes, by default, we use the service_id when not using the new sms fair queue, otherwise we prioritize by priority lane. If we don't pass in the priority lane for the new sms fair queue, we don't have priority lanes anymore, and only prioritize by volume per services. 

One refinement we might be able to do is to have both priority lanes and service ids appended into the message group ID. I will think about it.

.vscode/launch.json
andrewleith: Makes sense - should you remove these changes then? I think checking them in will affect other devs?

app/notifications/process_notifications.py
andrewleith: Yeah I wonder if under the current paradigm big senders still bully their way over everyone else?

andrewleith (approved):
LGTM - approved but waiting on a couple of additional changes flagged by copilot.

app/notifications/process_notifications.py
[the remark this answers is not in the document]
jimleroyer: @andrewleith I've actually fixed that one, along with other bugs around research mode which should have higher priority on other delivery modes.

app/notifications/process_notifications.py
```diff
@@ -231,30 +251,49 @@ def db_save_and_send_notification(notification: Notification):
     )
```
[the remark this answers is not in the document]
jimleroyer: Submitted a fix..

app/notifications/process_notifications.py
jimleroyer: Yes when using the fair queue. Let's try to implement the service ID discrimination in the fair queues as well in later changes.

app/notifications/process_notifications.py
```diff
@@ -267,24 +306,42 @@ def send_notification_to_queue(notification, research_mode, queue=None):
     if research_mode or notification.key_type == KEY_TYPE_TEST:
         queue = QueueNames.RESEARCH_MODE
 
+    # Pioritize sending per services within the same queue by default.
```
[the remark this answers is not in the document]
jimleroyer: Applying fix locally.

app/rate_limiter.py
jimleroyer: Yes and possibly not! We will run tests in staging to know how it behaves. The feature flag will be disabled in production.

app/rate_limiter.py
jimleroyer: SMS_RATE_LIMITER_BACKEND might be better outside of the module.. if we want it to be more isolated, modular and testable. But given this is module level logic, it's more of a utility to initialize the rate limiter with some sensitive defaults and overrides. We could talk on how we'd prefer to handle this in our code base in person (code review session).

PR #2846, opened 2026-04-23
Task/fix script build

jzbahrai:
# Summary | Résumé

1. Change the simulated data to use simulated emails/ phone numbers
2. Only allow terminal statuses in the notifications table
3. Allow other failure and delivered statuses in the notification_history table


whabanks (approved):
LGTM tested it out locally.

PR #2848, opened 2026-04-24
Add bulk API to get sms parts+costs for all services

whabanks:
# Summary | Résumé
This PR adds a new API endpoint to bulk fetch SMS parts + cost for all services. Previously we fetched these stats one service at a time by id, which resulted in massive slow downs on the platform admin service list page and unnecessary load on the DB. 

 - [ ] Test steps in the 

None.

### DIFF
--- a/app/dao/fact_billing_dao.py
+++ b/app/dao/fact_billing_dao.py
@@ -105,12 +105,12 @@ def dao_fetch_sms_cost_for_service_in_range(service_id, start_date, end_date):
 
 
 def dao_fetch_sms_cost_for_all_services_in_range(start_date, end_date):
-    """Return the total SMS cost and fragment count for ALL services in the given date range (inclusive).
+    """Return the total SMS cost and fragment count for ALL services in the given date range.
+    This EXCLUDES the current_day, as the ft_table aggregation happens in a nightly task
 
     Returns a dictionary keyed by service_id with fragment_count and total_cost for each service.
 
     Uses only FactBilling (the nightly-populated aggregate table) for efficiency.
-    Today's data will not be included since FactBilling is populated by a nightly task.
     For current-day data for a single service, use dao_fetch_sms_cost_for_service_in_range instead.
     """
     fact_results = (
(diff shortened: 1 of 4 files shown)


app/dao/fact_billing_dao.py
```diff
@@ -104,6 +104,87 @@ def dao_fetch_sms_cost_for_service_in_range(service_id, start_date, end_date):
...
+
+    Returns a dictionary keyed by service_id with fragment_count and total_cost for each service.
+
+    Uses the same logic as dao_fetch_sms_cost_for_service_in_range:
+    - FactBilling for historical data (up to yesterday)
+    - Notification table for current day data (with carrier-reported costs)
+
+    This is a bulk operation suitable for fetching costs for many services efficiently.
+    """
+    today = convert_utc_to_local_timezone(datetime.utcnow()).date()
```
jzbahrai: I think this will still be an expensive query depending on the data that exists in notifications.. i think you need to paginate by service.

This is an internal tool, so I dont want to spend too much time on this, but lets only use fact billin g data for historical range. 

So either they can get today's data? or they can get a range of data (which excludes today -> as it gets it from the aggregate table). This is me just testing against the really alrge service and see this as an issue. I also dont want to preoptimize too much for an internal tool

whabanks: We aren't paginating on the service list page currently, we can just exclude the current day's data from the all services fetch and later we can evaluate if there's any value including the current day's data somewhere else for platform admins.

app/dao/fact_billing_dao.py
```diff
@@ -104,6 +104,42 @@ def dao_fetch_sms_cost_for_service_in_range(service_id, start_date, end_date):
     return {"fragment_count": fragment_count, "total_cost": total_cost}
 
 
+def dao_fetch_sms_cost_for_all_services_in_range(start_date, end_date):
+    """Return the total SMS cost and fragment count for ALL services in the given date range (inclusive).
```
jzbahrai: This EXCLUDES current_day, as the ft_table aggregation happens once a day

^ as an edited comment

jzbahrai (approved):

PR #2851, opened 2026-04-28
Do not select all columns when just trying to do a count

ben851:
# Summary | Résumé

We're seeing long running queries (~4 seconds under load) on getting the count when doing /v2/notifications GET. SQL Alchemy appears to be selecting all columns just to do the count, which is wasteful. This should tell it to just get the count.

* https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/761

Re-run perf test and verify whether or not the performance is improved.

None.

### DIFF
--- a/app/v2/notifications/get_notifications.py
+++ b/app/v2/notifications/get_notifications.py
@@ -85,6 +85,7 @@ def get_notifications():
         client_reference=data.get("reference"),
         page_size=current_app.config.get("API_PAGE_SIZE"),
         include_jobs=data.get("include_jobs"),
+        count_pages=False,
     )
 
     def _build_links(notifications):


app/v2/notifications/get_notifications.py
```diff
@@ -85,6 +85,7 @@ def get_notifications():
         client_reference=data.get("reference"),
         page_size=current_app.config.get("API_PAGE_SIZE"),
         include_jobs=data.get("include_jobs"),
+        count_pages=False,
```
jzbahrai, line 88: this is get_notifications, so if this is used for getting the full notifications, editing this funciton might cause an issue?

jzbahrai (changes requested):

app/v2/notifications/get_notifications.py
ben851: For this specific endpoint, the response never includes a total count or page number... So the count was always computed and then silently discarded.

Unless I'm missing something?

jzbahrai: no you are right this should be ok

jzbahrai (approved):

PR #2854, opened 2026-04-29
Add permissions for ApiKeys

jzbahrai:
## Summary

Adds an independent `permissions` field to API keys so that template-management
access can be granted on a per-key basis, separately from the existing
`key_type` (`normal` / `team` / `test`).

https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/3139

No behaviour changes for existing API keys — the new column is nullable and
existing rows simply read back as `NULL` (treated as "no extra permissions").

## Changes

- **`app/models.py`**
  - New `ApiKeyPermission` `StrEnum` (values: `manage_templates`)
  - New constant `API_KEY_PERMISSION_TYPES` (frozenset of valid values)
  - New `permissions` column on `ApiKey` — `ARRAY(String)`, **nullable**, model-level `default=list` for new rows
  - New `ApiKey.has_permission(name)` helper (treats `None` as empty)
  - SQLAlchemy `@validates("permissions")` hook rejects unknown values and dedupes as a backstop
- **Alembic migration** (`migrations/versions/0510_api_key_permissions.py`)
  - `ALTER TABLE api_keys ADD COLUMN permissions varchar[] NULL`
  - Same column added to `api_keys_history` so the existing `Versioned` mixin captures changes for free

## Design notes

- **Why a column on `api_keys` and not a separate `api_key_permissions` table?**
  API keys are loaded on every authenticated request (hot path). Storing
  permissions as a Postgres array on the existing row means zero extra
  queries / JOINs. It also gets versioning for free via the existing
  `Versioned` mixin on `ApiKey`.
- **`StrEnum`** (Python 3.11+) is used so the enum members are real strings —
  they can be written straight to the DB column and JSON-serialised without
  `.value` boilerplate.

  | Enum | StrEnum
-- | -- | --
isinstance(member, str) | False | True
member == "foo" | False | True (if value is "foo")
JSON-serializable | No | Yes
Member values | Any type | Must be str
Available since | 3.4 | 3.11

## Testing

- [ ] Migration upgrade + downgrade run cleanly against a local DB
- [ ] Existing API key flows unchanged (sending notifications, listing keys, revoking keys)

## Related

- Issue: #3139

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1,7 +1,7 @@
 import datetime
 import itertools
 import uuid
-from enum import Enum
+from enum import Enum, StrEnum
 from typing import Any, Iterable, Literal
 
 from flask import current_app, url_for
@@ -25,10 +25,11 @@ from notifications_utils.timezones import (
     convert_utc_to_local_timezone,
 )
 from sqlalchemy import CheckConstraint, Index, UniqueConstraint
-from sqlalchemy.dialects.postgresql import JSON, JSONB, UUID
+from sqlalchemy.dialects.postgresql import ARRAY, JSON, JSONB, UUID
 from sqlalchemy.ext.associationproxy import association_proxy
 from sqlalchemy.ext.declarative import declared_attr
 from sqlalchemy.ext.hybrid import hybrid_property
+from sqlalchemy.orm import validates
 
 from app import (
     DATETIME_FORMAT,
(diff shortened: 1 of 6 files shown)


whabanks (approved):
LGTM. Existing and new API keys still work, can send notifications, revocations working, migrations both up and down are working.

PR #2859, opened 2026-05-06
Updating and expanding nightly perf test scripts

ben851:
# Summary | Résumé

These scripts/changes are the culmination of my redline perf testing. The base usage should not be affected, so the nightly perf tests should still run properly.

* https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/761

Verify that nightly perf tests still work

None.


tests_nightly_performance/src/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+    && pip install --no-cache-dir "locust>=2,<3" "python-dotenv>=1,<2" \
+    && adduser --disabled-password --gecos "" locust
+
+COPY --chown=locust:locust *.py ./
+
+USER locust
+
+EXPOSE 8089 5557 5558
+
+ENTRYPOINT ["locust"]
```
jimleroyer: Great addition.

tests_nightly_performance/src/blast_api.py
```diff
@@ -10,34 +111,61 @@
...
             "phone_number": Config.PHONE_NUMBER,
             "template_id": Config.SMS_TEMPLATE_ID_ONE_VAR,
             "personalisation": {"var": "single sms"},
         }
-        self.client.post("/v2/notifications/sms", json=json, headers=self.headers)
+        response = self.client.post("/v2/notifications/sms", json=json, headers=self.headers)
+        if response.status_code == 201:
+            try:
+                NotifyApiUser._sent_ids.append(response.json()["id"])
+            except Exception:
```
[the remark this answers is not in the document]
jimleroyer: A print statement would make sense at least, no? instead of swallowing exceptions. We could even store the failure into a `NotifyApiUser._failed_ids` so we could track these later if we want to.

tests_nightly_performance/src/scaling_api.py
```diff
@@ -0,0 +1,325 @@
...
+
+
+def _push_id(notification_id: str) -> None:
+    with _ids_lock:
+        _notification_ids.append(notification_id)
+
+
+def _sample_id() -> "str | None":
+    """Return a random notification ID from the pool without removing it."""
+    with _ids_lock:
```
jimleroyer: Do we really need to block the lock for a read?

tests_nightly_performance/src/scaling_api.py
```diff
@@ -0,0 +1,325 @@
...
+#   9-12 min →  70 users
+#   …
+#   ~42 min  → 300 users  (--max-users reached → test ends)
+#
+# All parameters are tunable at runtime via CLI flags or env vars.
+# Pass --max-users 0 to run until stopped manually.
+# ---------------------------------------------------------------------------
+
+
+class StepwiseLoadShape(LoadTestShape):
```
jimleroyer: The comments above should be the class doc under the class definition.

tests_nightly_performance/src/blast_api.py
```diff
@@ -1,7 +1,130 @@
...
+        "--pacing",
+        type=float,
+        default=60.0,
+        help="Seconds between tasks per user using constant_pacing (default: 60). Ignored if --wait-min/--wait-max are set.",
+    )
+    parser.add_argument(
+        "--wait-min",
+        type=float,
+        default=None,
+        help="Minimum wait seconds between tasks (enables between() mode instead of constant_pacing)",
```
jimleroyer: Can we add some documentation in the CLI to describe that choice of between or constant pacing? or an explicit `--mode` parameter?

jimleroyer (commented):
Fantastic work. I'd like to sync in a meet on the few comments I left, that will probably go faster.

tests_nightly_performance/src/scaling_api.py
jimleroyer: After some back and forth with the AI, that could be summarized to:

> Yes — with those assumptions, I agree: the lock is probably not needed.
>
> Your argument is basically:
>
> * _sample_id() does not require a consistent snapshot
> * it only needs some valid notification ID from the deque
> * the deque is append-only from the caller’s perspective
> * once full, entries may shift, but shifting is acceptable because selection is already random
> * reading from a slightly stale view of the deque does not violate the test’s intent
>
> A polished reviewer comment could be:
>> I think we can drop the lock in _sample_id(). This method doesn’t require a consistent snapshot of the deque; it only needs to return any valid recent notification ID. Since IDs are sampled randomly anyway, concurrent appends/evictions in the bounded deque only affect which valid ID is returned, not the behavior we care about in this performance test.

But it also noted that with the network being the most delay in this logic, that slight improvement won't matter too much, as this would be a negligible performance cost to keep the lock in this context.

tests_nightly_performance/src/blast_api.py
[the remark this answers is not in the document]
ben851: We would end up with hundreds of these I would think... I can add it I guess..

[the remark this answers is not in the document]
jimleroyer: Talked separately with Ben: we agreed to keep the IDs and print the overall IDs in one print statement when tests are done.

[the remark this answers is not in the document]
ben851: Looking at this more - we don't actually have the Ids at this point (that's what the exception is catching). Instead, I've added a counter for each one that gets output at the end of the tests so we can track how many were missing.

tests_nightly_performance/src/scaling_api.py
ben851: Removed

tests_nightly_performance/src/scaling_api.py
ben851: Fixed

jimleroyer (approved):
LGTM

PR #2868, opened 2026-05-20
Fix/ratelimit nonetype on empty template

jimleroyer:
# Summary | Résumé

Fixes an `AttributeError: 'NoneType' object has no attribute 'template_type'` crash
in `send_notification_to_queue` when `FF_SMS_RATELIMIT=True` on the batch SMS path
(`save_smss` → `persist_notifications`).

**Root cause**: `persist_notifications` uses `bulk_save_objects`, which bypasses
SQLAlchemy's session identity map. The returned `Notification` objects are not
session-tracked, so the `template` lazy relationship stays `None`. The
`FF_SMS_RATELIMIT` code path in `send_notification_to_queue` was the first place
to access `notification.template` after a batch save.

**Changes**:

* Eagerly sets `notification_obj.template` in `persist_notifications` — the template
  is already fetched there, so this is zero-cost.
* Also calculates `billable_units` in `persist_notifications` when `FF_USE_BILLABLE_UNITS`
  is enabled, mirroring what `persist_notification` (single-send path) already does.
  This makes the `FF_USE_BILLABLE_UNITS` shortcut in `send_notification_to_queue`
  actually reachable for batch sends.
* Defensively falls back to `dao_get_template_by_id` in `send_notification_to_queue`
  if `notification.template` is still `None`, covering any future callers that don't
  go through `persist_notifications`.

* [SMS Rate Limiter](https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/786)
#786

* Send SMS through API.

None.


P0NDER0SA (approved):
Looks good to me!

ben851:
Reviewed this - I think it makes sense but I'd like someone from app to review.

app/notifications/process_notifications.py
```diff
@@ -320,14 +320,17 @@ def send_notification_to_queue(notification, research_mode, queue=None):
             queue = QueueNames.SEND_THROTTLED_SMS
         elif current_app.config.get("FF_SMS_RATELIMIT") and queue != QueueNames.RESEARCH_MODE:
             deliver_task = provider_tasks.deliver_sms_rate_limited
+            template = notification.template or dao_get_template_by_id(
```
jzbahrai, line 323: this is so interesting! I didnt realize this is why it was happening cause we have had peristent issues in the past while fetching notifications/ templates. good find

jzbahrai (approved):

app/notifications/process_notifications.py
```diff
@@ -320,14 +320,17 @@ def send_notification_to_queue(notification, research_mode, queue=None):
...
+                notification.template_id, notification.template_version, use_cache=True
```
whabanks, line 324: Good call using the cache here 🚀. 
Initially I wondered if there could be a template version mismatch if an update happened recently and the cache wasn't refreshed, but it looks like we [refresh cache at update time](https://github.com/cds-snc/notification-admin/blob/main/app/notify_client/service_api_client.py#L250) so all good!

whabanks (approved):

app/notifications/process_notifications.py
```diff
@@ -320,14 +320,17 @@ def send_notification_to_queue(notification, research_mode, queue=None):
...
+            )
```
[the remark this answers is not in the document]
jimleroyer: We have strong assumptions about these values being set. The suggestion isn't bad , but highly defensive and noisy.

app/notifications/process_notifications.py
jimleroyer: Hmm yeah, this is a bit tricky with this part of the code, as the `notification` object isn't managed by the sqlalchemy ORM. So the `template` property won't get loaded automatically. We have to manually leave that reference for downstream consumers. Hopefully that should make this a bit more stable.

app/notifications/process_notifications.py
jimleroyer: Yeah, this is defensive to avoid hitting the db too much with the same call. The template should be populated upstream by `persist_notification/s` functions, so it might never be used, but just in case..

PR #2878, opened 2026-05-27
Add GET templates under v3

jzbahrai:
# Summary | Résumé

Added a GET template method. This method currently exists, but we want to gate it behind APIKEYs with manage_template permissions.

1. Test getting a template locally 
2. If you aren't use a the new API key with a manage template permission, you should get a forbidden error
3. Otherwise the template should return
4. Test both scenarios

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -314,6 +314,12 @@ def register_v2_blueprints(application):
         v2_template_blueprint,
     )
     from app.v2.templates.get_templates import v2_templates_blueprint as get_templates
+    from app.v3.template import (  # noqa
+        get_template as v3_get_template,
+    )
+    from app.v3.template import (
+        v3_template_blueprint,
+    )
 
     register_notify_blueprint(application, v2_notification_blueprint, requires_auth)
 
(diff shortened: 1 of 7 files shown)


app/__init__.py
```diff
@@ -314,6 +314,12 @@ def register_v2_blueprints(application):
         v2_template_blueprint,
     )
     from app.v2.templates.get_templates import v2_templates_blueprint as get_templates
+    from app.v3.template import (  # noqa
+        get_template as v3_get_template,
+    )
+    from app.v3.template import (
```
jzbahrai, line 320: the formatter `make format` is forcing this :/

app/v3/template/__init__.py
smcmurtry, line 1: I think we should move this code under `v2/template`. I believe the concept with doing versioning like v2, v3, is that you can make breaking changes to the api interface for `v2/template` by creating `v3/template`, and then you don't break your users' existing integrations. And then you would tell users to move to v3 and eventually turn off the v2 endpoints.
Are there any breaking changes here? Could we just make some updates to our existing `v2/template` endpoint? It look like we are still allowing GET requests like `v2/template/<template_id>`

smcmurtry (changes requested):

PR #2885, opened 2026-06-02
Move some technical-failure statuses to be permanent failures

smcmurtry:
# Summary | Résumé

We have been assigning a status of `technical-failure` for some cases where we are actually getting billed by AWS. And we had permanent-failure for a case that was not billable (sms recipient opted out). This is a problem for our billing calculations because `technical-failure` is not a billable status and permanent-failure is (https://github.com/cds-snc/notification-api/blob/main/app/models.py#L1737).

This PR makes the following changes to notification statuses for sms:
- if a user has opted out of sms messages, treat this as a technical-failure (this is not a billable case)
- add handling for "The delivery TTL has expired" -> permanent-failure (this is billable)
- if we don't handle the provider response, default to permanent-failure since we have already passed this notification to AWS and we will likely be billed for it

* https://github.com/cds-snc/notification-planning/issues/3274

Tests should pass.

None.

### DIFF
--- a/app/celery/process_sns_receipts_tasks.py
+++ b/app/celery/process_sns_receipts_tasks.py
@@ -141,6 +141,6 @@ def determine_status(sns_status, provider_response):
     if not status:
         # TODO: Pattern matching in Python 3.10 should simplify this overall function logic.
         if "is opted out" in provider_response:
-            return NOTIFICATION_PERMANENT_FAILURE
+            return NOTIFICATION_TECHNICAL_FAILURE
 
     return status
--- a/app/celery/process_sns_receipts_tasks.py
+++ b/app/celery/process_sns_receipts_tasks.py
@@ -12,6 +12,7 @@ from app.models import (
     NOTIFICATION_DELIVERED,
     NOTIFICATION_PERMANENT_FAILURE,
     NOTIFICATION_SENT,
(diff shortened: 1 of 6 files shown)


app/celery/process_sns_receipts_tasks.py
```diff
@@ -131,7 +130,7 @@ def determine_status(sns_status, provider_response):
         "Phone has blocked SMS": NOTIFICATION_TEMPORARY_FAILURE,
         "Phone is on a blocked list": NOTIFICATION_TEMPORARY_FAILURE,
         "Phone is currently unreachable/unavailable": NOTIFICATION_PERMANENT_FAILURE,
-        "Phone number is opted out": NOTIFICATION_TECHNICAL_FAILURE,
+        "Phone number is opted out": NOTIFICATION_PERMANENT_FAILURE,
```
jzbahrai: lets check we get charged for this

jzbahrai: and adding delivery status ttl as a permanent failure

smcmurtry: > lets check we get charged for this

Looks like we don't get charged for this case so I'll keep this as tech failure

jzbahrai (approved):

PR #2900, opened 2026-06-10
Run tests with more parallelization

jzbahrai:
# Summary | Résumé

Tests currently take upwards of 9 mins. Run tests to use further parallelization


tests/conftest.py
```diff
@@ -78,14 +78,14 @@ def grant_test_db(writer_uri, uri_db_reader):
     postgres_db = sqlalchemy.create_engine(writer_uri, echo=False, isolation_level="AUTOCOMMIT", client_encoding="utf8")
 
     statements = [
-        f"CREATE ROLE {db_reader} LOGIN PASSWORD '{db_reader_password}';",
+        f"DO $$ BEGIN CREATE ROLE {db_reader} LOGIN PASSWORD '{db_reader_password}'; EXCEPTION WHEN duplicate_object OR unique_violation THEN ALTER ROLE {db_reader} WITH LOGIN PASSWORD '{db_reader_password}'; END $$;",
```
jzbahrai, line 81: when parallelizing the tests, we are creating the same role on the DB, hence we needed this in case the role was already created

pytest.ini
```diff
@@ -19,6 +19,7 @@ env =
     D:FF_SALESFORCE_CONTACT=True
     D:FF_CLOUDWATCH_METRICS_ENABLED=True
     D:REDIS_URL=redis://localhost:6380
+    D:REDIS_ENABLED=True
```
jzbahrai, line 22: tests assume we have redis on in most cases, this makes it explicit for when it was not

migrations/versions/0451_create_db_users.py
```diff
@@ -32,7 +32,7 @@ def create_role_if_not_exist(role):
         DO $$
         BEGIN
         CREATE ROLE {role};
-        EXCEPTION WHEN duplicate_object THEN RAISE NOTICE '%, skipping', SQLERRM USING ERRCODE = SQLSTATE;
+        EXCEPTION WHEN duplicate_object OR unique_violation THEN RAISE NOTICE '%, skipping', SQLERRM;
```
jzbahrai, line 35: this role will not run on production, its just that when we are setting up the test DB and running tests in parallel the role might already exist. This fix was made to allow for the parallel tests to run

tests/app/test_queue.py
```diff
@@ -428,6 +429,14 @@ def test_put_batch_saving_expiry_metric(self, redis, redis_queue, mocker):
             redis_queue.poll(10)
             redis_queue.poll(10)
             redis_queue.poll(10)
-            time.sleep(2)
-            redis_queue.expire_inflights()
-            assert pbsem_mock.assert_called_with(mock.ANY, mock.ANY, 3) is None
+            # Expiry uses Redis idletime (`idle > expire_after`) with second-level granularity,
```
jzbahrai, line 432: The failing test expects one exact call:
Expected count 3
Actual count 1

expiry metric count is len(expired) per expire_inflights() call.
Redis expiry logic uses idle > expire_after with second-level timing, so three inflights may expire across multiple passes, not always all in one pass.

smcmurtry (approved):
Tests are running much faster now! Code looks good. It is a bit of a tradeoff in terms of complexity of the workflows, but probably worth for the amount of dev time we will save.

PR #2935, opened 2026-07-07
Changed the product type to be a list

jzbahrai:
# Summary | Résumé

Airtable column heading had a mismatch from what we were sending through the API. The model sent a Single text field, but the Product field was a drop down. Made the change to send Notify as a value for the drop down


andrewleith (approved):
LGTM

PR #2933, opened 2026-07-06
Delete template attachment from scan-files S3 bucket post-verdict receival

whabanks:
# Summary | Résumé

Once the GuardDuty scan on a template attachment is complete, we have received the scan verdict and updated the file status in the DB then we'll remove the file from the scan-files S3 bucket

# Test instructions | Instructions pour tester la modification
- To be tested in staging once the front and back end are hooked up and we can do an E2E test

# Release Instructions | Instructions pour le déploiement

None.

# Reviewer checklist | Liste de vérification du réviseur

- [ ] This PR does not break existing functionality.
- [ ] This PR does not violate GCNotify's privacy policies.
- [ ] This PR does not raise new security concerns. Refer to our GC Notify Risk Register document on our Google drive.
- [ ] This PR does not significantly alter performance.
- [ ] Additional required documentation resulting of these changes is covered (such as the README, setup instructions, a related ADR or the technical documentation).

> ⚠ If boxes cannot be checked off before merging the PR, they should be moved to the "Release Instructions" section with appropriate steps required to verify before release. For example, changes to celery code may require tests on staging to verify that performance has not been affected.

PR #2937, opened 2026-07-08
Revert "Delete template attachment from scan-files S3 bucket post-verdict receival"

whabanks:
Reverts cds-snc/notification-api#2933

Deleting the file from the scan files bucket too early results in DDAPI failing during [`check_scan_verdict`](https://github.com/cds-snc/notification-document-download-api/blob/main/app/download/views.py#L47) when attempting to download the template attachment.

```
"Failed to get tags from document: {'Code': 'NoSuchKey', 'Message': 'The specified key does not exist.', 'Key': 'notification-canada-ca-staging-document-download-scan-files/template_attachments/70c2d41d-c085-4a10-81dd-17ecf548e389/fd312455-a6c3-400d-839f-76e58db3aa85'}
```

### DIFF
--- a/app/files/rest.py
+++ b/app/files/rest.py
@@ -173,12 +173,6 @@ def update_file_status():
     fetched_file.status = new_status
     dao_update_file(fetched_file)
 
-    try:
-        document_download_client.delete_document(service_id, document_id, "template_attach")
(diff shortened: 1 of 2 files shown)


whabanks:
~Closing because we're updating how this will work in DDAPI instead of here.~
We totally know what we're doing...

andrewleith (approved):
🚀

PR #2943, opened 2026-07-09
Folder session bug

jzbahrai:
## PR Summary

**Bug:** `POST /v2/manage-template` with `parent_folder_id` fails with `ForeignKeyViolation` on `template_folder_map_template_id_fkey`.

**Root Cause:** Setting `folder` during `Template.from_json()` causes SQLAlchemy's folder backref to implicitly add the template to the session. When `authenticated_service.permissions` is subsequently lazy-loaded, it triggers an autoflush that INSERTs the template (with an auto-generated PK) and the `template_folder_map` row. Then `dao_create_template` reassigns `template.id = uuid.uuid4()`, generating an `UPDATE` to change the PK — which violates the FK constraint from `template_folder_map`.

This never affected the v1/admin endpoint because it pre-loads `permissions` *before* creating the template (the v1 code even has a comment warning about this: *"permissions needs to be placed here otherwise marshmallow will interfere with versioning"*).

**Fix:** Don't pass `folder` to `Template.from_json()`. Instead, pass it to `dao_create_template()` which sets `template.folder` *after* assigning the id — ensuring the template enters the session with the correct, final PK.

**Changes:**
- templates_dao.py — Added optional `folder` param to `dao_create_template`; folder is set after `template.id` is assigned.
- post_template.py — Pass `folder` to `dao_create_template(template, folder=folder)` instead of `Template.from_json()`.
- test_post_template.py — Added tests for create-with-folder, invalid folder, and no-folder cases.

## Test:

1. `main` branch
2. curl --request POST \
  --url http://localhost:6011/v2/manage-template \
  --header 'Authorization:' \
  --header 'Content-Type: application/json' \
  --header 'User-Agent: insomnia/12.6.0' \
  --data '{"name": "Welcome email3",
 "template_type": "email",
 "subject": "Welcome to  notify 3",
 "content": "Hi ((first_name)), welcome!", "template_category_id": "7c16aa95-e2e1-4497-81d6-04c656520fe4",
 "parent_folder_id": "e574f8bb-59c7-4b1f-8b2e-de7ad61e002a"
}'

### you will need to update the parent_folder_id with a folder_id in your service
1. you will see an error similar to this:
```
sqlalchemy.exc.IntegrityError: (raised as a result of Query-invoked autoflush; consider using a session.no_autoflush block if this flush is occurring prematurely)
(psycopg2.errors.ForeignKeyViolation) update or delete on table "templates" violates foreign key constraint "template_folder_map_template_id_fkey" on table "template_folder_map"
DETAIL:  Key (id)=(1b7385fd-8d9b-4930-982e-b705e3ac015d) is still referenced from table "template_folder_map".
```

1. Switch to this branch
1. Send the same curl request. No error

### DIFF
--- a/app/dao/templates_dao.py
+++ b/app/dao/templates_dao.py
@@ -22,9 +22,16 @@ from app.models import (
 
 @transactional
 @version_class(VersionOptions(Template, history_class=TemplateHistory))
-def dao_create_template(template, redact_personalisation=False):
+def dao_create_template(template, redact_personalisation=False, folder=None):
     # must be set now so version history model can use same id
     template.id = uuid.uuid4()
+
+    # Set folder AFTER id is assigned. Setting it before (e.g. in from_json) causes
+    # the folder backref to add the template to the session prematurely; any subsequent
+    # lazy-load query then triggers autoflush, persisting the template before this
+    # function can assign the id — leading to a PK-changing UPDATE that violates FKs.
+    if folder:
+        template.folder = folder
 
     redacted_dict = {
         "template": template,
--- a/app/v2/manage_template/post_template.py
+++ b/app/v2/manage_template/post_template.py
@@ -52,7 +52,6 @@ def post_manage_template():
             "service": authenticated_service.id,
             "created_by": api_user.created_by_id,
         },
-        folder,
     )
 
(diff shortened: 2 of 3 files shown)


smcmurtry (approved):
Tested locally, the fix works great!

PR #2947, opened 2026-07-13
Classify duplicate-record Celery errors and emit structured log tag

smcmurtry:
## What

Adds `DUPLICATE_RECORD` to `CeleryErrorCategory` and the error classifier, and updates `handle_batch_error_and_forward` to emit the structured tag `CELERY_KNOWN_ERROR::DUPLICATE_RECORD` in its warning log.

## Why

The stable-id fix (#2944) made SQS-redelivered batches hit a primary-key conflict on the `notifications` table instead of silently inserting duplicate rows. This is correct behaviour — `handle_batch_error_and_forward` catches the error, finds the existing notification, and skips the resend — but the resulting `IntegrityError` log was only distinguishable from other constraint violations by the raw Postgres error string.

With this change the log line looks like:
```
CELERY_KNOWN_ERROR::DUPLICATE_RECORD Batch saving: could not persist notifications with receipt <uuid>: ...duplicate key value violates unique constraint "notifications_pkey"...
```

This is required by a companion terraform PR which updates the `celery-error-duplicate-record` CloudWatch log filter to match the structured tag instead of the raw Postgres string, and adds a separate alarm for genuinely unexpected constraint violations.

## Testing

- `test_duplicate_record_by_message` — verifies `IntegrityError` wrapping a Postgres unique violation classifies as `DUPLICATE_RECORD`
- All existing classifier and batch-saving tests continue to pass

### DIFF
--- a/app/celery/error_registry.py
+++ b/app/celery/error_registry.py
@@ -40,6 +40,9 @@ class CeleryErrorCategory(str, Enum):
     # Notifications that timed out
     TIMEOUT_NOTIFICATION = "CELERY_KNOWN_ERROR::TIMEOUT_NOTIFICATION"
 
+    # Duplicate notification record — SQS redelivery or concurrent worker processed the same row
+    DUPLICATE_RECORD = "CELERY_KNOWN_ERROR::DUPLICATE_RECORD"
+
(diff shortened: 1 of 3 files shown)


smcmurtry:
Going with a simpler approach

PR #2951, opened 2026-07-15
Added bounce rate suspension

jzbahrai:
### Summary

When a service's email bounce rate exceeds the 10% critical threshold, we now take action based on how many emails the service has sent in the 24-hour sliding window:

- **≥ `BR_VOLUME_MINIMUM` emails sent**: Remove the service's `email` permission and notify service users that their email sending has been disabled.
- **< `BR_VOLUME_MINIMUM` emails sent**: Send a warning email — email sending continues but the service is told they may lose the ability to send if the rate persists.

This replaces the previous behaviour of only logging bounce rate status.

### How enforcement works

Instead of suspending the entire service (which would block SMS too), we remove the `email` permission from `service_permissions`. This is the same mechanism used by the Admin UI's "Send emails" toggle. The API's existing `check_service_has_permission` validator then blocks all subsequent email sends with `"Service is not allowed to send emails"`.

### Behaviour matrix

| Bounce rate | Emails sent (24h) | Action |
|---|---|---|
| Critical (≥10%) | ≥ `BR_VOLUME_MINIMUM` | Remove email permission + send suspension email |
| Critical (≥10%) | < `BR_VOLUME_MINIMUM` | Send warning email only |
| Warning (≥5%) | any | Log only |
| Normal | any | No action |

Redis cache keys with 24h TTL prevent duplicate emails within the same window.

> **Note:** Enforcement is currently gated behind `NOTIFY_ENVIRONMENT != "production"` for safe rollout.


app/delivery/send_to_providers.py
```diff
@@ -329,24 +329,6 @@ def check_for_malware_errors(document_download_response_code, notification):
         document_download_internal_error(notification=notification)
 
 
-def check_service_over_bounce_rate(service_id: str):
```
smcmurtry, line 332: I think we should leave these logs in place so that we can audit what is going on. But change the critical alarm in TF to be a warning. And then once we see this working we could remove the alarms in TF at that point.

app/delivery/bounce_rate.py
```diff
@@ -0,0 +1,66 @@
...
+
+def _bounce_rate_suspension_cache_key(service_id: str) -> str:
+    return f"bounce-rate-suspension-email-sent:{service_id}"
+
+
+def _bounce_rate_warning_cache_key(service_id: str) -> str:
+    return f"bounce-rate-warning-email-sent:{service_id}"
+
+
+def check_service_over_bounce_rate(service_id: str):
```
smcmurtry, line 21: Let's put either a feature flag or an environment check on this code so that it doesn't run in prod (`if environment == 'prod': return`). That way we can release it in a more controlled way if testing takes longer than 1 day.

smcmurtry (changes requested):
I ran the migrations locally and the new emails look good. Can you add an environment check add the logging back in?

app/delivery/bounce_rate.py
jzbahrai: good call added it here: https://github.com/cds-snc/notification-api/pull/2951/changes#diff-bb9bbd92a0b7432bc98f99fe96b5c95ca7119d353e4efba40adf98454297b6c5R41

app/delivery/send_to_providers.py
jzbahrai: yup it exists, just moved funcs: 
https://github.com/cds-snc/notification-api/pull/2951/changes#diff-bb9bbd92a0b7432bc98f99fe96b5c95ca7119d353e4efba40adf98454297b6c5R37

smcmurtry: Thanks I didn't notice that! Could you update the wording of the log though? The TF alarm looks for the words `critical bounce rate threshold of 10`, which is no longer present in the log : https://github.com/cds-snc/notification-terraform/blob/main/aws/eks/cloudwatch_log.tf#L250 

The warning log should be fine since that one looks for `warning bounce rate threshold of 5` which is still present.

smcmurtry (approved):
Looks good!

PR #2953, opened 2026-07-16
Validate total attachment size

whabanks:
# Summary | Résumé

This PR adds additional defensive validation to `files/rest.py::create_file` that ensures total file attachment size does not exceed 6MB. It takes the size of the newly requested file + existing file attachments into account. The `InvalidRequest` raised returns data to aid in constructing an appropriate error message on the front end.

```
{
  "error": "Total file size exceeds the 6MB limit.",
  "current_usage": existing_size,
  "requested": data["file_size"],
  "limit": MAX_TOTAL_FILE_SIZE,
}
```

Bug bash issue: https://docs.google.com/document/d/1m_EzXNJbqjJH-WLzxRKMtO4k5CAHwk51uCXi7xJQR44/edit?tab=t.0#task=NbqwOoo-dIWUUrb0

1. Hook your local up to staging
2. Add a few files to a template eventually adding one that would put you over the 6Mb limit
3. Note that the file isn't attached (currently nothing will happen, no error message yet as it needs to be caught in the front end and displayed)

None.

### DIFF
--- a/app/files/rest.py
+++ b/app/files/rest.py
@@ -51,6 +51,8 @@ GUARD_DUTY_STATUS_MAP = {
     "SKIPPED": FILE_STATUS_VIRUS_SCAN_FAILED,
 }
 
+MAX_TOTAL_FILE_SIZE = 6 * 1024 * 1024  # 6MB
+
 
 @files_blueprint.route("", methods=["POST"])
 def create_file(template_id):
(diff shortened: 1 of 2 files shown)


app/files/rest.py
```diff
@@ -72,13 +74,31 @@ def create_file(template_id):
...
+        raise InvalidRequest(
+            {
+                "error": "Total file size exceeds the 6MB limit.",
+                "current_usage": existing_size,
+                "requested": actual_file_size,
+                "limit": MAX_TOTAL_FILE_SIZE,
+            },
+            status_code=400,
+        )
+
```
andrewleith: This seems to be throwing an uncaught exception in API instead of return a 400 to ADMIN.

andrewleith: You could try something like:

```
return jsonify({"error": "over_file_limit"}), 400
```

We'll need to come up with a format there so ADMIN can read it on the other side.

whabanks: Updated! Let me know what you think

andrewleith: Looks good - I'll start an admin PR that can handle this payload and show the error.

andrewleith (approved):
Looks good!

Here is an ADMIN pr to handle this (and the other) errors API can throw: https://github.com/cds-snc/notification-admin/pull/2744

PR #2965, opened 2026-07-23
Task/fixes sus

jzbahrai:
# Summary | Résumé

Updated the utils code here + made fixes to the warning email


migrations/versions/0521_update_bounce_rate_warn.py
```diff
@@ -0,0 +1,139 @@
+"""
+Revision ID: 0520_update_bounce_rate_warn
```
smcmurtry:
```suggestion
Revision ID: 0521_update_bounce_rate_warn
```

smcmurtry (approved):
Looks good - just update the revision id before merging.

PR #2977, opened 2026-07-27
remove prod check

jzbahrai:
# Summary | Résumé

Remove the production check do this code will work on prod as well

## Test:

make a new service, so that the below numbers might work

You can test the below on staging

Case 1:
1. send 1000 successful emails 
2. Send 60 bounce emails
[100
[50bounce.csv](https://github.com/user-attachments/files/30434133/50bounce.csv)
0success.csv](https://github.com/user-attachments/files/30434122/1000success.csv)

3. Get an email indicating a warning message - your service might be suspended

Case 2:
1. Send 100 bounce emails
[100bounce.csv](https://github.com/user-attachments/files/30434134/100bounce.csv)
3. Get an email indicating that sending emails has been suspended
4. Go to the templates page - all your email templates will have disappeared

Case 3:
1. Go to service setting
2. Switch on email sending
3. you should be able to continue sending emails (even past the bounce rate)


smcmurtry:
I think I found a bug where no warning email is sent even though I have reached a 6.5% bounce rate - see Step 3 below.

Here is the test I did:
1. send 1000 success emails from this service: https://staging.notification.cdssandbox.xyz/services/0143906c-c6f4-4ddf-b538-43ed1a695f7d
2. send 50 bounce emails - no warning yet because my bounce rate is 4.8%
3. send 20 bounce emails - still no warning email but my bounce rate is now 6.5% which is over the 5% warning threshold
4. send 100 bounce emails - now I get both the warning email and the critical/suspended email
5. turn email sending back on and I can send to myself successfully

Also I noticed an unrelated bug - the annual email limit on the following page doesn't match what is on my service settings:
Screenshot 2026-07-28 at 12 04 27 PM

jzbahrai:
@smcmurtry 
The warning/critical check runs during send flow, but hard-bounce counts are updated asynchronously when SES receipts are processed. So it is possible for Redis to show a 6.5% bounce rate after receipts land, but the warning check did not run exactly at that moment. It only runs on subsequent sends, and with enough new bounce traffic it can move quickly from warning range into critical range, which is why you can end up seeing warning and suspension close together.

So in your steps:

4.8% after first batch makes sense
6.5% can appear in Redis before the next effective check point
next high-bounce sends trigger checks again, and by then we can hit both warning and critical paths close together
Also, once suspension triggers we intentionally set the warning key too, so we do not send a follow-up warning after a suspension event.

Given we  checked for bounce rate on sends anyways, I do not want to move this check to when we process receipts. I think this is good enough for a solution

smcmurtry:
That sounds okay in this case, but will this code stop the service from sending midway through a job if there is a really high failure rate? Let's say a service queues up a 50k row bulk job and 20k of those rows are hard bounces. I tried to run a test like that but for some reason the jobs are not starting and I'm now seeing `expired-inflight-queue-outnumber-email-bulk-warning` alarms in staging. Here is the csv file I'm using - 600 bounces, 1000 success, 600 bounces, 1000 success:

[3000_very_bad_emails.csv](https://github.com/user-attachments/files/30475728/3000_very_bad_emails.csv)

smcmurtry (approved):
My sending test is passing now so it looks like the failures/alarms were just due to other concurrent sends that were happening

PR #2990, opened 2026-08-05
Improve Celery thread safety

whabanks:
# Summary | Résumé

This PR aims to improves thread-safety by ensuring that we do not use mutable default parameters, in this case empty `dicts` or `lists`.

## How are mutable parameter defaults`{}` & `[]`  thread-unsafe?
- Python creates the method param object exactly once when the function is first defined, it does not create new objects per method call at runtime. 
- Because `dicts` and `lists` are **mutable objects**, their values can change (e.g. the function body mutates the parameter) without the memory address changing. 
- When `n` threads call the same function concurrently, each thread works with a pointer referencing the same parameter object, thus any mutations that occur within that function will mutate the object across all running threads, poisoning the data.

* https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/909

- [ ] CI is passing
- [ ] Threading is turned on in staging or dev and changes have been tested 

None.


app/celery/tasks.py
```diff
@@ -926,7 +926,7 @@ def seed_bounce_rate_in_redis(service_id: str, interval: int = 24):
 
 @notify_celery.task(name="generate-report")
 @statsd(namespace="tasks")
-def generate_report(report_id: str, notification_statuses=[]):
+def generate_report(report_id: str, notification_statuses=None):
     current_app.logger.info(f"Generating report for Report ID {report_id}")
```
jzbahrai, line 930: ooo damn i totally missed this when i reviwed this the first time

jzbahrai (approved):
great find and fixed!

PR #2994, opened 2026-08-06
Task/bounce rate support

jzbahrai:
# Summary | Résumé

This PR updates bounce-rate notification handling so Freshdesk support  receives a copy of the email when a service is actually suspended for bounce rate

## What changed
Added support-copy behavior for suspension emails only.
Explicitly does not send a support copy for warning emails.
Limited support-copy sending to production environment.
Added/used a dedicated configuration key for the Freshdesk support recipient: FRESHDESK_SUPPORT_EMAIL_ID.

### DIFF
--- a/app/celery/bounce_rate_tasks.py
+++ b/app/celery/bounce_rate_tasks.py
@@ -2,40 +2,60 @@ from flask import current_app
 
 from app import notify_celery
 from app.models import Service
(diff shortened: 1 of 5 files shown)


andrewleith (changes requested):
Looks good to me code-wise.  Could we get some tests for the new functions?

andrewleith (approved):
LGTM

PR #2997, opened 2026-08-10
Fix devcontainer

jzbahrai:
# Summary | Résumé

Fix devcontainer failures


.devcontainer/scripts/notify-dev-entrypoint.sh
```diff
@@ -85,14 +93,28 @@ echo "source ${POETRY_VENV_PATH}/bin/activate" >> ~/.zshrc
...
+    break
+  fi
+
+  if [ "$attempt" -lt "$db_upgrade_attempts" ]; then
+    echo "flask db upgrade failed (attempt ${attempt}/${db_upgrade_attempts}). Retrying in ${db_upgrade_delay_seconds}s..."
+    sleep "$db_upgrade_delay_seconds"
+  else
+    echo "WARNING: flask db upgrade failed after ${db_upgrade_attempts} attempts. Run 'poetry run flask db upgrade' once the database is ready."
+  fi
+done
```
andrewleith, line 117: What is all this stuff?  This is not related to ssh keys?

andrewleith (commented):
I expected you to add a mount like we did in admin - not totally clear how the key is accessible, but this does indeed work.  Just some questions about the extra changes that appear to be unrelated to this.

andrewleith (approved):

.devcontainer/scripts/notify-dev-entrypoint.sh
jzbahrai: not related, this is when your DB is not up at the same time as the dev container.

PR #2998, opened 2026-08-10
Only send clients email if report is not generated by api_key

jzbahrai:
# Summary | Résumé

Only send clients an email if the report is not generated by api_key (the email informs the client that the report is ready to download).

1. uploaded a csv of 10 files
2. POST /v2/reports
3. then GET /v2/reports
4. then GET /v2/reports/{report_ID}/content

^ there should be no error in celery when you run the above


andrewleith (approved):
LGTM - no error now when report generated via API.  Email still sent when report generated via UI.

PR #3002, opened 2026-08-12
Add migration to callback suspension

jzbahrai:
# Summary | Résumé

Add a template to suspend callbacks automatically

### DIFF
--- a/migrations/versions/0523_add_callback_susp_template.py
+++ b/migrations/versions/0523_add_callback_susp_template.py
@@ -3,7 +3,7 @@ Revision ID: 0523_add_callback_susp_template
 Revises: 0522_add_api_key_to_reports
 Create Date: 2026-08-12
 
-Add Notify template used when suspending a service's callbacks for not working.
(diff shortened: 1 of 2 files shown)


migrations/versions/0523_add_callback_susp_template.py
```diff
@@ -0,0 +1,128 @@
...
+            "[[fr]]",
+            "",
+            "(la version française suit)",
+            "",
+            "[[/fr]]",
+            "",
+            "[[en]]",
+            "Hello ((name)),",
+            "",
+            "We've suspended the callbacks for \"((service_name))\" because they're not working. This means that you will no longer receive any callbacks for your service.",
```
andrewleith, line 33: "because they're not working" is a bit vague.  What about something like "We've suspended the callbacks for ((service_name)) due to repeated errors with your callback URL"?

andrewleith (approved):
Just a suggestion about wording, looks fine though.

PR #3003, opened 2026-08-12
Improve redis script registry thread safety

whabanks:
# Summary | Résumé
This PR improves thread safety of the LUA script registry found in `RedisQueue` instances.

- Previously `RedisQueue` used a class-level mutable registry dict. This PR makes the `scripts` dict instance-scoped.
- `__register_scripts` now utilizes a lock to prevent cross-instance overwrites and mixed registry states during queue initialization
- Bumped utils version to include decorator thread-safety fixes

## Examples
Since we initialize multiple queues at module import time (sms_queue, bulk_queue, email_queue, etc.) and init them during application startup there's a couple different threading/concurrency risks that could occur here.

**1. Cross-instance overwrites (not currently a concern but could be in the future)**
Queue instance **A** could register lua scripts, then instance **B** overwrites those same script keys. This would result in instance **A** executing scripts created for instance **B**.

**2. Mixed registry state during initialization**
Script registration is two writes, (move_inflights & expire_inflights). Threads could read the registry in between writes and see a partially initialized state, leading to missing key errors.

## Future concerns
Once we start using highly threaded workers there the chance that, under high load/polling rate, that lock contention could should up. The impact should be minor as the lock is established only for the duration needed to acquire the registry dict and doesn't affect any of the Redis round-trips.
 

* https://github.com/cds-snc/notification-planning-core/issues/909

- [ ] CI is passing
- [ ] We test in staging once merged.

- [ ] Keep an eye on any lock contention around acquisition of the scripts registry dict under heavy system load.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -110,7 +110,7 @@ jobs:
         cp -f .env.example .env
 
     - name: Checks for new endpoints against AWS WAF rules
-      uses: cds-snc/notification-utils/.github/actions/waffles@53.2.32
+      uses: cds-snc/notification-utils/.github/actions/waffles@53.2.34
       with:
         app-loc: '/github/workspace'
         app-libs: '/github/workspace/env/site-packages'
--- a/app/queue.py
+++ b/app/queue.py
@@ -2,6 +2,7 @@ import random
 import string
 from abc import ABC, abstractmethod
 from enum import Enum
+from threading import Lock
 from typing import Any, Dict, Optional
 from uuid import UUID, uuid4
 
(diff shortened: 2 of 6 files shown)


jimleroyer (approved):
This is safer this way in case that the LUA scripts are parameterized in the future, but as these are not, my understanding is that overwrites by concurrent instances would end up the same in the end. All good though with extra safety.

whabanks:
> This is safer this way in case that the LUA scripts are parameterized in the future, but as these are not, my understanding is that overwrites by concurrent instances would end up the same in the end. All good though with extra safety.

Yea this is what I had in mind, more of a future proofing change as yes each instance would end up the same.

PR #3005, opened 2026-08-13
Pr1: Add exponential backoff for callbacks

jzbahrai:
## Summary
API PR1: Callback retries use exponential backoff with jitter

This PR updates service callback retry behaviour to use exponential backoff with jitter instead of fixed retry delays.

The goal is to reduce repeated pressure on unhealthy callback endpoints while preserving existing callback retry eligibility rules.

### What Changed
1. Introduced callback retry backoff policy configuration in the API callback task path.
2. Added jittered exponential countdown calculation for callback retries.
3. Applied the computed countdown when re-queueing retryable callback failures.
4. Added/updated unit tests to verify retry countdown behavior and boundaries.

**Behavior Before**
- Callback task retries used a fixed delay between attempts.
- Retryable error classes remained:
- request/network failures
- HTTP 5xx
- HTTP 429
- Non-retryable 4xx (except 429) did not retry.

**Behavior After**
- Same retryable/non-retryable classification as before.
- For retryable failures, delay now follows exponential backoff with jitter:
- $delay = \min(base \cdot 2^{attempt}, cap)$
base: base seconds for delat
attempt: attempt number
cap: maximum allowed delay in seconds
- $jittered = delay \cdot U(1-j, 1+j)$
U(a,b) means “pick a random number uniformly between a and b
j is jitter percent as a decimal (for example, j=0.2 for 20%).

> **When a callback fails, we do not keep retrying at the same speed. Instead, we wait a little longer each time before trying again (this is the delay), which gives the other service time to recover and prevents us from hammering it. Jitter adds a small random variation to that wait time, so retries from many failed messages do not all happen at the exact same second. Together, delay and jitter reduce traffic spikes, make retries smoother, and improve the chance that callbacks succeed once the target service is healthy again.**

## Why
- Fixed retry intervals can create retry storms against already struggling callback services.
- Exponential backoff with jitter improves system resilience and endpoint recovery chances.

## Testing**
1. Unit tests validate:
- countdown increases per attempt (exponential)
- countdown respects configured cap
- jitter stays within configured bounds
- retry invocation still happens only for retryable failures


andrewleith (commented):
This looks good in general.  I like the jitter idea so we dont end up nailing their service while its trying to recover.  One question about the overall retry period as this PR seems to reduce it substantially:

The old behaviour was:
- initial attempt: immediately
- retry 1: after 5 minutes
- retry 2: after 5 minutes
- retry 3: after 5 minutes
- retry 4: after 5 minutes
- retry 5: after 5 minutes

And the new behaviour is:
- retry 1: 5 seconds, plus or minus jitter
- retry 2: 10 seconds, plus or minus jitter
- retry 3: 20 seconds, plus or minus jitter
- retry 4: 40 seconds, plus or minus jitter
- retry 5: 80 seconds, plus or minus jitter

So before it would retry for a window of **25 minutes** (which means if the service came "back" in that time it would get the callback), but in the new version we only keep checking for **155 seconds.**  Is this new much smaller window intended?

jzbahrai:
@andrewleith good point, i changed the config to align to the old 25m spread

andrewleith (approved):
Looks good.  Only comment is that the delay is capped to `300s` but that may be intentional: the delay is calculated as a value between 240–360, but then is capped at a max of `300s`. The actual range is therefore 240–300 seconds, not +/- 20%.

PR #3007, opened 2026-08-14
Pr2: Add callback suspension task

jzbahrai:
## Summary
This PR adds automatic callback suspension when callback delivery retries are exhausted, and sends notification emails so both service owners and support have a record of the suspension event.

## Why
When a service callback endpoint keeps failing, retries alone do not protect overall callback system health. Suspending that callback after retry exhaustion limits repeated failures and creates a clear operational signal to owners and support.

## What Changed

1. Added retry-exhaustion suspension flow in notification-api/app/celery/service_callback_tasks.py.
1. On max retries exceeded, the callback config is suspended and an internal notify task is queued in notification-api/app/celery/service_callback_tasks.py.
1. Added a new task to send suspension emails to service owners and (production only) support in notification-api/app/celery/service_callback_suspension_tasks.py.
1. Registered the new task module in Celery imports in notification-api/app/config.py.
1. Added/updated tests for suspension and notification behavior in:
    notification-api/tests/app/celery/test_service_callback_tasks.py
    notification-api/tests/app/celery/test_service_callback_suspension_tasks.py

### Behavior Before
Callback retries stopped at max retries.
No automatic callback suspension on exhaustion.
No automatic service-owner/support email for callback suspension.

### Behavior After
Retryable callback failures still retry as before (including PR1 backoff+jitter).
When retries are exhausted, callback config is suspended automatically.
Suspension email is sent to service owners.
Support copy is sent in production for audit/operational visibility.
If callback is already suspended, duplicate suspend/email actions are avoided.

### DIFF
--- a/app/celery/service_callback_suspension_tasks.py
+++ b/app/celery/service_callback_suspension_tasks.py
@@ -1,0 +1,29 @@
+from flask import current_app
+
+from app import notify_celery
+from app.dao.services_dao import dao_fetch_service_by_id
+from app.service.sender import send_notification_to_email_address, send_notification_to_service_users
+
+
+@notify_celery.task(name="send-service-callback-suspension-email")
+def send_service_callback_suspension_email(service_id: str):
+    service = dao_fetch_service_by_id(service_id)
+    personalisation = {
+        "service_name": service.name,
+        "service_id": str(service.id),
+    }
+
+    send_notification_to_service_users(
+        service_id=service_id,
+        template_id=current_app.config["SERVICE_CALLBACK_SUSPENDED_TEMPLATE_ID"],
+        personalisation=personalisation.copy(),
+        include_user_fields=["name"],
+    )
+
+    # Keep support-copy behavior production-only to avoid noise in test/sandbox environments.
+    if current_app.config["NOTIFY_ENVIRONMENT"] == "production":
+        send_notification_to_email_address(
+            email_address=current_app.config["FRESHDESK_SUPPORT_EMAIL_ID"],
+            template_id=current_app.config["SERVICE_CALLBACK_SUSPENDED_TEMPLATE_ID"],
+            personalisation={**personalisation, "name": "Freshdesk support"},
+        )
--- a/app/celery/service_callback_suspension_tasks.py
+++ b/app/celery/service_callback_suspension_tasks.py
@@ -5,19 +5,25 @@ from app.dao.services_dao import dao_fetch_service_by_id
 from app.service.sender import send_notification_to_email_address, send_notification_to_service_users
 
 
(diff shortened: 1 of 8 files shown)


app/celery/service_callback_tasks.py
```diff
@@ -106,6 +114,43 @@ def _send_data_to_service_callback_api(self, service_id, data, service_callback_
             try:
                 self.retry(queue=QueueNames.CALLBACKS_RETRY, countdown=countdown)
             except self.MaxRetriesExceededError:
+                _suspend_service_callback_and_send_email(service_id=service_id, callback_type=callback_type)
```
andrewleith: Could we carry the callback API ID through the task payload, or verify that the currently configured callback still matches the failed URL before suspending it? Otherwise an old retry task may suspend a newer callback configuration for the same service and callback type.

app/celery/service_callback_suspension_tasks.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def _build_callback_suspension_personalisation(service_id: str, service_name: str) -> dict:
+    admin_base_url = current_app.config["ADMIN_BASE_URL"].rstrip("/")
+    return {
+        "service_name": service_name,
+        "service_id": service_id,
+        "service_callback_url_en": f"{admin_base_url}/services/{service_id}/api/callbacks/delivery-status-callback",
+        "service_callback_url_fr": f"{admin_base_url}/services/{service_id}/api/callbacks/delivery-status-callback?lang=fr",
+    }
```
andrewleith, line 15: We use delivery callback urls even if its the complaint type that failed. Should `callback_type` be passed so we can generate the correct url?

andrewleith (commented):
A few comments below.  One question I have is that this seems to set the plumbing to send out emails when a complaint callback fails, but I dont think this is something users can even do?  All complaints seem to come back to notify itself (not positive on this).

Another potential issue is that we may send multiple emails when we are suspending due to multiple celery workers.  Before suspending and sending the notification, can we make the “already suspended?” check and the update atomic? Also, should retries already queued when the callback becomes suspended be skipped, or is it acceptable for them to continue?

jzbahrai:
ok so the new changes:
1. I made updating the service callback table atomic. So if one celery worker is updating it, the other worker will need that update to finish before it can read the state of the callback. This check is used to decide whether to send a suspension email or not.

2. this change im still on the fence about. Before we send a callback, we go and fetch the current state of the callback from the DB. If the callback the process has and the callback from the DB are different, we will now send the callback with the new url.

andrewleith (approved):
LGTM. You are on the fence about updating the URL mid-retry cycle when it changes?  I understand that, maybe we should just let the broken URL die out and only newer notifications get the latest one?

PR #3008, opened 2026-08-14
Emit a log line to search on for international messages

jzbahrai:
# Summary | Résumé

Emit a log line to capture if we have sent an international text message
https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/3367


whabanks (approved):
Aside from the one AI comment, this looks good. Will approve but the comment should be addressed.

PR #3009, opened 2026-08-17
Fix template history drift for category and process-type updates

jzbahrai:
# Summary:
This PR fixes Template history drift by making Template-specific history writes deterministic and keeping template category state consistent before versioned history capture.

It removes reliance on shared history behavior changes and keeps the fix scoped to Template paths only.

## What Changed
- Consolidated manual Template history row creation in templates_dao.py so all manual update paths write a consistent set of fields, including template_category_id and process_type override column.
- Fixed process type update path in templates_dao.py to:
    update mapped attributes correctly
    increment version
    set updated_at
    persist matching history rows
- Added Template-only category state synchronization in templates_dao.py before create and update versioned writes, so FK and relationship state cannot diverge in ways that cause stale history values.
- Expanded regression coverage in test_templates_dao.py, including:
    category history correctness on normal updates
    manual update-path history correctness (reply-to, process type set/unset)
    stale loaded relationship scenario where template_category_id changes
    create path where category is set via relationship
    
### Why This Fix
The mismatch came from in-memory FK and relationship state not always being aligned at history-copy time. This PR fixes that in Template DAO flows directly, while keeping shared history behavior unchanged for other models.

## Tests:

Case 1
1. Create a template, with category "authentication"
2. Check the template category is the same in "templates" and "template_history" for the same version

Case 2:
1. Edit the above template
2. Add an "admin" override for the process_type (at the bottom of the edit page). Click "NORMAL" or "BULK"
3. "Templates" table should have the process_type column filled and "template_history" as well. Make sure it is the same version (for the edit)
4. Open a python shell CHANGE the UUID.
```
python - <<'PY'
from uuid import UUID
from application import app
from app.models import Template, TemplateHistory

tid = UUID("202d52a2-1c74-4ca5-93eb-8d00436cc3ed")

print("---------------------------------------------")
with app.app_context():
    t = Template.query.filter_by(id=tid).one_or_none()
    if t is None:
        print("Template not found")
    else:
        print("template_type:", t.template_type)
        print("override process_type_column:", t.process_type_column)
        print("resolved process_type:", t.process_type)
        print("template_category_id:", t.template_category_id)
        if t.template_category:
            print("category sms_process_type:", t.template_category.sms_process_type)
            print("category email_process_type:", t.template_category.email_process_type)

    print("\nHistory:")
    rows = (
        TemplateHistory.query
        .filter_by(id=tid)
        .order_by(TemplateHistory.version.asc())
        .all()
    )
    for h in rows:
        print(
            "v=", h.version,
            "override=", h.process_type_column,
            "resolved=", h.process_type,
            "category_id=", h.template_category_id,
            "updated_at=", h.updated_at,
        )
PY
```

Case 3:
1. Edit the above template
2. Under admin click "Use template category"
3. Both templates and template history, the process_type column should be None
4. When you get the template it will use the process type from the template cat
5. you can use the above script to figure out the process types

### DIFF
--- a/app/dao/templates_dao.py
+++ b/app/dao/templates_dao.py
@@ -40,6 +40,9 @@ def _create_template_history_row(template):
             "process_type": template.process_type_column,
             "template_category_id": template.template_category_id,
             "service_letter_contact_id": template.service_letter_contact_id,
+            "hidden": template.hidden,
+            "text_direction_rtl": template.text_direction_rtl,
+            "use_custom_unsubscribe_url": template.use_custom_unsubscribe_url,
         }
     )
 
--- a/app/dao/templates_dao.py
+++ b/app/dao/templates_dao.py
@@ -6,7 +6,7 @@ from typing import Union
 from flask import current_app
 from notifications_utils.clients.redis import template_version_cache_key
 from sqlalchemy import asc, desc
-from sqlalchemy.orm import joinedload
+from sqlalchemy.orm import attributes, joinedload
 
 from app import db, redis_store
 from app.dao.dao_utils import VersionOptions, transactional, version_class
@@ -15,6 +15,7 @@ from app.models import (
     LETTER_TYPE,
     SECOND_CLASS,
     Template,
+    TemplateCategory,
     TemplateHistory,
     TemplateRedacted,
 )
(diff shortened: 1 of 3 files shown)


app/dao/templates_dao.py
```diff
@@ -15,11 +15,60 @@
...
+            "archived": template.archived,
+            # Keep history aligned with the DB column: only explicit overrides are stored here.
+            "process_type": template.process_type_column,
+            "template_category_id": template.template_category_id,
+            "service_letter_contact_id": template.service_letter_contact_id,
+        }
+    )
+
+
+def _sync_template_category_state(template):
```
jzbahrai, line 50: we needed this cause at times, the template table would be updated to a new temp category, but the new category would not make its way to template history.

app/dao/templates_dao.py
```diff
@@ -15,11 +15,60 @@
...
+            "postage": template.postage,
+            "created_by_id": template.created_by_id,
+            "version": template.version,
+            "archived": template.archived,
+            # Keep history aligned with the DB column: only explicit overrides are stored here.
+            "process_type": template.process_type_column,
+            "template_category_id": template.template_category_id,
+            "service_letter_contact_id": template.service_letter_contact_id,
+        }
+    )
```
andrewleith, line 47: Should `_create_template_history_row()` copy all persisted Template fields that `create_history()` copies, particularly `hidden`, `text_direction_rtl`, and `use_custom_unsubscribe_url`? Could these drift as well?

andrewleith (approved):
LGTM - works when I test it.  Just a question about a few other fields.

app/dao/templates_dao.py
jzbahrai: fixed

PR #3011, opened 2026-08-18
Rudimentary impl of dogpile caching

whabanks:
# Summary | Résumé
This PR implements the basics required to enable [dogpile caching](https://dogpilecache.sqlalchemy.org/en/latest/) in our system. In essence dogpile caches DB responses resulting from DAO method calls; similar to how we cache API responses in Admin, just at the DAO level.

### Current progress
- Add dogpile-cache to pyproject.toml
- Set up a basic, redis-backed, dogpile cache region with `distributed_lock` turned on and `thread_local_lock` turned off to play nicely in a threaded environment
- Implement a cache-key generation strategy
- Implement dogpile cached version of `dao_fetch_service_by_id`
- ~Invalidate cache whenever a service's settings are updated~ Will settle on an invalidation implementation after evaluating the [viability of ORM-event hooks](https://github.com/cds-snc/notification-api/pull/3029/) vs manually invoking custom invalidation methods. I may be a hybrid of both approaches.

### TODO
- [x] [Implement a custom `EncodeProxy`](https://dogpilecache.sqlalchemy.org/en/latest/recipes.html#encoding-decoding-data-into-another-format) to support JSON encoding of cache values (it defaults to pickling)
- [x] Explore different eviction strategies
- [x] Document findings, pitfalls, caveats, things to keep in mind, etc.

1. Set `FF_USE_DOGPILE_CACHING=true` in your .env
2. Run admin and api locally
3. **From the live services list** click on any service
4. Browse your local redis and note the new cached key:
image

**Note:** If the cache key didn't appear, you may need to check redis for an `service-<id>` key and delete it then access that service again to trigger the DB call + dogpile caching.

None.


app/caching.py
```diff
@@ -0,0 +1,62 @@
...
+        return f"{namespace}:{fname}-{primary_id}"
+
+    return generate_key
+
+
+dogpile_region = make_region(function_key_generator=cache_key_generator)
+
+
+def init_dogpile_cache(app):
+    redis_url = app.config.get("REDIS_URL", "redis://localhost:6379/0")
```
jimleroyer, line 91: So expiring cache such as triggered by service update will affect all instances relying on it via a Redis cleared cache, right?

app/service/rest.py
```diff
@@ -309,6 +311,7 @@ def update_service(service_id):
         service.letter_branding = None if not letter_branding_id else LetterBranding.query.get(letter_branding_id)
 
     dao_update_service(service)
+    invalidate_service_cache(str(service_id))
```
jimleroyer: Is there a way to use annotation to indicate that a call to this function should invalidate the service cache? (and you might have preferred here to have a function call instead to have more control over the cache expiration)

whabanks: We can certainly take an annotation approach here, similar to what we do in [Admin](https://github.com/cds-snc/notification-admin/blob/main/app/notify_client/service_api_client.py#L18-L19) and add those annotations at the DAO level instead.

app/caching.py
whabanks: Yes, that's correct. Each running instance relies on the same Redis backend. The `distributed_lock=True` setting ensures that the dogpile mutex is coordinated via Redis. So even across all running pods / instances of dogpile cache regions, only one instance is elected to perform the updated value while the other instances are still served the "stale" values until that lock is released and the cache update is complete.

app/service/rest.py
whabanks: I ended up changing the implementation a bit, rather than having a per-model / per-dao method style cache invalidation I made it more generic. It can now target a group (e.g service, user), narrow down by entity id and perform sweeping invalidations. 

e.g. A service is updated, then all cached dao calls related to that service's id can be cleared in a batch with patterns instead of manually tracking down all possible dao calls that should be cleared. It's simple right now, with just groups + entity ids but we have a lot of options to place groups into namespaces of inter-related data to make cache invalidation even easier.

app/caching.py
```diff
@@ -0,0 +1,104 @@
...
+def _get_redis_client_from_region():
+    backend = dogpile_region.backend
+    client = getattr(backend, "writer_client", None) or getattr(backend, "client", None)
+
+    if client is None:
+        raise RuntimeError("Dogpile region is not using a Redis backend client or the region has not yet been initialized")
+    return client
+
+
+def invalidate_group_keys(group_name, group_id, batch_size=500, namespace=None):
```
jimleroyer, line 68: We are overriding the dogpile cache expiration here? In favor of being able to have groups? Would that make sense to have the ability to delete the entire group, and not only the group_id then?

whabanks: Technically yes, we are overriding the default cache expiration when manually invalidating. I took this approach because dogpile doesn't have native "delete by group" functionality built in, it can only invalidate 1 key at a time, which would lead to repetitious calls to their API when we can just do a redis `SCAN` + `DEL` and do it faster. 

For the `group` vs `group_id` it's a bit ambiguously named, `group_id` should really be `entity_id`. 

`group_name` = service, template, user, etc.
`group_id` = service_id, template_id, user_id, etc.

There's nothing stopping us from just broadly invalidating all `service:*` keys `if group_id is None` for example. It just depends on the code path / action that triggers invalidation and whether we need a cross-cutting / cascading delete.

Once we map and better understand current caching behaviour that will better inform our invalidation strategy impl. (I outlined a few examples of this in the card)

Let me know if that makes sense at all 😸

jimleroyer (approved):
LGTM - this is good to test and go forward. We likely need to experiment on the best setup and get app's team feedback on this. cc @andrewleith

PR #3019, opened 2026-08-20
Fix py.test to pytest

jzbahrai:
# Summary | Résumé

Most likely because of cache + venv recreation timing, not because your test code changed.

What changed underneath:

The GitHub runner Python patch version changed (your log shows 3.12.14).
Your venv cache key in test.yaml:47 and test.yaml:172 is keyed as py312 + lock hash, so a cached venv from an earlier patch can still be restored.
Poetry then detected that restored venv as broken and recreated it during the test step (exactly what your log says).
In this workflow, install is skipped on cache hit at test.yaml:77, test.yaml:185, and test.yaml:285, so the recreated venv can be empty.
Then the command fails with command not found.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -188,6 +188,13 @@ jobs:
     - name: Generate version file
       run: make generate-version-file
 
+    - name: Ensure pytest is available
+      run: |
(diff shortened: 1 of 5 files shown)


andrewleith (approved):
LGTM

Issue #2, opened 2019-06-27
When using our copy of the notify repo we should connect it to our github actions/apps so we can automate as much testing as we can


willeybryan:
**What?**
- Figure out what Github actions we can apply to the repos
- What actions we should
- Then apply them

**Why?**
- Test our own tech and ensure we don't make little mistakes.

maxneuvians:
All repos now have CI/CD

https://github.com/cds-snc/notification-utils/pull/2
https://github.com/cds-snc/notification-api/pull/75
https://github.com/cds-snc/notification-admin/pull/8

Issue #1, opened 2019-06-27
Get rid of Gov.uk Branding in UI


willeybryan:
**What?**
- Colours are pulled in via an NPM from a deprecated UK repo
- We need to update the colours to CDS agreeable ones (Use the colour pallette from digital.canada.ca, Black, white, Dark grey, yellow highlights) As a placer holder until we get design help
- Replace Gov.uk UI logos with GoC logo

**Why?**
- Doesn't look very good and creates some design faults
- Should be considered with the Font change in https://github.com/cds-snc/CANotifier/issues/24

willeybryan:
@willeybryan  First step, confirm Notifier name

timarney:
Going to close this - we can adjust further under fresh tickets for any specific updates.

PR #3030, opened 2026-08-27
Fix the API bulk notifications priority being ignored

jimleroyer:
# Summary | Résumé

Fixes a bug in the batch-save path that caused priority notifications to be routed to the wrong Celery queue. In `try_to_send_notifications_to_queue`, the `template` loop variable leaked out of the batch loop, so every notification in a mixed-priority batch was routed as if it had the last notification's priority. A priority email batched with a normal email would land on `send-email-medium` and warn `queue_name send-email-high but was sent to queue send-email-medium`.

Each notification is now routed by its own `queue_name` (correctly set per-notification in `persist_notifications`), so mixed-priority batches route correctly. The queue-override map has been left keyed by strings, matching the in-session type of `notification.id` inside these functions; a defensive `str()` cast at the lookup site makes the code robust if a future refactor causes `.id` to be exposed as a `uuid.UUID`.

* [Investigate 5 minutes delay for failed high priority notifications](https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/919)

# What changed

### `app/celery/tasks.py`

- **Per-notification routing in `try_to_send_notifications_to_queue`**: fall back to `notification.queue_name` before the stale batch template, so mixed-priority batches route correctly. Added a `TODO` noting that the override map can be removed once `persist_notifications` handles the CSV bulk-redirect rule.
- **Defensive `str()` cast on the map lookup**: `notification_id_queue.get(str(notification_obj.id))`. Inside `save_smss`/`save_emails`, `notification.id` is a string in-session (SQLAlchemy's `bulk_save_objects` doesn't refresh objects), so the lookup works as before — but the cast keeps it robust if a future change ever refreshes the object and turns `.id` into a `uuid.UUID`. Added a short comment explaining the ambiguity, plus a `Dict[str, Optional[str]]` type hint on the map for clarity.

### `tests/app/celery/test_tasks.py`

New `TestTryToSendNotificationsToQueue` class with 5 tests:

- `test_uses_each_notifications_own_queue_name_in_mixed_priority_batch` — mixed batch routes each notification to its own queue.
- `test_notification_id_queue_override_wins_over_queue_name` — CSV bulk-redirect override takes precedence when set.
- `test_falls_back_to_template_queue_when_queue_name_and_map_are_both_empty` — legacy fallback path.
- `test_priority_notification_not_routed_to_last_batch_notifications_queue` — direct regression for the production bug (mirrors the 5-notification batch from the log).
- `test_lookup_is_robust_to_notification_id_being_a_uuid_or_string` — documents the defensive `str()` cast so the lookup keeps working whether `notification.id` is a string (current behavior) or a `uuid.UUID` (future).

Also updated one existing `TestSaveEmails` test expectation to reflect the new (correct) behavior where a mocked `choose_queue` return value now actually flows through to the delivery call.

# Impact / risk

- **What was broken**: in mixed-priority batches, priority notifications occasionally landed on the medium queue, defeating the intent of the priority lane.
- **User-visible effect**: high-priority messages could sit behind normal traffic on the medium queue, showing up as extra latency (and, when delivery failed, as a 5-minute retry delay instead of 25 seconds — see companion PR on the retry path).
- **After this PR**: each notification is routed by its own priority; SMS and email paths behave consistently; CSV bulk-redirect and other pre-existing map overrides continue to work as before.

1. Run the new regression tests:
   ```
   pytest tests/app/celery/test_tasks.py::TestTryToSendNotificationsToQueue -v
   ```
   All 5 tests should pass.

2. Run the broader batch-save tests to confirm no regressions:
   ```
   pytest tests/app/celery/test_tasks.py::TestBatchSaving tests/app/celery/test_tasks.py::TestSaveEmails tests/app/celery/test_tasks.py::TestSaveSmss -v
   ```

3. After deploying to staging, verify with CloudWatch Logs Insights that the `has queue_name send-email-high but was sent to queue send-email-medium` warning no longer appears for mixed-priority batches:
   ```
   fields @timestamp, @message
   | filter @message like /has queue_name send-email-high but was sent to queue/
   | sort @timestamp desc
   | limit 20
   ```

None.


app/celery/tasks.py
```diff
@@ -517,7 +517,16 @@ def try_to_send_notifications_to_queue(notification_id_queue, service, saved_not
...
+            # TODO: remove notification_id_queue once persist_notifications knows about the CSV bulk-redirect
+            # rule (choose_sending_queue). Until then, the map carries the only signal for that override.
+            queue = (
+                # CSV bulk-redirect override (only useful for CSV jobs)
+                notification_id_queue.get(notification_obj.id)
+                # per-notification correct value from persist_notifications
+                or notification_obj.queue_name
+                # legacy fallback (safe to keep, but essentially unreachable)
+                or get_delivery_queue_for_template(template)
+            )
```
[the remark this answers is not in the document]
jimleroyer: A valid bug that wasn't found previously. An explicit cast to `str` was added to avoid this issue.

tests/app/celery/test_tasks.py
```diff
@@ -513,6 +514,110 @@ def test_process_smss_job_metric(self, sample_template_with_placeholders, mocker
...
+        priority_id, normal_id, bulk_id = uuid.uuid4(), uuid.uuid4(), uuid.uuid4()
+        saved_notifications = [
+            self._make_notification(priority_id, QueueNames.SEND_EMAIL_HIGH),
+            self._make_notification(normal_id, QueueNames.SEND_EMAIL_MEDIUM),
+            self._make_notification(bulk_id, QueueNames.SEND_EMAIL_LOW),
+        ]
+        # API-buffered flow: notification_id_queue is populated with None values because the
+        # signed payload from post_notifications does not include a "queue" key.
+        notification_id_queue = {priority_id: None, normal_id: None, bulk_id: None}
+
```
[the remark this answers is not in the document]
jimleroyer: A valid bug that wasn't found previously. An explicit cast to `str` was added to avoid this issue, as well as a new tests explicitly targeting the mismatch and that would fail without the explicit `str` cast.

tests/app/celery/test_tasks.py
```diff
@@ -513,6 +514,130 @@ def test_process_smss_job_metric(self, sample_template_with_placeholders, mocker
...
+        # Stale template from the last (normal) iteration of the save_emails loop.
+        stale_template = MagicMock(template_type=EMAIL_TYPE, process_type="normal")
+
+        try_to_send_notifications_to_queue(notification_id_queue, service, saved_notifications, stale_template)
+
+        # The priority notification (index 3) must go to send-email-high, not send-email-medium.
+        priority_call = send_mock.call_args_list[3]
+        assert priority_call == call(saved_notifications[3], False, QueueNames.SEND_EMAIL_HIGH)
+
+    def test_lookup_is_robust_to_notification_id_being_a_uuid_or_string(self, notify_api, mocker):
```
jimleroyer, line 594: This is the test that targets the new bug that was found by the AI reviewer in this PR. Different from the original bug that this PR also targets (i.e. missing/lost priority type for bulk API notifications).

app/celery/tasks.py
```diff
@@ -517,7 +517,19 @@ def try_to_send_notifications_to_queue(notification_id_queue, service, saved_not
...
+            # TODO: remove notification_id_queue once persist_notifications knows about the CSV bulk-redirect
+            # rule (choose_sending_queue). Until then, the map carries the only signal for that override.
+            # Map keys are strings; notification_obj.id is a string in-session (persist_notifications does
+            # not refresh the SQLAlchemy object) but could be a uuid.UUID after a DB refresh — the str()
+            # cast makes the lookup robust either way.
+            queue = (
+                # CSV bulk-redirect override (only useful for CSV jobs)
+                notification_id_queue.get(str(notification_obj.id))
+                # per-notification correct value from persist_notifications
+                or notification_obj.queue_name
```
jimleroyer, line 510: Bug fix #1

app/celery/tasks.py
```diff
@@ -517,7 +517,19 @@ def try_to_send_notifications_to_queue(notification_id_queue, service, saved_not
...
         try:
-            queue = notification_id_queue.get(notification_obj.id) or get_delivery_queue_for_template(template)
+            # TODO: remove notification_id_queue once persist_notifications knows about the CSV bulk-redirect
+            # rule (choose_sending_queue). Until then, the map carries the only signal for that override.
+            # Map keys are strings; notification_obj.id is a string in-session (persist_notifications does
+            # not refresh the SQLAlchemy object) but could be a uuid.UUID after a DB refresh — the str()
+            # cast makes the lookup robust either way.
+            queue = (
+                # CSV bulk-redirect override (only useful for CSV jobs)
+                notification_id_queue.get(str(notification_obj.id))
```
jimleroyer, line 508: Bug fix #2

andrewleith (commented):
This seems reasonable to me for email.  Any reason this fix should not also be applied to the SMS code? Doesn't that have the same issue?

jimleroyer:
@andrewleith Hmm I was under the impression that the `try_to_send_notifications_to_queue` function would cover both paths, but it seems it's only used for emails. I will add tests around SMS path first, to add necessary coverage, and see if we can reuse `try_to_send_notifications_to_queue` in there too.

app/celery/tasks.py
```diff
@@ -382,24 +382,8 @@ def save_smss(self, service_id: Optional[str], signed_notifications: List[Signed
...
-                "SMS {} created at {} for job {}".format(
-                    notification_obj.id,
-                    notification_obj.created_at,
-                    notification_obj.job,
-                )
-            )
-        except (LiveServiceTooManySMSRequestsError, TrialServiceTooManySMSRequestsError) as e:
-            current_app.logger.info(f"{e.message}: SMS {notification_obj.id} not created")
+    if saved_notifications:
+        try_to_send_notifications_to_queue(notification_id_queue, service, saved_notifications, template)
```
jimleroyer: The fixes are in the `try_to_send_notifications_to_queue` function, which I thought initially was called by the SMS path but wasn't. Now that path really does call that function, which was likely a previous refactor from years ago but was only applied to email and not SMS. We have a function that will handle it for both now.

app/celery/tasks.py
```diff
@@ -501,38 +485,51 @@ def save_emails(self, _service_id: Optional[str], signed_notifications: List[Sig
...
-    """
-    current_app.logger.debug(f"Sending following email notifications to AWS: {notification_id_queue.keys()}")
+def try_to_send_notifications_to_queue(notification_id_queue, service, saved_notifications):
+    """Route each saved notification to its delivery queue. Works for both SMS and email batches."""
+    current_app.logger.debug(f"Sending following notifications to provider queue: {notification_id_queue.keys()}")
     # todo: fix this potential bug
     # service is whatever it was set to last in the for loop above.
     # at this point in the code we have a list of notifications (saved_notifications)
     # which could be from multiple services
     research_mode = service.research_mode  # type: ignore
```
whabanks, line 498: Could a situation similar to the template loop variable leaking out occur for research_mode as well?

e.g: The final service is in research mode and forces notifications into the research queue.

Should `research_mode` be derived per-notification?

whabanks (commented):
LGTM, just one question about `research_mode` possibly behaving similarly to the template bugs fixed here.

app/celery/tasks.py
jimleroyer: Yes absolutely, it's the 4th bug I plan to fix (and indicated by the comment). I will open a new PR for that one, considering the scope and size of the current one.

P0NDER0SA (approved):
OK!

PR #3033, opened 2026-08-31
Task/dogpile orm event invalidation

whabanks:
# Summary | Résumé

This PR experimentally implements an ORM-event driven invalidation strategy for `Service` entities. Rather than calling an explicit `invalidate_service_cache_keys()` function on every path that mutates service and service related data; we instead listen to `after_flush`, `after_commit`, and `after_rollback` sqlalchemy session events, weaving cache invalidation calls into that process. (similar in nature to [aspect oriented programming](https://en.wikipedia.org/wiki/Aspect-oriented_programming))

This nets us:
- Centralized invalidation logic
- Reduced code complexity, and improve maintainability particularly on larger models where multiple mutating code paths are involved
- Tighter control over when we invalidate i.e. if a transaction is rolled back there's no need to invalidate existing cache
- Since we're handling this logic much closer to the actual time the DB is updated, we reduce the chances of stale cache reads after a successful write.

## How it works
SQLAlchemy’s `after_flush` event examines new, changed, and deleted ORM instances. The invalidation registry maps each model to one or more cache groups, such as `Template.id` to the template namespace or ServiceUser.user_id to the user and service namespaces.

Invalidation targets accumulate in `Session.info` and when a successful `after_commit` event is fired, we process and invalidate these keys. If a `rollback` occurs we discard the invalidations.

<details>
  <summary>High level diagram</summary>

  ```mermaid
  flowchart TD
      A[DAO mutates an ORM entity] --> B[Session flushes changes]
      B --> C[after_flush event]
      C --> D[Inspect new, dirty, and deleted entities]
      D --> E[Match entity model against invalidation registry]
      E --> F[Resolve cache namespace and entity ID]
      F --> G[Queue namespace and entity ID in Session.info]
      G --> H{Transaction result}
  
      H -->|Commit| I[after_commit event]
      I --> J[Remove queued targets from Session.info]
      J --> K[Scan Redis for namespace:entity_id:*]
      K --> L[Delete matching Dogpile cache keys]
  
      H -->|Rollback| M[after_rollback event]
      M --> N[Discard queued targets]
      N --> O[Existing cache remains unchanged]
  ```
</details>

### Invalidation registry
Maps models to cache namespaces / keys. It's purpose is to provide cross-entity dependency mapping when the model being mutated is part of a cache group, or multiple cache groups.
 
```python
Service: [
        CacheInvalidationRule(namespace="service", entity_id_attribute="id"),
    ],
    ServicePermission: [
        CacheInvalidationRule(namespace="service", entity_id_attribute="service_id"),
    ],
    ServiceUser: [
        CacheInvalidationRule(namespace="service", entity_id_attribute="service_id"),
        CacheInvalidationRule(namespace="user", entity_id_attribute="user_id"),
    ],
```

**NOTE:** If you previously accessed whatever service you use during testing you may already have a `service-{id}` key in Redis, delete this key first.

1. Access a service, note the cache key is created
6. Navigate to service settings and turn off email sending (or change any setting) -> note the cache key was updated
7. Add or remove a team member from a service -> note the cache key reflects this change

None.


whabanks:
Original convo from [here](https://github.com/cds-snc/notification-api/pull/3029#issuecomment-5444377286)
From @jimleroyer 

I am concerned that this approach (and the other PR too but maybe slightly less) might be difficult to scale, i.e. every time a developer wants certain cache to be added, they would have to make sure proper specific events fire, and that these are for service IDs, is that right? And when we want to add a cache specific to users, and templates, and keys, and etc.. we'd have also to add specific glue code for these? The cache isn't generic enough to handle these automatically?

Response
---

> every time a developer wants certain cache to be added, they would have to make sure proper specific events fire, and that these are for service IDs, is that right? 

With the current implementation, yes this would have to be done to accommodate more than just the service entity.

> And when we want to add a cache specific to users, and templates, and keys, and etc.. we'd have also to add specific glue code for these? The cache isn't generic enough to handle these automatically?

The baseline dogpile library isn't smart enough to graph and evaluate the potential entity relationships and invalidate corresponding keys on it's own. One possibility to facilitate a generic implementation would be to leverage a registry that centrally stores (and therefore documents) entity <-> cache key dependency metadata. 

Then developers only need to update the registry when new relationships are added in the DB, or additional DAO methods are marked via annotation to leverage dogpile.

Something like this:

image


The two new things here would be the `registry` and `dependency resolver`. 
- The session events are already in place 
- The change collector would be adjusted to generically map collected ids to an entity type 
- The cache invalidator would be changed to map entities to cache keys base on the registry and execute the invalidation

I've taken the liberty to implement the beginnings of such an approach and it is now included in this PR.

app/dao/templates_dao.py
```diff
@@ -154,6 +156,13 @@ def dao_update_template_category(template_id, category_id):
     return template
 
 
+@cache_on_arguments(namespace="template", group_by="template_id")
+def dao_get_template_by_id_cached(template_id, service_id) -> dict:
+    """Return a JSON-safe cached representation of a service's template."""
+    template = dao_get_template_by_id_and_service_id(template_id=template_id, service_id=service_id)
+    return template_schema.dump(template)
```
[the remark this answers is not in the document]
whabanks: This is expected for the progress on this PR and is addressed in this follow up PR:
https://github.com/cds-snc/notification-api/pull/3045

app/__init__.py
```diff
@@ -156,6 +156,9 @@ def create_app(application, config=None):
     redis_store.init_app(application)
     bounce_rate_client.init_app(application)
     init_dogpile_cache(application)
+    from app.cache.cache_events import register_cache_orm_events
+
+    register_cache_orm_events()
```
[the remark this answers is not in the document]
whabanks: Fixed

app/cypress/rest.py
```diff
@@ -141,6 +142,7 @@ def _destroy_test_user(email_name):
         # and is a bug
         cypress_service = Service.query.filter_by(id=current_app.config["CYPRESS_SERVICE_ID"]).first()
         cypress_service.created_by_id = current_app.config["CYPRESS_TEST_USER_ID"]
+        invalidate_service_cache_keys(cypress_service.id)
         # update the smoktest templates' created_by to be the main cypress user
```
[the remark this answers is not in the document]
whabanks: Will address in this PR https://github.com/cds-snc/notification-api/pull/3045 as a partial solution exists there (queuing cache invalidations).

app/caching.py
```diff
@@ -142,6 +137,31 @@ def invalidate_group_keys(group_name, group_id, batch_size=500, namespace=None):
...
+    This invalidates the exact service fetch keys and then best-effort clears
+    all grouped service keys for the id prefix.
+    """
+    normalized_service_id = str(service_id)
+
+    # Prefix invalidation is best-effort and requires Redis backend access.
+    try:
+        invalidate_group_keys("service", normalized_service_id)
+    except Exception:
+        pass  #  Failures are swallowed because request/transaction paths should not fail due to cache backend issues.
```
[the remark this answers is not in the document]
whabanks: Won't fix for now, will be fixed in https://github.com/cds-snc/notification-api/pull/3045

app/cache/cache_events.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def _invalidate_cache_after_commit(session):
+    """Delete grouped cache keys affected by the committed transaction."""
+    invalidations = session.info.pop(_CACHE_INVALIDATIONS_KEY, set())
+    for namespace, entity_id in invalidations:
+        try:
+            invalidate_group_keys(namespace, entity_id)
+        except Exception:
+            pass
```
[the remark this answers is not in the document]
whabanks: Fixed

whabanks:
Regarding the AI suggested changes, most of those are implemented already in the [other PR in this stack](https://github.com/cds-snc/notification-api/pull/3045)

app/cache/cache_events.py
```diff
@@ -0,0 +1,77 @@
...
+    """Resolve changed ORM instances to cache groups and collect them.
+
+    SQLAlchemy calls this after pending ORM changes are flushed to SQL. We do
+    not invalidate here because the transaction can still roll back.
+
+    Targets are stored as ``(namespace, entity_id)`` tuples in ``session.info``.
+    The set deduplicates cases where several changed rows affect the same cache
+    group in one transaction.
+    """
+    invalidations = session.info.setdefault(_CACHE_INVALIDATIONS_KEY, set())
```
jzbahrai, line 32: so if we are running this after every insert/ update to the DB is that expensive? Do we need to run cache invalidation on every model?

jzbahrai: like it loops through the entire registry?

app/cache/cache_invalidation_registry.py
```diff
@@ -0,0 +1,47 @@
...
+
+from app.models import (
+    Permission,
+    Service,
+    ServicePermission,
+    ServiceUser,
+    Template,
+    TemplateRedacted,
+    User,
+)
```
jzbahrai, line 13: so we dont have history tables here. Could there be a mismatch between template, template history and template cache and the redis cache? :/

app/cache/cache_invalidation_registry.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+@dataclass(frozen=True)
+class CacheInvalidationRule:
+    """Describe how an ORM model change maps to a cache group."""
```
jzbahrai, line 18: the orm model change is ID? but the ID doesn't change, its the associated data with that Id?

app/dao/service_permissions_dao.py
```diff
@@ -14,9 +14,19 @@ def dao_add_service_permission(service_id, permission):
...
 
 def dao_remove_service_permission(service_id, permission):
-    deleted = ServicePermission.query.filter(
+    """Remove matching service permissions via ORM deletes.
+
+    We intentionally delete loaded ORM objects instead of issuing a bulk
+    ``query.delete()`` so SQLAlchemy session events can observe these changes.
+    This allows service cache invalidation hooks to run consistently.
+    """
+    service_permissions = ServicePermission.query.filter(
```
jzbahrai, line 23: does this need to be behind the feature flag? cause the ff dog pile feature flag needs to be on for this behaviour?

app/dao/service_permissions_dao.py
```diff
@@ -14,9 +14,19 @@ def dao_add_service_permission(service_id, permission):
...
         ServicePermission.service_id == service_id,
         ServicePermission.permission == permission,
-    ).delete()
+    ).all()
+
+    for service_permission in service_permissions:
+        db.session.delete(service_permission)
+
```
jzbahrai, line 30: this is a small model but essentially we are doing a get first and then delete? Why this pattern?

app/dao/templates_dao.py
```diff
@@ -154,6 +156,13 @@ def dao_update_template_category(template_id, category_id):
     return template
 
 
+@cache_on_arguments(namespace="template", group_by="template_id")
+def dao_get_template_by_id_cached(template_id, service_id) -> dict:
+    """Return a JSON-safe cached representation of a service's template."""
```
jzbahrai, line 161: im hesitant about this. We just need a dao_get_template_by_id. Whether it gets it from cache, DB, or elsewhere it shouldnt matter?

tests/app/cache/test_cache_events.py
```diff
@@ -0,0 +1,112 @@
...
+    user = create_user()
+    mocked_invalidate = mocker.patch("app.cache.cache_events.invalidate_group_keys")
+
+    save_model_user(user, update_dict={"name": "Updated name"})
+
+    mocked_invalidate.assert_called_once_with("user", str(user.id))
+
+
+def test_template_update_invalidates_template_cache(notify_db_session, mocker):
+    register_cache_orm_events()
```
jzbahrai, line 104: can we add tests for _clear_cache_invalidations_after_rollback

app/cache/cache_events.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def register_cache_orm_events():
+    """Register registry-driven cache listeners once per process."""
+    global _EVENTS_REGISTERED
+
+    if _EVENTS_REGISTERED:
+        return
+
+    event.listen(Session, "after_flush", _collect_cache_invalidations)
```
jzbahrai, line 73: this event api is a whole thing! didnt know about it

app/cache/cache_events.py
```diff
@@ -0,0 +1,77 @@
...
+    event.listen(Session, "after_commit", _invalidate_cache_after_commit)
+    event.listen(Session, "after_rollback", _clear_cache_invalidations_after_rollback)
+
+    _EVENTS_REGISTERED = True
```
jzbahrai, line 77: do this is done to makesure we only invoke this once right. but this is a global on the python level.. we have the app initialize across multiple pods. This will lead to the correct behaviour?

app/cache/cache_events.py
whabanks: The listener runs after each flush for a given session object, but it only looks at ORM instances in session.new, session.dirty, and session.deleted and only queues invalidations for models present in the registry. The actual Redis invalidation happens after commit. 

It does loop through the entire registry, which I don't think will be a final solution. This was just the simplest way to get this up and running.

app/cache/cache_invalidation_registry.py
whabanks: Correct, the ID itself does not need to change. Any insert, update, or delete of the ORM instance triggers the rule. `entity_id_attribute` identifies which cache group owns the affected data. For example, changing a template name uses the unchanged template ID to invalidate that template’s cache entries.

app/cache/cache_invalidation_registry.py
whabanks: Yea currently the history tables aren't included here. Versioned reads from TemplateHistory would bypass the cache as is. A template update would immediately invalidate the existing template cache post-commit. That said, the dogpile cache keys use method params as fingerprints. Take `dao_get_template_by_id(template_id, version=None)` as an example, if you're querying a non-history template, the cache key would be something like:

`template-<id>:dao_get_template_by_id:version=None`

Whereas something from template history would be still be distinct.

`template-<id>:dao_get_template_by_id:version=3`

Will muck around with this more though, you may be right here.

app/dao/service_permissions_dao.py
whabanks: The event listeners are registered only when `FF_USE_DOGPILE_CACHING` is enabled, so this change would not invalidate dogpile entries while the flag is off.

app/dao/service_permissions_dao.py
whabanks: Yea this was a limitation between deleting an entity loaded into the ORM session and straight DML delete statements. This pattern won't be used once the second PR is in. In short, deletes via DML statements don't trigger the usual `on_commit` or `on_flush` events so a different solution was needed.

https://github.com/cds-snc/notification-api/pull/3045/changes#diff-2be2376b302522845d59e928b03be0589bab91c70ee32146ff8ba3f0d4c44636

app/dao/templates_dao.py
whabanks: Agreed, the callers don't need to know where this comes from, and we won't use separate methods. I initially separated these because the cached function returns a serialized dictionary while the existing DAO returns a session-bound ORM object. And it was a simple way to test via a feature flag so we aren't calling the dogpile cache at all in prod.

tests/app/cache/test_cache_events.py
whabanks: Good call, I'll add a test for this.

app/cache/cache_events.py
whabanks: Yea me neither! There's a whole list of events that can be hooked into. Pretty neat stuff: https://docs.sqlalchemy.org/en/21/orm/events.html

app/cache/cache_events.py
whabanks: Yes, it's to ensure that per-process / app init only one set of listeners is attached to the global SQLAlchemy session at a time. Currently the logic relies on listener registration occurring during single-threaded app startup,  whic  I suppose may not always be true. If two threads could call this method concurrently, both see `_EVENTS_REGISTERED=False` and both attach event listeners there could be an issue.

I'll look into utilizing a lock for this, or see if SQLAlchemy has any built in helper methods to detect what events are currently being listened to.

tests/app/cache/test_cache_events.py
whabanks: Ah looks like I added that in the followup PR for users, i'll also add it for templates/services: https://github.com/cds-snc/notification-api/pull/3045/changes#diff-c083a965e879c0b699ae19b5cdc473c57abe3e3251dc20be8f1c9a4d53ac8d62R116

app/cache/cache_events.py
jimleroyer: Yep, this will run in a configuration of 1 process that has 1 thread per workers (4 total on top of my head).

app/cache/cache_invalidation_registry.py
jimleroyer: I am unsure if there is an issue here. Isn't that okay if we don't support the template history table to begin with? That would mean that we don't use the cache to read these. Unless it's on a hot path for reads, we are okay not having it. If that is, we can still add later.

app/cache/cache_invalidation_registry.py
```diff
@@ -0,0 +1,47 @@
...
+
+    namespace: str
+    entity_id_attribute: str
```
jimleroyer, line 21: Did you encounter a situation where an ID might be a compound key? i.e. 2 different IDs together?

app/cypress/rest.py
```diff
@@ -141,6 +142,7 @@ def _destroy_test_user(email_name):
         # and is a bug
         cypress_service = Service.query.filter_by(id=current_app.config["CYPRESS_SERVICE_ID"]).first()
         cypress_service.created_by_id = current_app.config["CYPRESS_TEST_USER_ID"]
+        invalidate_service_cache_keys(cypress_service.id)
```
jimleroyer, line 145: In this case, a manual call to invalidation is made: sqlalchemy does not get the event when updating the  service?

app/dao/service_permissions_dao.py
jimleroyer: Yeah so that's a good point here. Performance might not be so bad here because we wouldn't have that much permissions per service, but we will have sql operations when we want to skip the ORM operations (too expensive) and a manual call to cache invalidation will be necessary. It would be good to document and support such cases, so that developers know which route they need to take.

app/dao/users_dao.py
```diff
@@ -32,7 +33,9 @@ def save_user_attribute(usr: User, update_dict=None):
     if updates.get("blocked"):
         updates["current_session_id"] = "00000000-0000-0000-0000-000000000000"
 
-    db.session.query(User).filter_by(id=usr.id).update(updates)
+    for attribute, value in updates.items():
+        setattr(usr, attribute, value)
+    db.session.add(usr)
```
jimleroyer, line 39: Hmm ok, we will need to coach devs on what to use and when like in this case, otherwise it's manual cache invalidation.

app/dao/users_dao.py
```diff
@@ -116,6 +120,12 @@ def get_user_by_id(user_id=None) -> User:
     return User.query.filter_by().all()
 
 
+@cache_on_arguments(namespace="user", group_by="user_id")
+def dao_get_user_by_id_cached(user_id) -> dict:
```
jimleroyer: Is that one also part of previous code iteration? would that one be removed too? 

If I get this right: to be removed per previous PR questions/answers.

app/service/rest.py
```diff
@@ -223,8 +223,7 @@ def get_service_by_id(service_id):
         data = get_detailed_service(service_id, today_only=request.args.get("today_only") == "True")
     else:
         if current_app.config.get("FF_USE_DOGPILE_CACHING", False) is True:
-            fetched = dao_fetch_service_by_id_cached(service_id)
-            data = fetched
+            data = dao_fetch_service_by_id_cached(service_id)
```
jimleroyer: To be removed per previous PR questions/answers.

app/template/rest.py
```diff
@@ -234,6 +235,9 @@ def get_all_templates_for_service(service_id):
 
 @template_blueprint.route("/<uuid:template_id>", methods=["GET"])
 def get_template_by_id_and_service_id(service_id, template_id):
+    if current_app.config.get("FF_USE_DOGPILE_CACHING", False) is True:
+        return jsonify(data=dao_get_template_by_id_cached(template_id, service_id))
```
jimleroyer: To be removed per previous PR questions/answers.

app/user/rest.py
```diff
@@ -604,6 +605,9 @@ def send_new_template_category_request(user_id):
 @user_blueprint.route("/<uuid:user_id>", methods=["GET"])
 @user_blueprint.route("", methods=["GET"])
 def get_user(user_id=None):
+    if user_id and current_app.config.get("FF_USE_DOGPILE_CACHING", False) is True:
+        return jsonify(data=dao_get_user_by_id_cached(user_id))
```
jimleroyer: To be removed per previous PR questions/answers.

tests/app/cache/test_cache_events.py
```diff
@@ -0,0 +1,152 @@
...
+    ServicePermission,
+    ServiceUser,
+    Template,
+    TemplateRedacted,
+    User,
+)
+from tests.app.db import create_service, create_template, create_user
+
+
+def test_register_cache_orm_events_is_thread_safe(mocker):
```
jimleroyer: Let's use test classes to regroup the same functionality, as we usually do on the team's best practice. In this case, a single file such as `test_cache.py` with a `TestCacheEvents` class within the module would do the trick.

tests/app/test_dogpile_caching.py
jimleroyer: That's what I meant by grouping tests within test classes here. This overall module could be renamed test_cache.py and we can regroup the events tests in the same file.

app/cache/cache_invalidation_registry.py
jzbahrai: so one of the issues we had with billing was at send time we use TemplateHistory based on the version the client decided to send with. But when a client was changing the category of a template it wasn't getting written to template history. So template v15 != templatehistory v15 (this has been fixed)
now with cache invalidation, I want to make sure that the model that is caching the result is still writing it all the way back to template history

app/cache/cache_events.py
jzbahrai: what was the decision around this? we want to keep the global?

app/cache/cache_events.py
```diff
@@ -0,0 +1,78 @@
...
+            logger.exception(
+                "Failed to invalidate cache group",
+                extra={
+                    "cache_namespace": namespace,
+                    "cache_entity_id": entity_id,
+                },
+            )
+
+
+def _clear_cache_invalidations_after_rollback(session):
```
jimleroyer, line 62: Let's add type annotations when possible on new code.

app/cache/cache_events.py
whabanks: I ended up going with a lock instead of a boolean flag here. We acquire the lock during event registration now, it should be more robust against any potential edge cases.

app/cache/cache_invalidation_registry.py
whabanks: After a bit of digging I think I understand what you mean here. It isn't just that the cache keys are distinct, it's that caching shouldn't bypass or alter writes that keep `Template` and `TemplateHistory` aligned. I dug into code on these paths and it should be working like so:

- `dao_update_template()` uses the `version_class` decorator to create `TemplateHistory` rows.
- The manual category/process-type/reply-to update methods increment the template version, reload the updated `Template`, and explicitly call `_create_template_history_row(template)` to load the new `TemplateHistory` object into the SQLAlchemy session.
- The template change and history row are then committed in the same transaction.
- The commit event fires and post-commit we invalidate the current-template dogpile caches related to templates with that `id`.

So the order is: update Template and write matching `TemplateHistory` → commit both → invalidate the cached template representation.

That said, I will look into adding an integration test that:
1. Populates the dogpile cache
2. Changes a category/process-type/reply-to
3. Assert the `Template` and `TemplateHistory` match correctly
4. Assert the old cache is invalidated
5. Assert the new cached object contains the updated data

app/cache/cache_invalidation_registry.py
whabanks: Yes, there are compound keys, ServiceUser(service_id, user_id) is an example of this but none of the current cache groups have compound ownership. That model invalidates the service and user groups independently rather than utilizing a single (service_id, user_id) group. We rely on the cross-namespace invalidation definitions in the registry to handle cross-cutting concerns at the moment.

If we introduce a genuinely pair-scoped cache later, we’ll need to extend the rule to accept and deterministically serialize multiple attributes. It's definitely on my todo list if/when it becomes necessary.

tests/app/cache/test_cache_events.py
whabanks: Done!

app/dao/users_dao.py
whabanks: Noted in my TODOs. Will put together a guidance document for these things.

app/cypress/rest.py
whabanks: In this case we don't need the manual invalidation. I think this change was an artifact from prior to using the ORM event driven style. Since we're loading the service object into the session, then changing the `.created_by_id` this would mark that model `dirty` and trigger invalidation post-commit. I'll remove this line.

In this PR, which doesn't have the DML changes. If we were querying directly with DML statements, then yes the changes here would both go around the event triggers and require the manual invalidation.

app/cache/cache_invalidation_registry.py
whabanks: We do have models with compound database keys. For example, `ServicePermission` is keyed by (service_id, permission) and `ServiceUser` by (service_id, user_id). However, none of the cache groups currently require those compound keys as their invalidation identity.

Currently the `entity_id_attribute` just identifies the owner of the cache group, and we're invalidating somewhat broadly. 

e.g: A committed change to a `ServiceUser` will invalidate two groups: `service:<service_id>*` and `user:<user_id>*`. It does not invalidate one compound (service_id, user_id) group.

If we later add a cache whose ownership is composite, such as a value belonging specifically to one (service_id, user_id) pair, this type would need to evolve. We could allow `entity_id_attributes: tuple[str, ...]`, serialize those values deterministically into one group ID, and use the same representation for both key generation and invalidation. This would facilitate more surgical targeting during invalidation. It's definitely something I am considering for a future enhancement.

jimleroyer (approved):
LGTM - We met with @whabanks this morning and discussed this PR and its readiness. Will had a few more items to polish this PR but this is ready to be merged into `main` and activated in staging environment.

app/cache/cache_events.py
whabanks: Added!

PR #3034, opened 2026-08-31
Add v2 endpoint to delete scheduled notifications

smcmurtry:
# Summary | Résumé

This PR:
- Adds v2 endpoint to delete scheduled notifications: `v2/notifications/<id>`
- Adds tests

- https://github.com/cds-snc/notification-planning/issues/3202

Run this branch locally. You can test the new delete endpoint with the following curl commands:

Create a new scheduled email notification:
```
curl -X POST http://localhost:6011/v2/notifications/email \
  -H "Authorization: ApiKey-v1 YOUR_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{
    "email_address": "user@example.com",
    "template_id": "YOUR_TEMPLATE_ID",
    "scheduled_for": "2026-09-02T15:15:00"
  }'
```

Note the returned notification id and use it in the following delete request:
```
curl -X DELETE http://localhost:6011/v2/notifications/NOTIFICATION_ID \
  -H "Authorization: ApiKey-v1 YOUR_API_KEY"
```

You should get a success response.

Then try deleting a notification ID that does not exist, and one that matches an already sent notification and verify that you get error responses.


app/dao/notifications_dao.py
```diff
@@ -564,6 +564,15 @@ def dao_delete_notifications_by_id(notification_id):
     db.session.query(Notification).filter(Notification.id == notification_id).delete(synchronize_session="fetch")
 
 
+@statsd(namespace="dao")
+@transactional
+def dao_delete_scheduled_notification_by_id(notification_id):
```
andrewleith, line 569: Do we want a guard here as well to protect future-devs from using this method to delete non-scheduled notifications? Or maybe we leave it like this to be flexible for other uses later?

app/v2/notifications/delete_notifications.py
```diff
@@ -0,0 +1,30 @@
...
+        return jsonify(result="error", message="Notification not found in database"), 404
+
+    scheduled_notification = notification.scheduled_notification
+    if scheduled_notification is None:
+        raise BadRequestError(message="Notification is not scheduled and cannot be deleted")
+
+    if not scheduled_notification.pending:
+        raise BadRequestError(message="Notification has already been sent and cannot be deleted")
+
+    notifications_dao.dao_delete_scheduled_notification_by_id(notification_id)
```
andrewleith: Potential concurrency issue here: the endpoint reads `scheduled_notification.pending`, then calls a DAO method that unconditionally deletes both rows. Another worker can mark the scheduled notification as processed between those operations.

app/dao/notifications_dao.py
andrewleith, line 569: Should this method return the row count matched to indicate to the caller it was successful?

andrewleith (commented):
A few comments below

app/dao/notifications_dao.py
smcmurtry: I made this change

smcmurtry: Good idea, I added a guard to filter by `ScheduledNotification.pending.is_(True)` so that it can't be used to delete already sent notifications.

app/v2/notifications/delete_notifications.py
smcmurtry: I made a fix for this:
> the pending check and delete now happen atomically inside the `DAO's DELETE ... WHERE pending = true`, and the endpoint checks the returned row count — if another worker processed the notification between the initial check and the delete call, it now returns a 400 instead of deleting a sent notification.

andrewleith (approved):
LGTM

Out of scope here, but we may want to consider a fix in the code around https://github.com/cds-snc/notification-api/blob/main/app/dao/notifications_dao.py#L743 so when we pick up scheduled notifications we make sure they aren't selected by the delete operation.

PR #3036, opened 2026-09-01
Allow local docsite with CORS

smcmurtry:
# Summary | Résumé

Allow local docsite url with CORS - this is so we can run the documentation site locally and point it at an api spec being served from a locally running api.

- https://github.com/cds-snc/notification-planning/issues/3202


andrewleith (approved):
LGTM

PR #3044, opened 2026-09-03
security: remove staging smoke access keys

P0NDER0SA:
## Summary

Remove long-lived staging AWS access-key usage from the API repository's test paths.

## Changes

- Removed the static `STAGING_AWS_ACCESS_KEY_ID` / `STAGING_AWS_SECRET_ACCESS_KEY` credential step from `.github/workflows/docker.yaml`. The job already authenticates to ECR through GitHub Actions OIDC.
- Removed the optional `SMOKE_AWS_ACCESS_KEY_ID` / `SMOKE_AWS_SECRET_ACCESS_KEY` fallback from `tests_smoke/smoke/common.py`. Smoke tests now always use the ambient AWS credential provider chain, including OIDC-provided credentials.

## Validation

- YAML parsing passed
- Python compilation passed
- `git diff --check` passed
- Ruff passed
- Ruff-format passed
- No tracked workflow or smoke-test code consumes staging AWS access-key secrets

Terraform IAM-user deactivation should be performed separately after this change is deployed and staging behavior is observed, so any remaining consumers can be identified without disabling the identity prematurely.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -54,13 +54,6 @@ jobs:
       run: |
         docker push $DOCKER_SLUG:latest && docker push $DOCKER_SLUG:${GITHUB_SHA::7}
 
-    - name: Configure AWS credentials
-      uses: aws-actions/configure-aws-credentials@5fd3084fc36e372ff1fff382a39b10d03659f355 # v2.2.0
-      with:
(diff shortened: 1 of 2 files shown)


andrewleith (commented):
Can you still run the smoke tests without that?

PR #3045, opened 2026-09-03
Use do_orm_execute for DML driven invalidation

whabanks:
# Summary | Résumé
Experimentally branched off of: https://github.com/cds-snc/notification-api/pull/3033

Many dao methods for updating / deleting leverage legacy sqlalchemy DML statements `Query.update()` and `Query.delete()`. These DML queries bypass the typical session ORM events like `after_commit` and `after_flush`. This prevents us from relying solely on ORM events for cache invalidation. 

However, modernized DML syntax (`session.execute(<someQuery>)`) _can_ be intercepted via the `do_orm_execute` event allowing us to follow the same event driven invalidation pattern for both session ORM changes and DML statement changes.
- Add `do_orm_execute` event listener
- Add `cache_invalidating_dml` wrapper to execute DML statements and attach cache metadata to the `session.info`
- Add `_queue_cache_invalidation` to store deduplicated cache invalidation actions to be executed post-commit
- Migrate / modernize DML statements from legacy `Query.update()/delete()` to `session.execute(<query))`

## How does this work?

Bulk DML queries in SQLAlchemy bypass session.dirty and after_flush instance detection. Statements like the one below are both [legacy syntax from v1.4](https://docs.sqlalchemy.org/en/21/orm/queryguide/query.html#sqlalchemy.orm.Query.update) and do not trigger typical session ORM events. (`after_flush`, `after_commit`, etc.)

```python
# Legacy syntax & no event triggers here
db.session.query(User).filter_by(id=user.id).update(updates)
```

The helper method [`cache_invalidating_dml()`](https://github.com/cds-snc/notification-api/pull/3045/changes#diff-77b89800c3c1f4f17540369f9c461cc6d60f31e0153b6c36037443f450f9fca2R19) adds cache metadata to the SQL statement that can be later extracted. Note the [updated query syntax](https://docs.sqlalchemy.org/en/21/orm/session_basics.html#querying), this facilitates listening to `do_orm_execute` events and is central to enabling event driven invalidation for DML queries. 

```python
statement = cache_invalidating_dml(
    delete(Permission).where(Permission.user_id == user.id),
    user_id=user.id,
)

db.session.execute(statement)
```

Behind the scenes  cache metadata is included:

```python
{
    "cache_invalidation_entity_ids": {
        "user_id": user.id,
    }
}
```

[`_intercept_bulk_operations()`](https://github.com/cds-snc/notification-api/pull/3045/changes#diff-77b89800c3c1f4f17540369f9c461cc6d60f31e0153b6c36037443f450f9fca2R85) is attached to the `do_orm_execute` event. When `db.session.execute(statement)` is run, it executes and:

1. Verifies the statement is an `update` or `delete`
2. Determines the mapped model is of type `Permission`
3. calls [`_queue_model_invalidations()`](https://github.com/cds-snc/notification-api/pull/3045/changes#diff-77b89800c3c1f4f17540369f9c461cc6d60f31e0153b6c36037443f450f9fca2R36) 
    * Note: We queue the invalidation(s) in the transaction so that we can wait for either the `after_commit` or `after_rollback` events to fire, so that we only invalidate after successful mutation, or drop the invalidations when rolling back
4. Looks up the `Permission` entity in the [`CACHE_INVALIDATION_REGISTRY`](https://github.com/cds-snc/notification-api/pull/3033/changes#diff-cc53cbd7495b50eee747cbc901bc8aa77746a7d999922a5b40fd7c683d942c6eR38)
5. Validates that the required id's are present and builds the invalidation actions
6. `after_commit` event fires, signalling successful data changes and `_invalidate_cache_after_commit` is called to invalidate the key

### At a high level

```mermaid
flowchart TD
    A[Database mutation] --> B{Mutation style}

    B -->|ORM objects| C[Flush]
    C --> D[after_flush]
    D --> E[Resolve model through registry]

    B -->|Bulk DML| F[cache_invalidating_dml]
    F --> G[Session.execute]
    G --> H[do_orm_execute]
    H --> E

    E --> I[Store namespace and ID in Session.info]
    I --> J{Transaction result}

    J -->|Commit| K[after_commit]
    K --> L[Pop pending set]
    L --> M[Delete grouped Redis keys]

    J -->|Rollback| N[after_rollback]
    N --> O[Discard pending set]
```

### Benefits of this approach
1. Developers need only to do two things in any given situation to leverage dogpile while everything else is generically taken care of in the background.

**ORM session based actions**
1. Annotate the dao method, specifying the group it belongs to. User for example:
```python
@cache_on_arguments(namespace="user", group_by="user_id")
def dao_get_user_by_id(user_id) -> dict:
    . . .
```
2. Check that the [`CACHE_INVALIDATION_REGISTRY`](https://github.com/cds-snc/notification-api/pull/3033/changes#diff-cc53cbd7495b50eee747cbc901bc8aa77746a7d999922a5b40fd7c683d942c6eR38) entries for the associated entity exist and clear the keys that it should.

```python
    User: [
        CacheInvalidationRule(namespace="user", entity_id_attribute="id"),
    ],
```

**DML based statements**
1. Wrap any bulk `update` or `delete` statements that affect cached values with the helper method
```python
statement = update(User).where(User.id == usr.id).values(**updates)
db.session.execute(cache_invalidating_dml(statement, id=usr.id))
```
2. Check that the [`CACHE_INVALIDATION_REGISTRY`](https://github.com/cds-snc/notification-api/pull/3033/changes#diff-cc53cbd7495b50eee747cbc901bc8aa77746a7d999922a5b40fd7c683d942c6eR38) entries for the associated entity exist and clear the keys that it should.

```python
    User: [
        CacheInvalidationRule(namespace="user", entity_id_attribute="id"),
    ],
```

2. The current approach is compatible with threaded workers

- SQLAlchemy sessions are scoped per request context. Concurrent requests should have distinct sessions
    - Therefore the events we are listening to are also run against distinct, scoped sessions
- While the `CACHE_INVALIDATION_REGISTRY` is technically global, it's read-only at runtime and will never be mutated.
- Redis clients (used for scan/delete invalidations) use thread-safe connection pools. Each `invalidate_group_keys` call maintains it's own local scan cursor, therefore parallel scans should not corrupt one another. 

# Drawbacks
1. Cache invalidation is synchronous with the SQLAlchemy commit path, and we add a small amount of overhead to DB write operations.
```
database COMMIT
    -> SQLAlchemy after_commit
    -> Redis SCAN
    -> Redis DELETE
    -> commit() returns to caller
```

This would add some latency to API responses when executing an update or delete. The silver lining here is that all subsequent gets on that entity would hit cache cache instead of the DB and thus should offset added latency. You trade one slightly longer API call for many subsequent faster API calls.

_TODO: Fill in test instructions for the reviewer._

None.


app/dao/permissions_dao.py
```diff
@@ -1,4 +1,7 @@
+from sqlalchemy import delete
+
 from app import db
+from app.cache.cache_events import cache_invalidating_dml
```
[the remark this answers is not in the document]
whabanks: Fixed

app/cache/cache_events.py
```diff
@@ -12,9 +12,40 @@
...
+_CACHE_INVALIDATION_ENTITY_IDS_OPTION = "cache_invalidation_entity_ids"
 _EVENTS_REGISTERED = False
 
 
+def cache_invalidating_dml(statement, **entity_ids):
+    """Attach entity IDs for registry-driven invalidation of ORM bulk DML."""
+    return statement.execution_options(**{_CACHE_INVALIDATION_ENTITY_IDS_OPTION: entity_ids})
+
+
+def _as_entity_id_collection(value):
```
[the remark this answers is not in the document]
whabanks: Fixed

app/cache/cache_events.py
```diff
@@ -10,16 +10,43 @@
 from sqlalchemy import event
 from sqlalchemy.orm import Session
 
+from app.cache.cache_dml import _CACHE_INVALIDATION_ENTITY_IDS_OPTION
 from app.cache.cache_invalidation_registry import CACHE_INVALIDATION_REGISTRY
 from app.caching import invalidate_group_keys
 
 _CACHE_INVALIDATIONS_KEY = "cache_invalidations_to_run"
+_CACHE_INVALIDATION_ENTITY_IDS_OPTION = "cache_invalidation_entity_ids"
```
jimleroyer: This is a string constant named `_CACHE_INVALIDATION_ENTITY_IDS_OPTION` defined on that line, as well as an import with the same name at line 13, but a tuple1(string). It's the end of the day so I might be confused, but yes I am confused atm. 😅

app/cache/cache_events.py
```diff
@@ -10,16 +10,43 @@
...
 _EVENT_REGISTRATION_LOCK = Lock()
-
 # ORM events are tied to a global sqlalchemy session which could exist outside of
 # a flask context, so use a module level logger instead.
 logger = logging.getLogger(__name__)
 
+def _as_entity_id_collection(value):
```
jimleroyer: Let's add type annotations on parameters and return types.

app/cache/cache_events.py
```diff
@@ -10,16 +10,43 @@
...
+
+
+def _queue_cache_invalidation(session, namespace, entity_id):
+    session.info.setdefault(_CACHE_INVALIDATIONS_KEY, set()).add((namespace, str(entity_id)))
+
+
+def _queue_model_invalidations(session, model, entity_ids):
+    rules = CACHE_INVALIDATION_REGISTRY.get(model, ())
+    required_attributes = {rule.entity_id_attribute for rule in rules}
+    missing_attributes = required_attributes.difference(entity_ids)
```
jimleroyer, line 64: I would add some comments to say what's the potential failing scenario and why we are trying to prevent this.

app/cache/cache_events.py
```diff
@@ -10,16 +10,43 @@
...
+        yield from value
+    else:
+        yield value
+
+
+def _queue_cache_invalidation(session, namespace, entity_id):
+    session.info.setdefault(_CACHE_INVALIDATIONS_KEY, set()).add((namespace, str(entity_id)))
+
+
+def _queue_model_invalidations(session, model, entity_ids):
```
jimleroyer: I wonder if `entity` would be a clearer name than `model` here. In SQLAlchemy/ORM terms, this helper seems to queue invalidations for mapped entities rather than for “models” in the abstract, so `_queue_entity_invalidations` may better reflect what it does.

```suggestion
def _queue_entity_invalidations(session, model, entity_ids):
```

app/cache/cache_events.py
```diff
@@ -64,6 +89,19 @@ def _clear_cache_invalidations_after_rollback(session):
...
+    if not (orm_context.is_delete or orm_context.is_update):
+        return
+
+    mapper = orm_context.bind_mapper
+    if mapper is None:
+        return
+
+    entity_ids = orm_context.execution_options.get(_CACHE_INVALIDATION_ENTITY_IDS_OPTION)
+    if entity_ids:
+        _queue_model_invalidations(orm_context.session, mapper.class_, entity_ids)
```
jimleroyer: Ah there we actually refer to the entity via `entity_ids` (and not model). 😛

app/cache/cache_events.py
```diff
@@ -64,6 +89,19 @@ def _clear_cache_invalidations_after_rollback(session):
     session.info.pop(_CACHE_INVALIDATIONS_KEY, None)
 
 
+def _intercept_bulk_operations(orm_context):
```
jimleroyer: We didn't hook this function to any listener yet, right?

app/dao/service_permissions_dao.py
```diff
@@ -14,19 +17,11 @@ def dao_add_service_permission(service_id, permission):
...
-    service_permissions = ServicePermission.query.filter(
+    statement = delete(ServicePermission).where(
         ServicePermission.service_id == service_id,
         ServicePermission.permission == permission,
-    ).all()
-
-    for service_permission in service_permissions:
-        db.session.delete(service_permission)
+    )
+    result = db.session.execute(cache_invalidating_dml(statement, service_id=service_id))
```
jimleroyer, line 24: cool cool

app/dao/users_dao.py
```diff
@@ -33,9 +34,8 @@ def save_user_attribute(usr: User, update_dict=None):
     if updates.get("blocked"):
         updates["current_session_id"] = "00000000-0000-0000-0000-000000000000"
 
-    for attribute, value in updates.items():
-        setattr(usr, attribute, value)
-    db.session.add(usr)
+    statement = update(User).where(User.id == usr.id).values(**updates)
+    db.session.execute(cache_invalidating_dml(statement, id=usr.id))
```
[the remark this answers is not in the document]
jimleroyer: @copilot what do you mean that the users could remain stale? That these wouldn't be updated on the update? or after the update? what if we don't expect these user objects to change after and we are okay with these remaining stale?

jimleroyer (commented):
Very nice additions. This likely needs some more logic to hook in the actual listener registration with `_intercept_bulk_operations` and more unit-tests.

app/dao/users_dao.py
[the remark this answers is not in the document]
whabanks: Yea this is the case here and is specific to the code that follows this save operation. By default [SQLAlchemy will choose `synchronize_session='auto'](https://docs.sqlalchemy.org/en/21/orm/queryguide/dml.html#selecting-a-synchronization-strategy) which could resolve to either `evaluate` (doesn't refresh the model in session) or `fetch` (does refresh).

Needing to manually specify this is extra cognitive load but I think it's a good thing to explicitly document and guarantee object synchronization when it's needed. We already explicitly specify this in our code: 

- https://github.com/cds-snc/notification-api/blob/main/app/dao/api_key_dao.py#L84-L85
- https://github.com/cds-snc/notification-api/blob/main/app/dao/inbound_sms_dao.py#L108-L109

app/cache/cache_events.py
whabanks: Cleaned up!

app/cache/cache_events.py
whabanks: This is hooked up to `do_orm_execute` in `register_cache_orm_events`

app/cache/cache_events.py
whabanks: Updated the name here and in tests.

app/cache/cache_events.py
whabanks: Documented!

app/cache/cache_events.py
whabanks: Added a bunch of type annotations. Since `value` here can be quite a few things, I defined a custom type `EntityIdValue` for this. Let me know what you think.

PR #3053, opened 2026-09-14
Revert "Summoning heavenly logging to smite mismatches on priority assignment"

jimleroyer:
Reverts cds-snc/notification-api#3046

We have an issue with this one as we deployed to production and likely took code paths it did not in staging env.


ben851 (approved):
gogogogogogo

PR #3059, opened 2026-09-15
Migrate image consumers away from latest

P0NDER0SA:
## Summary

Migrate image consumers away from mutable \ tags while keeping deployment and scan tag selection aligned.

## Details

- Removes the build cache dependency on \\`notify-api:latest\\`
- Generates the build SBOM from the build SHA
- Makes the scheduled vulnerability scan consume \\`API_DOCKER_TAG\\` from manifests staging config

## Validation

- Scheduled scans read the exact image tag from \
- Scan workflows fail when the required manifest tag is missing or mutable
- Build-time SBOMs continue to use the same SHA tag dispatched to manifests
- Ran \
- Workflow linting was not run locally because \ is unavailable


.github/workflows/docker-vulnerability-scan.yml
```diff
@@ -31,12 +31,24 @@ jobs:
...
+      - name: Read staging image tag from manifests
+        id: image-tag
+        run: |
+          curl --fail --silent --show-error \
+            --output staging.env \
+            https://raw.githubusercontent.com/cds-snc/notification-manifests/main/helmfile/overrides/staging.env
+          IMAGE_TAG="$(sed -n 's/^API_DOCKER_TAG:[[:space:]]*"\([^"]*\)"/\1/p' staging.env)"
+          test -n "$IMAGE_TAG"
+          test "$IMAGE_TAG" != "latest"
+          test "$IMAGE_TAG" != "bootstrap"
```
jimleroyer, line 43: Love the `test` instructions over here

.github/workflows/docker.yaml
```diff
@@ -42,9 +42,7 @@ jobs:
 
     - name: Build
       run: |
-        docker pull $DOCKER_SLUG:latest
```
jimleroyer, line 45: Hmm why and what were we pulling before? 🤔

jimleroyer (approved):
LGTM!

PR #3063, opened 2026-09-16
 Add a increment log

jzbahrai:
# Summary | Résumé

Seeing this error show up in production 
```
File "/app/.venv/lib/python3.12/site-packages/redis/connection.py", line 159, in pack for arg in map(self.encode, args): ^^^^^^^^^^^^^^^^^^^^^^ File "/app/.venv/lib/python3.12/site-packages/redis/_parsers/encoders.py", line 29, in encode raise DataError( redis.exceptions.DataError: Invalid input of type: 'NoneType'. Convert to a bytes, string, int or float first. "

```
I can not replicate this on staging or locally. I used the same templateID the above is being caused for:
```
| filter kubernetes.container_name like /^notify-api/
| filter log like /DataError/
| parse log /template_id:\s*'(?<template_id>[a-f0-9-]+)'/
| stats count(*) as error_count by  template_id
```
It return two separate templates. I used those templates in localhost and sent a message, I still couldn't replicate the bug above. 

I am not submitting a log to hopefully get more data, when this gets merged to production. I will revert it from prod after I get data from the logs

# Ticket

https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/3490


P0NDER0SA (approved):
Looks good

